Issue #25, opened 2018-05-09
core/drm: Implement cursor planes


avdgrinten:
This would increase performance on drivers like gfx/virtio.

PR #79, opened 2019-05-15
Implement accelerated cursors.

qookei:
This pull request implements functionality mentioned in issue #25.

### DIFF
--- a/core/drm/src/core.cpp
+++ b/core/drm/src/core.cpp
@@ -103,6 +103,28 @@ drm_core::Device::Device()
 		};
 	};
 	_modeIdProperty = std::make_shared<ModeIdProperty>();
+
+	struct CrtcXProperty : drm_core::Property {
+		CrtcXProperty()
+		: drm_core::Property{drm_core::IntPropertyType{}} { }
+
+		bool validate(const Assignment& assignment) override {
+			return true;
+		};
+	};
+	_crtcXProperty = std::make_shared<CrtcXProperty>();
+	
+	struct CrtcYProperty : drm_core::Property {
+		CrtcYProperty()
+		: drm_core::Property{drm_core::IntPropertyType{}} { }
+
+		bool validate(const Assignment& assignment) override {
+			return true;
+		};
+	};
+	_crtcYProperty = std::make_shared<CrtcYProperty>();
+
+
 }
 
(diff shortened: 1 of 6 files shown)


core/drm/src/core.cpp
```diff
@@ -873,6 +907,105 @@ drm_core::File::ioctl(void *object, managarm::fs::CntRequest req,
...
+		if (cursor_plane == nullptr) {
+			resp.set_result(ENXIO);
+			resp.set_error(managarm::fs::Errors::SUCCESS);
+			auto ser = resp.SerializeAsString();
+			auto &&transmit = helix::submitAsync(conversation, helix::Dispatcher::global(),
+			helix::action(&send_resp, ser.data(), ser.size()));
+			COFIBER_AWAIT transmit.async_wait();
+			HEL_CHECK(send_resp.error());
+		}else{
+			std::vector<Assignment> assignments;
```
avdgrinten: Returning (`COFIBER_RETURN();`) in the `if` case would lower the indentation level by 1 here.

drivers/gfx/vmware/src/main.cpp
```diff
@@ -209,6 +232,251 @@ std::unique_ptr<drm_core::Configuration> GfxDevice::createConfiguration() {
 	return std::make_unique<Configuration>(this);
 }
 
+COFIBER_ROUTINE(cofiber::no_future, GfxDevice::waitIrq(uint32_t irq_mask), ([=](){
```
avdgrinten: This does not wait for the IRQ: to properly wait for the IRQ, the function needs to return a `async::result<void>` and the caller needs to `COFIBER_AWAIT` the result.

drivers/gfx/vmware/src/main.cpp
```diff
@@ -226,14 +494,43 @@ bool GfxDevice::Configuration::capture(std::vector<drm_core::Assignment> assignm
 	for (auto &assign : assignment) {
 		if (assign.property == _device->srcWProperty()) {
 			assert(assign.property->validate(assign));
-			_width = assign.intValue;
+			if (assign.object == _device->_cursorPlane) {
```
avdgrinten: When checking for `cursorPlane`, we also need to check for the capability as `cursorPlane` can be `nullptr` (?).

Issue #9, opened 2017-10-01
general: Use libarch API to allocate DMA buffers


avdgrinten:
Inspect the drivers to check where this makes sense.

Some locations that need to be refactored:
- [ ] core/virtio: Performs manual contiguous allocations
- [ ] drivers/virtio: Use DMA API inside the block driver

avdgrinten:
This has been done during the IOMMU introduction.

PR #82, opened 2019-05-19
core/drm: properly return errors when handling DRM_IOCTL_MODE_CURSOR

qookei:
Related to https://github.com/managarm/mlibc/pull/9

### DIFF
--- a/core/drm/src/core.cpp
+++ b/core/drm/src/core.cpp
@@ -918,8 +918,7 @@ drm_core::File::ioctl(void *object, managarm::fs::CntRequest req,
 		auto cursor_plane = crtc->cursorPlane();
 
 		if (cursor_plane == nullptr) {
-			resp.set_result(ENXIO);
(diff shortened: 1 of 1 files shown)


core/drm/src/core.cpp
```diff
@@ -990,7 +989,7 @@ drm_core::File::ioctl(void *object, managarm::fs::CntRequest req,
 			});
 		}else{
 			printf("\e[35mcore/drm: invalid request whilst handling DRM_IOCTL_MODE_CURSOR\e[39m\n");
-			resp.set_result(EINVAL);
+			resp.set_error(managarm::fs::Errors::ILLEGAL_ARGUMENT);
```
avdgrinten: `resp.error` is overwritten a few lines below.

avdgrinten:
I'd still prefer not to call into `Configuration::commit()` in the error case but this is good enough for now. I am merging the PR. Thank you very much!

PR #102, opened 2019-12-19
drivers/libblockfs: implement flock()

Matt8898:


bragi/proto/fs.proto
```diff
@@ -29,6 +29,13 @@ enum OpenFlags {
 	OF_NONBLOCK = 1;
 }
 
+enum FlockFlags {
+	LOCK_SH = 1;
+	LOCK_EX = 2;
+	LOCK_NB = 4;
+	LOCK_UN = 8;
+}
```
avdgrinten: I assume that this matches mlibc's constants?

drivers/libblockfs/src/ext2fs.hpp
```diff
@@ -228,6 +230,8 @@ struct Inode : std::enable_shared_from_this<Inode> {
 	struct timespec accessTime;
 	struct timespec dataModifyTime;
 	struct timespec anyChangeTime;
+	async::doorbell flock_notify;
+	std::vector<std::weak_ptr<protocols::fs::Flock<ext2fs::Inode> > > flocks;
```
avdgrinten: I don't like that we need to template `Flock` on the inode. Could we make a `FlockManager` class that stores `flock_notify` and `flocks` instead? Then the inode would store a `FlockManager`.

drivers/libblockfs/src/libblockfs.cpp
```diff
@@ -107,6 +176,7 @@ constexpr auto fileOperations = protocols::fs::FileOperations{}
 	.withSeekRel(&seekRel)
 	.withSeekEof(&seekEof)
 	.withRead(&read)
+	.withFlock(&flock)
 	.withWrite(&write)
 	.withReadEntries(&readEntries)
 	.withAccessMemory(&accessMemory)
```
avdgrinten: If we replace this by a designated initializer, we don't need to create the `withFlock` function (I want to get rid of those `withSomething` functions in the long term).

protocols/fs/include/protocols/fs/file-locks.hpp
```diff
@@ -0,0 +1,29 @@
...
+		Flock(FLockState t, std::weak_ptr<T> i) {
+			type = t;
+			inode = i;
+		}
+
+		~Flock() {
+			inode.lock()->flock_notify.ring();
+		}
+
+		Flock(const Flock &f2) {
```
avdgrinten: Why do we need this? Why not `= default`?

drivers/libblockfs/src/ext2fs.hpp
```diff
@@ -292,11 +298,13 @@ struct FileSystem {
 
 struct OpenFile {
 	OpenFile(std::shared_ptr<Inode> inode);
+	~OpenFile();
 
 	async::result<std::optional<std::string>> readEntries();
 
 	std::shared_ptr<Inode> inode;
 	uint64_t offset;
+	Flock* flock = nullptr;
```
avdgrinten: We can use `std::unique_ptr<Flock>` here to avoid manual memory management.

drivers/libblockfs/src/libblockfs.cpp
```diff
@@ -39,6 +40,72 @@ async::result<protocols::fs::SeekResult> seekEof(void *object, int64_t offset) {
 	co_return self->offset;
 }
 
+using FlockManager = protocols::fs::FlockManager;
+using Flock = protocols::fs::Flock;
+
+async::result<protocols::fs::Error> flock(void *object, managarm::fs::CntRequest req) {
```
avdgrinten: Instead of passing the entire `CntRequest`, it would be better to just pass the flags (e.g., similar to `read()` which also does not pass the full request).

drivers/libblockfs/src/libblockfs.cpp
```diff
@@ -39,6 +40,72 @@ async::result<protocols::fs::SeekResult> seekEof(void *object, int64_t offset) {
...
+	auto self = static_cast<ext2fs::OpenFile *>(object);
+	co_await self->inode->readyJump.async_wait();
+	auto inode = self->inode;
+	bool nonblock  = req.flock_flags() & managarm::fs::FlockFlags::LOCK_NB;
+	bool shared    = req.flock_flags() & managarm::fs::FlockFlags::LOCK_SH;
+	if((shared && (req.flock_flags() & managarm::fs::FlockFlags::LOCK_EX)) ||
+			(req.flock_flags() & managarm::fs::FlockFlags::LOCK_UN &&
+			 (shared || (req.flock_flags() & managarm::fs::FlockFlags::LOCK_EX)))) {
+		co_return protocols::fs::Error::illegalArguments;
+	}
```
avdgrinten: This `if` is hard to read. It would be good to break it up into multiple nested `if`s.

protocols/fs/include/protocols/fs/file-locks.hpp
```diff
@@ -0,0 +1,36 @@
...
+	enum FLockState {
+		LOCKED_EXCLUSIVE = 0,
+		LOCKED_SHARED = 1
+	};
+
+	struct FlockManager;
+
+	typedef boost::intrusive::list_base_hook<boost::intrusive::link_mode<boost::intrusive::auto_unlink> > auto_unlink_hook;
+	struct Flock : public boost::intrusive::list_base_hook<> {
+		FlockManager* list = nullptr;
```
avdgrinten: Why not call the member `manager`?

protocols/fs/include/protocols/fs/file-locks.hpp
```diff
@@ -0,0 +1,36 @@
...
+		FLockState type;
+
+		Flock(FLockState t, FlockManager* l) {
+			type = t;
+			list = l;
+		}
```
avdgrinten: This can use an initiializer list.

drivers/libblockfs/src/libblockfs.cpp
```diff
@@ -39,6 +40,72 @@ async::result<protocols::fs::SeekResult> seekEof(void *object, int64_t offset) {
...
+				self->flock = new_flock;
+				co_return protocols::fs::Error::none;
+			} else {
+				inode->flockManager.flocks.push_back(*new_flock);
+				self->flock = new_flock;
+				co_return protocols::fs::Error::none;
+			}
+		}
+	}
+	inode->flockManager.flocks.push_back(*new_flock);
```
avdgrinten: I think this loop could be moved to a `FlockManager::lock` function to avoid duplication once we also apply to in tmpfs.

avdgrinten:
Looks good now. Let's merge this.

PR #111, opened 2020-01-22
net: create netserver, add AF_INET support

ArsenArsen:
currently, netserver is integrated within the nic/virtio driver


drivers/nic/virtio/src/main.cpp
```diff
@@ -60,6 +62,73 @@ async::result<protocols::svrctl::Error> bindDevice(int64_t base_id) {
...
+		co_await header.async_wait();
+		HEL_CHECK(accept.error());
+		HEL_CHECK(recv_req.error());
+
+		auto conversation = accept.descriptor();
+
+		managarm::fs::CntRequest req;
+		req.ParseFromArray(recv_req.data(), recv_req.length());
+
+		if (req.req_type() == managarm::fs::CntReqType::CREATE_SOCKET) {
```
avdgrinten: There sould be a "illegal request" response if the server gets a different message type.

drivers/nic/virtio/src/main.cpp
```diff
@@ -60,6 +62,73 @@ async::result<protocols::svrctl::Error> bindDevice(int64_t base_id) {
...
+			helix::PushDescriptor push_socket;
+			auto [local_lane, remote_lane] = helix::createStream();
+			std::cout << "netserver: proto " << req.protocol()
+				  << " type " << req.type() << std::endl;
+
+			async::detach(protocols::fs::servePassthrough(
+						std::move(local_lane), nullptr, &fileOps));
+
+			auto transmit = helix::submitAsync(conversation, helix::Dispatcher::global(),
+					helix::action(&push_socket, remote_lane));
```
avdgrinten: We should also send a protobuf response (before sending the passthrough lane) containing an error code.

hel/include/hel.h
```diff
@@ -5,6 +5,7 @@
 #define HEL_H
 
 #include <stddef.h>
+#include <string.h>
```
avdgrinten: Does managarm-kernel still build with this change?

posix/subsystem/src/main.cpp
```diff
@@ -1508,6 +1510,13 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 			}else if(req.domain() == AF_NETLINK) {
 				assert(req.socktype() == SOCK_RAW || req.socktype() == SOCK_DGRAM);
 				file = nl_socket::createSocketFile(req.protocol());
+			} else if (req.domain() == AF_INET) {
+				assert(req.socktype() == SOCK_DGRAM);
```
avdgrinten: This should be an error response. (The other address families don't do this correctly either but we should not introduce more of these assertions.)

posix/subsystem/src/extern_socket.cpp
```diff
@@ -0,0 +1,47 @@
...
+			helix::action(&offer, kHelItemAncillary),
+			helix::action(&send_req, req_data.data(), req_data.size(), kHelItemChain),
+			helix::action(&recv_lane));
+	co_await transmit.async_wait();
+	HEL_CHECK(offer.error());
+	HEL_CHECK(send_req.error());
+	HEL_CHECK(recv_lane.error());
+
+	co_return File::constructHandle(
+			smarter::make_shared<Socket>(recv_lane.descriptor())
```
avdgrinten: Strange indentation.

posix/subsystem/src/extern_socket.cpp
```diff
@@ -0,0 +1,47 @@
+#include "extern_socket.hpp"
+
+#include "fs.pb.h"
+#include "protocols/fs/client.hpp"
+
+namespace {
+struct Socket : File {
+	Socket(helix::UniqueLane sock_lane)
+		: File { StructName::get("socket") },
```
avdgrinten: Strange indentation.

hel/include/hel.h
ArsenArsen: indeed it does (assuming `xbstrap install --rebuild managarm-kernel` does the right thing)

ArsenArsen:
squashed all commits

PR #123, opened 2020-02-18
thor: add support for store/loadregisters on VirtualizedCpuDescriptors

Matt8898:


hel/include/hel.h
```diff
@@ -112,6 +112,49 @@ enum {
 	kHelErrNoMemory = 17,
 };
 
+struct SegmentRegister {
```
avdgrinten: Should we prefix this (and the other structs) with the architecture? For example, `HelX86SegmentRegisters`?

kernel/thor/generic/virtualization.hpp
```diff
@@ -23,6 +23,8 @@ namespace thor {
 
 	struct VirtualizedCpu {
 			virtual HelVmexitReason run();
+			virtual void storeRegs(VirtualizationRegs *regs);
+			virtual VirtualizationRegs loadRegs();
```
avdgrinten: `storeRegs` takes a pointer, while `loadRegs` returns by value. Make this consistent: either use pointers/const pointers (probably a good idea due to the size of the struct) or take/return values in both functions.

kernel/thor/generic/hel.cpp
```diff
@@ -1452,16 +1452,21 @@ HelError helLoadRegisters(HelHandle handle, int set, void *image) {
 	auto this_universe = this_thread->getUniverse();
 
 	frigg::SharedPtr<Thread> thread;
+	VirtualizedCpuDescriptor vcpu{0};
```
avdgrinten: Shouldn't the default constructor do the right thing here? If it doesn't, should we add a proper default?

kernel/thor/generic/hel.cpp
```diff
@@ -1494,6 +1499,13 @@ HelError helLoadRegisters(HelHandle handle, int set, void *image) {
 		regs[0] = thread->_executor.general()->clientFs;
 		regs[1] = thread->_executor.general()->clientGs;
 		writeUserArray(reinterpret_cast<uintptr_t *>(image), regs, 2);
+	}else if(set == kHelRegsVirtualization){
```
avdgrinten: What if vcpu is nullptr here?

kernel/thor/generic/hel.cpp
```diff
@@ -1516,9 +1529,19 @@ HelError helStoreRegisters(HelHandle handle, int set, const void *image) {
 		auto thread_wrapper = this_universe->getDescriptor(universe_guard, handle);
 		if(!thread_wrapper)
 			return kHelErrNoDescriptor;
-		if(!thread_wrapper->is<ThreadDescriptor>())
+		if(thread_wrapper->is<ThreadDescriptor>()) {
+			if(set == kHelRegsVirtualization) {
```
avdgrinten: Instead of this check, I would prefer a check whether vcpu (or thread) is nullptr below.

hel/include/hel.h
avdgrinten: Is the struct portable to SVM in the future?

Matt8898: Looking at the VMCB structure this should be fine for both

kernel/thor/arch/x86/vmx.hpp
```diff
@@ -149,6 +149,8 @@ namespace thor::vmx {
 		Vmcs(const Vmcs& vmcs) = delete;
 		Vmcs& operator=(const Vmcs& vmcs) = delete;
 		HelVmexitReason run();
+		void storeRegs(VirtualizationRegs *regs);
```
avdgrinten: `const VirtualizationRegs *`?

avdgrinten:
LGTM now.

PR #127, opened 2020-03-22
posix/subsystem: implement clone supercall

qookei:


posix/subsystem/src/process.cpp
```diff
@@ -765,6 +765,60 @@ std::shared_ptr<Process> Process::fork(std::shared_ptr<Process> original) {
...
+			nullptr, 0, 0x1000, kHelMapProtRead | kHelMapProtWrite,
+			&process->_clientThreadPage));
+	HEL_CHECK(helMapMemory(process->_fileContext->fileTableMemory().getHandle(),
+			process->_vmContext->getSpace().getHandle(),
+			nullptr, 0, 0x1000, kHelMapProtRead,
+			&process->_clientFileTable));
+	HEL_CHECK(helMapMemory(clk::trackerPageMemory().getHandle(),
+			process->_vmContext->getSpace().getHandle(),
+			nullptr, 0, 0x1000, kHelMapProtRead,
+			&process->_clientClkTrackerPage));
```
avdgrinten: We should be able to just copy the `_clientFileTable` and `_clientClkTrackerPage` pointers. We do need an own `_clientThreadPage`, though.

PR #134, opened 2020-04-10
thor: integrate VirtualizedPageSpace with VirtualSpace

Matt8898:


kernel/thor/arch/x86/vmx.cpp
```diff
@@ -350,7 +350,36 @@ namespace thor::vmx {
 				frigg::infoLogger() << "vmx: hlt" << frigg::endLog;
 				exitInfo.exitReason = khelVmexitHlt;
 				return exitInfo;
-			} else if(reason == VMEXIT_EXTERNAL_INTERRUPT) {
+			} else if(reason == 48) {
```
avdgrinten, line 353: It looks strange that there is no symbolic constant for this.

kernel/thor/arch/x86/vmx.cpp
```diff
@@ -350,7 +350,36 @@ namespace thor::vmx {
...
+				frigg::UnsafePtr<Thread> this_thread = getCurrentThread();
+				size_t address = vmread(0x00002400);
```
avdgrinten, line 355: Do we use symbolic constants for `vmread()`?

kernel/thor/arch/x86/vmx.cpp
```diff
@@ -350,7 +350,36 @@ namespace thor::vmx {
...
+				frigg::infoLogger() << "vmx: ept violation at address: " << address << frigg::endLog;
+				WorkScope wqs{this_thread->pagingWorkQueue()};
```
avdgrinten, line 357: I think we don't need the scope here since this is always called in a syscall.

kernel/thor/arch/x86/vmx.cpp
```diff
@@ -350,7 +350,36 @@ namespace thor::vmx {
...
+				if(exitFlags & 1) {
+					flags |= AddressSpace::kFaultWrite;
+				}
+				if(exitFlags & (1 << 2)) {
+					flags |= AddressSpace::kFaultExecute;
+				}
+				if(!space->handleFault(address, flags, &closure.fault))
+					Thread::blockCurrent(&closure.blocker);
+
+				bool handled = closure.fault.resolved();
```
avdgrinten, line 381: What if `!handled`?

kernel/thor/generic/address-space.cpp
```diff
@@ -386,7 +386,7 @@ CowChain::~CowChain() {
 
 VirtualSpace::VirtualSpace(VirtualOperations *ops)
 : _ops{ops} {
-	auto hole = frigg::construct<Hole>(*kernelAlloc, 0x100000, 0x7ffffff00000);
+	auto hole = frigg::construct<Hole>(*kernelAlloc, 0, 0x7ffffff00000);
```
avdgrinten, line 389: For AddressSpace, it makes sense to forbid mapping pages to 0. To handle this difference between AddressSpace and the EPT, we could add a `setupInitialHole(VirtualAddr address, size_t size)` method to VirtualSpace that must be called after the constructor.

kernel/thor/generic/hel.cpp
```diff
@@ -740,9 +733,38 @@ HelError helMapMemory(HelHandle memory_handle, HelHandle space_handle,
...
+
+				if(flags & kHelMapProtRead)
+					map_flags |= AddressSpace::kMapProtRead;
+				if(flags & kHelMapProtWrite)
+					map_flags |= AddressSpace::kMapProtWrite;
+				if(flags & kHelMapProtExecute)
+					map_flags |= AddressSpace::kMapProtExecute;
+
+				if(flags & kHelMapDontRequireBacking)
+					map_flags |= AddressSpace::kMapDontRequireBacking;
```
avdgrinten, line 758: Isn't this exactly the same for AddressSpace and the EPT?

kernel/thor/arch/x86/vmx.cpp
avdgrinten: I guess we should return an error to userspace.

PR #135, opened 2020-04-11
eir: add Stivale/qloader2 target

thomtl:


kernel/eir/load64.S
```diff
@@ -0,0 +1,25 @@
...
+.section .text
+.global eirRtEnterKernel
+eirRtEnterKernel:
+    mov %rdi, %cr3
+    mov %rdx, %rsp
+
+    mov $0xC0000080, %rcx
+	rdmsr
+	or $(kEferNx), %eax
+	wrmsr
```
avdgrinten, line 19: Strange indentation.

kernel/eir/main.cpp
```diff
@@ -417,6 +417,8 @@ address_t mapBootstrapData(void *p) {
 
 // ----------------------------------------------------------------------------
 
+// We don't need GDT code when running in 64bit mode, in fact eirRtLoadGdt isn't even defined then
+#if UINTPTR_MAX == 0xFFFFFFFF
```
avdgrinten, line 421: I don't like this check. We could add a function `initializeCpu()`, implement it once in a `i386.cpp` and once in a `x86_64.cpp`. That would be much more generic also w.r.t. future ports to other archs/loaders. What do you think?

PR #137, opened 2020-04-11
posix: Added FD_SET_FLAGS

Dennisbonke:


posix/subsystem/src/main.cpp
```diff
@@ -1667,6 +1667,41 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+				managarm::posix::SvrResponse resp;
+				resp.set_error(managarm::posix::Errors::NO_SUCH_FD);
+
+				auto ser = resp.SerializeAsString();
+				auto &&transmit = helix::submitAsync(conversation, helix::Dispatcher::global(),
+						helix::action(&send_resp, ser.data(), ser.size()));
+				co_await transmit.async_wait();
+				HEL_CHECK(send_resp.error());
+				continue;
+			}
```
avdgrinten, line 1692: Here, we can use `sendErrorResponse()`.

posix/subsystem/src/main.cpp
```diff
@@ -1667,6 +1667,41 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+					std::cout << "posix: FD_SET_FLAGS unknown flags: " << req.flags() << std::endl;
+				}
+			}
+
+			managarm::posix::SvrResponse resp;
+			resp.set_error(managarm::posix::Errors::SUCCESS);
+
 			auto ser = resp.SerializeAsString();
 			auto &&transmit = helix::submitAsync(conversation, helix::Dispatcher::global(),
 					helix::action(&send_resp, ser.data(), ser.size()));
```
avdgrinten, line 1707: Here, we can use the new helix_ng API that was introduced last week:
```suggestion
				auto [resp_error, send_resp] = co_await helix_ng::exchangeMsgs(conversation, helix::Dispatcher::global(),
						helix_ng::sendBuffer(ser.data(), ser.size()));
				HEL_CHECK(resp_error);
```
This simplifies the code a bit. See this commit: 4bca53bd1c7bee865982ccab5caddaf8b057b454 for more example usages.

posix/subsystem/src/main.cpp
```diff
@@ -1667,6 +1667,41 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+				continue;
+			}
+
+			if(req.flags()) {
+				if(req.flags() & 1) {
+					descriptor->closeOnExec = !descriptor->closeOnExec;
+				} else {
+					std::cout << "posix: FD_SET_FLAGS unknown flags: " << req.flags() << std::endl;
+				}
+			}
```
avdgrinten, line 1700: This request should also be able to unset flags, right? Also: replace `1` by the symbolic constant.

avdgrinten: And consider returning EINVAL on invalid flags (in addition to the log message).

Dennisbonke: Yeah you're right, I think I messed it up. Leaves me wondering how one would check for the removal of the flag, can we just say if FD_CLOEXEC then set, otherwise unset?

Dennisbonke: About the symbolic constant, O_CLOEXEC doesn't work, so I probably need to define a new one then. Where are these stored? I'll implement the EINVAL response, good catch

avdgrinten:
Merged, will do some adjustments on `master`.

PR #142, opened 2020-04-16
thor: properly unmap pages and return PageStatus

Matt8898:


avdgrinten:
LGTM. Thank you for the quick fix!

PR #147, opened 2020-04-24
protocols/fs: Implemented pread()

Dennisbonke:


protocols/fs/fs.proto
```diff
@@ -98,6 +98,8 @@ enum CntReqType {
 	FLOCK = 73;
 
 	RENAME = 36;
+
+	PREAD = 37;
```
avdgrinten, line 102: Rename this request to `PT_PREAD` (I know others in this file don't follow the right convention; still, for new stuff, we should be consistent :smiley:).

PR #152, opened 2020-04-28
posix: implement named pipes

qookei:


posix/subsystem/src/fifo.cpp
```diff
@@ -178,6 +195,58 @@ struct WriterFile : File {
...
 
+// This maps FsNodes to Channels for named pipes (FIFOs)
+std::map<FsNode *, std::shared_ptr<Channel>> globalChannelMap;
+
+void createNamedChannel(FsNode *node) {
+	assert(globalChannelMap.find(node) == globalChannelMap.end());
+	globalChannelMap[node] = std::make_shared<Channel>();
+}
+
+void unrefNamedChannel(FsNode *node) {
```
avdgrinten, line 206: Let's call this `unlinkNamedChannel`, `unref` sounds as-if it decreases a refcount.

PR #154, opened 2020-05-07
Various small fixes for gcc and itstool

Dennisbonke:


avdgrinten:
Can we do the same error handle also for `pread()` (where we have the same assertion)?

Dennisbonke:
> Can we do the same error handle also for `pread()` (where we have the same assertion)?

This is fixed in 6458b6a

posix/subsystem/src/tmp_fs.cpp
```diff
@@ -333,6 +333,13 @@ struct MemoryNode final : Node {
 		co_return File::constructHandle(std::move(file));
 	}
 
+	FutureMaybe<FileStats> getStats() override {
+		std::cout << "\e[31mposix: Fix tmpfs getStats() in MemoryNode\e[39m" << std::endl;
+		FileStats stats{};
+		stats.fileSize = _fileSize;
+		co_return stats;
```
avdgrinten, line 340: Now we return the correct fileSize but we do not return the correct inode number anymore. That should be corrected. Maybe we could also store the other variables from `FileStats` in the `MemoryNode` struct and report them (while initializing them to zero for now?).

posix/subsystem/src/tmp_fs.cpp
```diff
@@ -333,6 +333,24 @@ struct MemoryNode final : Node {
 		co_return File::constructHandle(std::move(file));
 	}
 
+	FutureMaybe<FileStats> getStats() override {
+		std::cout << "\e[31mposix: Fix tmpfs getStats() in MemoryNode\e[39m" << std::endl;
+		FileStats stats{};
+		stats.inodeNumber = 0;
```
avdgrinten, line 339: `inodeNumber` can be set (see `Node::getStats()`; we probably need to add a public function to access it but that shouldn't be a problem).

For the other fields: I actually meant that we should add private fields (e.g., `uid_t _uid = 0;`) for them. I guess that these can actually be added to `Node` and not `MemoryNode` since they apply to all node types (except for the file size).

posix/subsystem/src/tmp_fs.cpp
```diff
@@ -32,11 +32,66 @@ struct Node : FsNode {
 		std::cout << "\e[31mposix: Fix tmpfs getStats()\e[39m" << std::endl;
 		FileStats stats{};
 		stats.inodeNumber = _inodeNumber;
+		stats.fileSize = _fileSize;
```
avdgrinten, line 35: The file size should not be maintained by `Node` since that depends on the node type (regular file vs. directory etc.).

posix/subsystem/src/tmp_fs.cpp
```diff
@@ -32,11 +32,66 @@ struct Node : FsNode {
...
+		return _mtime;
+	}
+
+	timespec ctime() {
+		return _ctime;
+	}
+
 private:
-	int64_t _inodeNumber;
+	int64_t _inodeNumber = 0;
```
avdgrinten, line 86: This is initialized by the constructor anyway, right? (I think we can drop the = 0).

posix/subsystem/src/tmp_fs.cpp
Dennisbonke: can I just return 0 there then?

PR #155, opened 2020-05-10
posix: Implement get/set uid/euid/gid/egid

Dennisbonke:
This PR implements get and set for user-id and related friends.
So far, the following has been implemented in at least a basic form

- `getuid()`
- `setuid()`, needs more work, as we do not check against effective user-id and friends. This should now be fixed and more POSIX compliant
- `geteuid()`
- `seteuid()`
- `getgid()`
- `setgid()`, doesn't exactly match linux in behaviour, probably `setgid` bit or `CAP_SETGID` capability.
- `getegid()`
- `setegid()`, doesn't exactly match linux in behaviour, probably `setgid` bit or `CAP_SETGID` capability.

### DIFF
--- a/posix/subsystem/src/process.cpp
+++ b/posix/subsystem/src/process.cpp
@@ -632,7 +632,7 @@ std::shared_ptr<Process> Process::findProcess(ProcessId pid) {
 }
 
 Process::Process(Process *parent)
-: _parent{parent}, _pid{0}, _clientPosixLane{kHelNullHandle}, _clientFileTable{nullptr},
+: _parent{parent}, _pid{0}, /*_uid{0},*/ _clientPosixLane{kHelNullHandle}, _clientFileTable{nullptr},
 		_notifyType{NotifyType::null} { }
 
 Process::~Process() {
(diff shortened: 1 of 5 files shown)


posix/subsystem/src/process.hpp
```diff
@@ -341,6 +341,18 @@ struct Process : std::enable_shared_from_this<Process> {
 	 	return _pid;
 	}
 
+	int setUid(int uid) {
```
avdgrinten, line 344: This should return the `Error` enum.

posix/subsystem/src/process.cpp
```diff
@@ -632,7 +632,7 @@ std::shared_ptr<Process> Process::findProcess(ProcessId pid) {
 }
 
 Process::Process(Process *parent)
-: _parent{parent}, _pid{0}, _clientPosixLane{kHelNullHandle}, _clientFileTable{nullptr},
+: _parent{parent}, _pid{0}, /*_uid{0},*/ _clientPosixLane{kHelNullHandle}, _clientFileTable{nullptr},
```
avdgrinten, line 635: Why the comment?

Dennisbonke: This is a leftover from debugging code, it will be removed

Dennisbonke:
I think this part of users/groups/permissions is done, so should be good for review

PR #156, opened 2020-05-10
Store allocation logs in a periodically flushed ring buffer

qookei:


kernel/eir/main.cpp
```diff
@@ -512,6 +519,7 @@ void initProcessorPaging(void *kernel_start, uint64_t& kernel_entry){
 		mapSingle4kPage(addr, addr, kAccessWrite | kAccessExecute);
 
 	mapRegionsAndStructs();
+	allocLogRingBuffer();
```
avdgrinten, line 522: Can we avoid this when trace support is compiled out? We're "wasting" quite a bit of physical RAM here ;D.

kernel/memory.txt
```diff
@@ -11,16 +11,21 @@ Thor uses the following PHYSICAL memory regions:
...
-	Length: (unlimited)
+FFFF'E000'0000'0000
+	Length: 1 GiB
 	Kernel heap
 	Referenced in thor/src/core.cpp
 
+FFFF'F000'0000'0000
+	Length: 256MiB
+	Log ring buffer memory
+	Referenced in eir/main.cpp
```
avdgrinten, line 27: Also referenced somewhere in thor.

kernel/thor/generic/kerncfg.cpp
```diff
@@ -56,6 +59,58 @@ coroutine<Error> handleReq(LaneHandle boundLane) {
...
+	managarm::kerncfg::CntRequest<KernelAlloc> req(*kernelAlloc);
+	req.ParseFromArray(reqBuffer.data(), reqBuffer.size());
+
+	if(req.req_type() == managarm::kerncfg::CntReqType::GET_BUFFER_CONTENTS) {
+		size_t old_dequeue = req.dequeue();
+		size_t wanted_size = req.size();
+		size_t actual_size = frg::min(allocLog->newDataSize(old_dequeue), wanted_size);
+
+		constexpr uint64_t nanos = 100000000;
+		constexpr size_t min_size = 65536; // 64K
```
avdgrinten, line 78: Can we make the `min_size` independent of the dequeue size? For example, let userspace wait until 1 MiB is available while dequeuing only 64KiB?

Also: camelCase.

kernel/thor/generic/ring-buffer.hpp
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	void enqueue(char c) {
+		stor_[enqueue_ & (size_ - 1)] = c;
+		enqueue_++;
+	}
+
+	size_t dequeueInto(void *buffer, size_t dequeue, size_t size) {
+		size_t i = 0;
+		while (i < size) {
```
avdgrinten, line 21: What if the user requests more data than what is available? We should adjust this code to handle that correctly (and return a partial result).

avdgrinten: For this to work, the function also needs to return the number of bytes read (e.g. as a `frg::tuple<size_t, size_t>`).

drivers/tty/virtio-console/src/console.cpp
```diff
@@ -4,6 +4,64 @@
...
+
+		kerncfgByteRingLane = helix::UniqueLane(co_await entity.bind());
+		foundKerncfgByteRing.trigger();
+	});
+
+	co_await root.linkObserver(std::move(filter), std::move(handler));
+	co_await foundKerncfgByteRing.async_wait();
+}
+
+async::result<size_t> getKerncfgByteRingPart(arch::dma_buffer &chunk, uint64_t &dequeue) {
```
avdgrinten, line 36: This can take a `arch::dma_buffer_view` instead of a ref to a buffer. I'd be more happy with a tuple return type that also returns the new dequeue pointer instead of using a reference hre.

drivers/tty/virtio-console/src/console.cpp
```diff
@@ -25,19 +83,23 @@ async::detached Device::runDevice() {
...
-	chain.append(co_await txQueue_->obtainDescriptor());
-	chain.setupBuffer(virtio_core::hostToDevice, buffer);
+	constexpr size_t chunk_size = 64 * 1024;
+	arch::dma_buffer chunk_buffer{&dmaPool_, chunk_size};
 
-	std::cout << "virtio-console: Submitting" << std::endl;
-	co_await txQueue_->submitDescriptor(chain.front());
-	std::cout << "virtio-console: Done" << std::endl;
+	while (true) {
+		size_t size = co_await getKerncfgByteRingPart(chunk_buffer, dequeue);
```
avdgrinten, line 94: Report an error to stderr if data was lost?

PR #157, opened 2020-05-11
Implemented chmod

Dennisbonke:
This PR implements a `FCHMODAT` request in posix, allowing the libc functions `chmod()`, `fchmod()` and `fchmodat()` to function.

### DIFF
--- a/drivers/libblockfs/src/ext2fs.cpp
+++ b/drivers/libblockfs/src/ext2fs.cpp
@@ -263,18 +263,14 @@ async::result<std::optional<DirEntry>> Inode::mkdir(std::string name) {
 	co_return co_await link(name, dir_node->number, kTypeDirectory);
 }
 
-async::result<int> Inode::chmod(int mode, int64_t ino) {
-	assert(ino);
+async::result<int> Inode::chmod(int mode) {
+	co_await readyJump.async_wait();
 
-	co_await this->readyJump.async_wait();
-
-	auto self = fs.accessInode(ino);
-
-	self->diskInode()->mode = (self->diskInode()->mode & 0xFFFFF000) | mode;
+	diskInode()->mode = (diskInode()->mode & 0xFFFFF000) | mode;
 
 	auto syncInode = co_await helix_ng::synchronizeSpace(
 			helix::BorrowedDescriptor{kHelNullHandle},
-			self->diskMapping.get(), fs.inodeSize);
+			diskMapping.get(), fs.inodeSize);
 	HEL_CHECK(syncInode.error());
 
 	co_return 0;
@@ -618,7 +614,7 @@ async::detached FileSystem::initiateInode(std::shared_ptr<Inode> inode) {
 
 	// filter out the file type from the mode
 	// TODO: ext2fs stores a 32-bit mode
-	inode->mode = disk_inode->mode & 0x0FFF;
+	inode->diskInode()->mode = disk_inode->mode & 0x0FFF;
 
 	inode->numLinks = disk_inode->linksCount;
 	// TODO: support large uid / gids
--- a/drivers/libblockfs/src/ext2fs.cpp
+++ b/drivers/libblockfs/src/ext2fs.cpp
@@ -263,7 +263,7 @@ async::result<std::optional<DirEntry>> Inode::mkdir(std::string name) {
 	co_return co_await link(name, dir_node->number, kTypeDirectory);
 }
 
-async::result<int> Inode::chmod(int mode) {
+async::result<protocols::fs::Error> Inode::chmod(int mode) {
 	co_await readyJump.async_wait();
 
 	diskInode()->mode = (diskInode()->mode & 0xFFFFF000) | mode;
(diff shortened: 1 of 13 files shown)


drivers/libblockfs/src/libblockfs.cpp
```diff
@@ -274,7 +274,7 @@ getStats(std::shared_ptr<void> object) {
 	protocols::fs::FileStats stats;
 	stats.linkCount = self->numLinks;
 	stats.fileSize = self->fileSize();
-	stats.mode = self->mode;
+	stats.mode = self->diskInode()->mode;
```
avdgrinten, line 277: We should remove `self->mode` entirely, in favor of `diskInode()->mode`.

avdgrinten: Also, when returning the mode here, we should mask out the file type.

posix/subsystem/src/fs.cpp
```diff
@@ -83,6 +83,10 @@ DeviceId FsNode::readDevice() {
 	throw std::runtime_error("readDevice() is not implemented for this FsNode");
 }
 
+async::result<int> FsNode::chmod(int mode) {
+	throw std::runtime_error("chmod() is not implemented for this FsNode");
+}
```
avdgrinten, line 88: Hm, we should return a `async::result<Error>` and handle this error gracefully. Probably by returning the equivalent of EPERM.

posix/subsystem/src/main.cpp
```diff
@@ -1476,6 +1476,68 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+					continue;
+				}
+
+				relative_to = {file->associatedMount(), file->associatedLink()};
+			}
+
+			if(req.flags()) {
+				if(req.flags() & AT_SYMLINK_NOFOLLOW) {
+					co_await sendErrorResponse(managarm::posix::Errors::NOT_SUPPORTED);
+					continue;
```
avdgrinten, line 1510: What does this translate to on the client side?

Dennisbonke: ENOTSUP, which is the error returned in this case according to the man page

posix/subsystem/src/fs.cpp
Dennisbonke: Personally, I'd do ENOTSUP, as it is unsupported

drivers/libblockfs/src/ext2fs.cpp
```diff
@@ -601,7 +614,7 @@ async::detached FileSystem::initiateInode(std::shared_ptr<Inode> inode) {
 
 	// filter out the file type from the mode
 	// TODO: ext2fs stores a 32-bit mode
-	inode->mode = disk_inode->mode & 0x0FFF;
+	inode->diskInode()->mode = disk_inode->mode & 0x0FFF;
```
avdgrinten, line 617: This line is not needed. (Before the change, it copied the mode to a member variable. Now it copies the mode to itself and discards the file type -- that's problematic.)

posix/subsystem/src/extern_fs.cpp
```diff
@@ -74,6 +74,32 @@ struct Node : FsNode {
...
+
+		managarm::fs::CntRequest req;
+		req.set_req_type(managarm::fs::CntReqType::NODE_CHMOD);
+		req.set_mode(mode);
+
+		auto ser = req.SerializeAsString();
+		auto &&transmit = helix::submitAsync(getLane(), helix::Dispatcher::global(),
+				helix::action(&offer, kHelItemAncillary),
+				helix::action(&send_req, ser.data(), ser.size(), kHelItemChain),
+				helix::action(&recv_resp));
```
avdgrinten, line 90: Use helix_ng here.

posix/subsystem/src/fs.cpp
```diff
@@ -83,6 +83,10 @@ DeviceId FsNode::readDevice() {
 	throw std::runtime_error("readDevice() is not implemented for this FsNode");
 }
 
+async::result<Error> FsNode::chmod(int mode) {
+	co_return Error::accessDenied;
```
avdgrinten, line 87: This change is good. But we should maybe still log about this.

posix/subsystem/src/main.cpp
avdgrinten: OK!

protocols/fs/include/protocols/fs/server.hpp
```diff
@@ -199,6 +199,8 @@ struct NodeOperations {
 	async::result<std::string> (*readSymlink)(std::shared_ptr<void> object);
 
 	async::result<MkdirResult> (*mkdir)(std::shared_ptr<void> object, std::string name);
+
+	async::result<int> (*chmod)(std::shared_ptr<void> object, int mode);
```
avdgrinten, line 203: Does protocols/fs have an `Error` enum? If yes, return that instead of int.

qookei:
LGTM

PR #180, opened 2020-05-28
posix: Fix a bug and implement RMDIR

Dennisbonke:
We should return ILLEGAL_ARGUMENTS when flags is more then the two specified, not when either of those is declared.

### DIFF
--- a/posix/subsystem/src/main.cpp
+++ b/posix/subsystem/src/main.cpp
@@ -1239,7 +1239,7 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 			helix::SendBuffer send_resp;
 			managarm::posix::SvrResponse resp;
 
-			if(!(req.flags() & ~(AT_EMPTY_PATH | AT_SYMLINK_FOLLOW))) {
+			if(!(req.flags() & (AT_EMPTY_PATH | AT_SYMLINK_FOLLOW))) {
 				co_await sendErrorResponse(managarm::posix::Errors::ILLEGAL_ARGUMENTS);
 				continue;
 			}
--- a/posix/subsystem/src/main.cpp
+++ b/posix/subsystem/src/main.cpp
@@ -1239,7 +1239,7 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 			helix::SendBuffer send_resp;
 			managarm::posix::SvrResponse resp;
 
-			if(!(req.flags() & (AT_EMPTY_PATH | AT_SYMLINK_FOLLOW))) {
+			if(req.flags() & ~(AT_EMPTY_PATH | AT_SYMLINK_FOLLOW)) {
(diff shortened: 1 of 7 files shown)


Dennisbonke:
The latest commit adds a basic RMDIR implementation, it doesn't respect the fact that a directory removed by `rmdir` must be empty before removing

posix/subsystem/src/main.cpp
```diff
@@ -2133,6 +2133,36 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+			co_await resolver.resolve();
+
+			target_link = resolver.currentLink();
+
+			if(!target_link) {
+				co_await sendErrorResponse(managarm::posix::Errors::FILE_NOT_FOUND);
+				continue;
+			} else {
+				auto owner = target_link->getOwner();
+				co_await owner->unlink(target_link->getName());
```
avdgrinten, line 2156: I think we should add a rmdir operation here (to `FsNode`, let it return a `async::result<Error>`), wire it up to extern_fs and only translate to unlink in libblockfs.

Dennisbonke: I believe the new commit implements this request

posix/subsystem/src/main.cpp
```diff
@@ -1239,7 +1239,7 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 			helix::SendBuffer send_resp;
 			managarm::posix::SvrResponse resp;
 
-			if(!(req.flags() & ~(AT_EMPTY_PATH | AT_SYMLINK_FOLLOW))) {
+			if(!(req.flags() & (AT_EMPTY_PATH | AT_SYMLINK_FOLLOW))) {
```
qookei, line 1242: Shouldn't this actually be `if (req.flags() & ~(AT_EMPTY_PATH | AT_SYMLINK_FOLLOW))`? As I see it, we are checking here whether we've been passed invalid flags and not whether we haven't been passed these proper flags.

protocols/fs/src/server.cpp
```diff
@@ -795,6 +795,19 @@ async::detached serveNode(helix::UniqueLane lane, std::shared_ptr<void> node,
...
 
+			auto ser = resp.SerializeAsString();
+			auto [send_resp] = co_await helix_ng::exchangeMsgs(
+				conversation,
+				helix_ng::sendBuffer(ser.data(), ser.size())
+			);
+			HEL_CHECK(send_resp.error());
+		}else if(req.req_type() == managarm::fs::CntReqType::NODE_RMDIR) {
+			// TODO: This should probably be it's own operation, for now, let it be
+			co_await node_ops->unlink(node, req.path());
```
qookei, line 806: I'd prefer this to be properly implemented before merging, unless it's a big problem, in which case it should be fine for now.

qookei: As discussed on discord, we will leave this as is for now.

PR #184, opened 2020-06-06
thor: Use x2apic if available

thomtl:


kernel/thor/arch/x86/pic.cpp
```diff
@@ -227,12 +249,24 @@ void initLocalApicOnTheSystem() {
...
+	picBase = ApicRegisterSpace(haveX2apic, register_ptr);
 
 	frigg::infoLogger() << "Booting on CPU #" << getLocalApicId() << frigg::endLog;
 }
 
 void initLocalApicPerCpu() {
+	uint64_t msr = frigg::arch_x86::rdmsr(frigg::arch_x86::kMsrLocalApicBase);
+	msr |= (1 << 11); // Enable APIC
+
+	if(frigg::arch_x86::cpuid(0x01)[2] & (uint32_t(1) << 21)){
```
avdgrinten, line 261: Here, we can just look at picBase to determine if the x2APIC is supported (and panic if it isn't). That would prevent the code from trying to use x2APICs on CPUs where they are not available.

kernel/thor/arch/x86/pic.hpp
```diff
@@ -5,12 +5,53 @@
...
+			auto msr = x2apic_msr_base + (r.offset() >> 4);
+			return static_cast<typename RT::rep_type>(frigg::arch_x86::rdmsr(msr));
+		} else {
+			auto p = reinterpret_cast<const typename RT::bits_type *>(_mem_base + r.offset());
+			auto b = arch::mem_ops<typename RT::bits_type>::load(p);
+			return static_cast<typename RT::rep_type>(b);
+		}
+	}
+
+	bool _x2apic;
```
avdgrinten, line 49: This is public, remove the underscore (or make it private + provide an accessor).

PR #188, opened 2020-06-09
Refactored some FsNode operations

Dennisbonke:
As requested on Discord, I have started refactoring some `FsNode` operations to use `frg::expected`. For now, I assume that nothing will throw errors, but this assumption may be wrong.
I have tested the changes and they seem to work, as in managarm boots into weston, although I fully expect there to be bugs and/or issues in this pr.

### DIFF
--- a/posix/subsystem/src/main.cpp
+++ b/posix/subsystem/src/main.cpp
@@ -1493,7 +1493,9 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 				continue;
 			}
 
-			auto stats = co_await target_link->getTarget()->getStats();
+			auto statsResult = co_await target_link->getTarget()->getStats();
+			assert(statsResult);
+			auto stats = statsResult.value();
 
 			managarm::posix::SvrResponse resp;
 			resp.set_error(managarm::posix::Errors::SUCCESS);
--- a/posix/subsystem/src/main.cpp
+++ b/posix/subsystem/src/main.cpp
@@ -1249,7 +1249,8 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 				continue;
 			}
 
-			co_await parent->mkfifo(resolver.nextComponent(), req.mode());
+			auto result = co_await parent->mkfifo(resolver.nextComponent(), req.mode());
+			assert(result);
 
 			co_await sendErrorResponse(managarm::posix::Errors::SUCCESS);
 		}else if(req.request_type() == managarm::posix::CntReqType::LINKAT) {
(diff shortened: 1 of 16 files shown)


posix/subsystem/src/main.cpp
```diff
@@ -1249,7 +1249,8 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 				continue;
 			}
 
-			co_await parent->mkfifo(resolver.nextComponent(), req.mode());
+			auto result = co_await parent->mkfifo(resolver.nextComponent(), req.mode());
+			assert(result);
```
avdgrinten, line 1253: In `serveRequests()`, instead of `assert()`ing, we could log a message (something like `posix: Unexpected failure from mkfifo()`) and `co_return`. That would make the client process unusable (since we stop processing requests for it) but avoid a crash of posix-subsystem.

posix/subsystem/src/main.cpp
```diff
@@ -1325,7 +1326,8 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 			auto target = resolver.currentLink()->getTarget();
 			auto directory = new_resolver.currentLink()->getTarget();
 			assert(target->superblock() == directory->superblock()); // Hard links across mount points are not allowed, return EXDEV
-			co_await directory->link(new_resolver.nextComponent(), target);
+			auto result = co_await directory->link(new_resolver.nextComponent(), target);
+			assert(result);
```
avdgrinten, line 1330: Dito.

posix/subsystem/src/main.cpp
```diff
@@ -2216,7 +2218,8 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 				HEL_CHECK(send_resp.error());
 			} else {
 				auto owner = target_link->getOwner();
-				co_await owner->unlink(target_link->getName());
+				auto result = co_await owner->unlink(target_link->getName());
+				assert(result);
```
avdgrinten, line 2222: Dito.

posix/subsystem/src/main.cpp
```diff
@@ -2247,7 +2250,8 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 				continue;
 			} else {
 				auto owner = target_link->getOwner();
-				co_await owner->rmdir(target_link->getName());
+				auto result = co_await owner->rmdir(target_link->getName());
+				assert(result);
```
avdgrinten, line 2254: Dito.

PR #190, opened 2020-06-17
posix: Implement truncate for ext2 and tmp_fs

Dennisbonke:


posix/subsystem/src/extern_fs.cpp
```diff
@@ -250,6 +273,8 @@ struct RegularNode final : Node {
 
 		auto file = smarter::make_shared<OpenFile>(pull_ctrl.descriptor(),
 				pull_passthrough.descriptor(), std::move(mount), std::move(link));
+		if(semantic_flags & semanticTruncate)
+			co_await file->truncate(0);
```
avdgrinten, line 277: We could do this in the OPEN/OPENAT handling code in `main.cpp`. That way, we don't need the `semanticTruncate` flag and we do not need to add code to each file system.

PR #192, opened 2020-06-24
drivers: Handle failure in unlink gracefully

Dennisbonke:


drivers/libblockfs/src/libblockfs.cpp
```diff
@@ -443,24 +447,41 @@ async::detached servePartition(helix::UniqueLane lane) {
 			auto newInode = fs->accessInode(req.inode_target());
 
 			auto old_file = co_await oldInode->findEntry(req.old_name());
+			managarm::fs::SvrResponse resp;
 			if(old_file) {
 				if(co_await newInode->findEntry(req.new_name()) != std::nullopt) {
-					co_await newInode->unlink(req.new_name());
+					auto result = co_await newInode->unlink(req.new_name());
+					if(!result) {
```
avdgrinten, line 454: Since `newInode->unlink` returns an error now, we can just try to unlink the new name and ignore the `FILE_NOT_FOUND` error. That way, we do not need `findEntry` anymore.

drivers/libblockfs/src/libblockfs.cpp
```diff
@@ -443,24 +447,41 @@ async::detached servePartition(helix::UniqueLane lane) {
...
 				auto [send_resp] = co_await helix_ng::exchangeMsgs(conversation,
 					helix_ng::sendBuffer(ser.data(), ser.size()));
 				HEL_CHECK(send_resp.error());
+				continue;
 			}
 
-			co_await oldInode->unlink(req.old_name());
+			auto result = co_await oldInode->unlink(req.old_name());
+			if(!result) {
+				resp.set_error(managarm::fs::Errors::FILE_NOT_FOUND);
```
avdgrinten, line 477: We should probably assert that `result.error` is indeed a "file not found" error.

The same holds for other calls of unlink.

PR #200, opened 2020-07-15
kernel/eir: begin working on arm port

qookei:


kernel/eir/arch/arm/load64.S
```diff
@@ -0,0 +1,31 @@
+.section .bss
+.align 16
+.global eirRtStackTop
```
avdgrinten, line 3: For symbols that are not referenced by generic code, we can drop `eirRt` in favor of an `eir` prefix. (I want to drop it everywhere but this requires some more renaming of functions.)

PR #203, opened 2020-07-20
posix: Implement non-blocking unix sockets

Dennisbonke:


posix/subsystem/src/un-socket.cpp
```diff
@@ -66,12 +66,14 @@ struct OpenFile : File {
 				smarter::shared_ptr<File>{file}, &File::fileOperations, file->_cancelServe));
 	}
 
-	OpenFile(Process *process = nullptr)
+	OpenFile(Process *process = nullptr, int flags = 0)
```
avdgrinten, line 69: Pass a `bool nonBlock` directly.

posix/subsystem/src/un-socket.cpp
```diff
@@ -359,6 +367,20 @@ struct OpenFile : File {
 		return _passthrough;
 	}
 
+	async::result<void> setFileFlags(int flags) override {
```
avdgrinten, line 370: Print a message if any other flags are specified.

posix/subsystem/src/un-socket.cpp
```diff
@@ -359,6 +367,22 @@ struct OpenFile : File {
...
+		if(flags & O_NONBLOCK)
+			nonBlock_ = true;
+		else {
+			std::cout << "posix: setFileFlags on socket \e[1;34m" << structName() << "\e[0m called with unknown flags" << std::endl;
```
avdgrinten, line 374: This message should be printed if `flags & ~O_NONBLOCK`, i.e., when a flag *other than* O_NONBLOCK is set. Right now it's also printed when O_NONBLOCK is *not* set, regardless of whether other flags are passed, too.

posix/subsystem/src/un-socket.cpp
```diff
@@ -66,12 +66,14 @@ struct OpenFile : File {
...
-	OpenFile(Process *process = nullptr)
+	OpenFile(Process *process = nullptr, bool nonBlock = false)
 	: File{StructName::get("un-socket"), File::defaultPipeLikeSeek}, _currentState{State::null},
 			_currentSeq{1}, _inSeq{0}, _ownerPid{0},
-			_remote{nullptr}, _passCreds{false} {
+			_remote{nullptr}, _passCreds{false}, nonBlock_{false} {
 		if(process)
 			_ownerPid = process->pid();
+		if(nonBlock)
+			nonBlock_ = nonBlock;
```
avdgrinten, line 76: This can be simplified to `nonBlock_{nonBlock}` ;)

posix/subsystem/src/file.cpp
```diff
@@ -150,6 +150,11 @@ async::result<void> File::ptSetFileFlags(void *object, int flags) {
 	return self->setFileFlags(flags);
 }
 
+async::result<size_t> File::ptPeername(void *object, void *addr_ptr, size_t max_addr_length) {
```
avdgrinten, line 153: Use `frg::expected` here to handle errors.

posix/subsystem/src/file.cpp
```diff
@@ -356,3 +361,9 @@ async::result<void> File::setFileFlags(int flags) {
 	co_return;
 }
 
+async::result<size_t> File::peername(void *addr_ptr, size_t max_addr_length) {
```
avdgrinten, line 364: Also use `frg::expected` here.

posix/subsystem/src/un-socket.cpp
```diff
@@ -69,11 +69,10 @@ struct OpenFile : File {
 	OpenFile(Process *process = nullptr, bool nonBlock = false)
 	: File{StructName::get("un-socket"), File::defaultPipeLikeSeek}, _currentState{State::null},
 			_currentSeq{1}, _inSeq{0}, _ownerPid{0},
-			_remote{nullptr}, _passCreds{false}, nonBlock_{false} {
+			_remote{nullptr}, _passCreds{false}, nonBlock_{nonBlock},
+			_sockpath{'\0'} {
```
avdgrinten, line 73: This should not make a difference and should still be debugged.

posix/subsystem/src/un-socket.cpp
```diff
@@ -295,11 +295,15 @@ struct OpenFile : File {
 
 		if(sa.sun_path[0] == '\0') {
 			path.resize(addr_length - sizeof(sa.sun_family) - 1);
+			_sockpath.resize(path.size());
```
avdgrinten, line 298: Instead of resize + memcpy in both cases, just set `_sockpath = path;` after the if.

posix/subsystem/src/un-socket.cpp
```diff
@@ -368,20 +372,42 @@ struct OpenFile : File {
...
 			co_return O_NONBLOCK;
 		co_return 0;
-}
+	}
+
+	async::result<size_t> peername(void *addr_ptr, size_t max_addr_length) override {
+		struct sockaddr_un sa;
+		memset(&sa, 0, sizeof(struct sockaddr_un));
+		sa.sun_family = AF_UNIX;
+		strcpy(sa.sun_path, this->_remote->_sockpath.c_str());
```
avdgrinten, line 396: This needs a `strncpy` - `sockaddr_un` has only limited space for the path.

posix/subsystem/src/un-socket.cpp
```diff
@@ -368,20 +372,42 @@ struct OpenFile : File {
...
+		memcpy(addr_ptr, &sa, std::min(sizeof(struct sockaddr_un), max_addr_length));
+
+		co_return _remote->_sockpath.size() + sizeof(sa.sun_family);
+	}
+
+	async::result<size_t> sockname(void *addr_ptr, size_t max_addr_length) override {
+		struct sockaddr_un sa;
+		memset(&sa, 0, sizeof(struct sockaddr_un));
+		sa.sun_family = AF_UNIX;
+		strcpy(sa.sun_path, this->_sockpath.c_str());
```
avdgrinten, line 406: Dito.

posix/subsystem/src/un-socket.cpp
```diff
@@ -368,20 +372,42 @@ struct OpenFile : File {
...
 	}
 
 	async::result<int> getFileFlags() override {
 		if(nonBlock_)
 			co_return O_NONBLOCK;
 		co_return 0;
-}
+	}
+
+	async::result<size_t> peername(void *addr_ptr, size_t max_addr_length) override {
```
avdgrinten, line 392: sockname and peername also need to handle abstract sockets, by prepending a \0 to the sun_path.

protocols/fs/src/server.cpp
```diff
@@ -370,6 +370,25 @@ async::detached handlePassthrough(smarter::shared_ptr<void> file,
...
+			helix_ng::sendBuffer(ser.data(), ser.size()),
+			helix_ng::sendBuffer(addr.data(),
+					std::min(size_t(req.size()), actual_length))
+		);
+		HEL_CHECK(send_resp.error());
+		HEL_CHECK(send_data.error());
+	}else if(req.req_type() == managarm::fs::CntReqType::PT_PEERNAME) {
+		std::vector<char> addr;
+		addr.resize(req.size());
+		assert(file_ops->peername);
```
avdgrinten, line 385: We should handle this more gracefully (by sending an error to the client). I know that existing cases don't do that, but we should not add more cases that need to be fixed later.

posix/subsystem/src/un-socket.cpp
Dennisbonke: Agreed, I can add a comment to the code indicating this, or do you want this resolved before merging?

PR #204, opened 2020-07-27
kernel/thor: begin work on arm port

qookei:


kernel/thor/arch/arm/system.cpp
```diff
@@ -0,0 +1,8 @@
+#include <thor-internal/arch/system.hpp>
+
+namespace thor {
+
+void initializeArchitecture() {}
+void initializeExtendedSystem() {}
```
avdgrinten, line 6: This function has been removed in the meantime.

kernel/thor/arch/arm/thor-internal/arch/ints.hpp
```diff
@@ -0,0 +1,123 @@
...
+	// TODO
+}
+
+inline void halt() {
+	asm volatile ("wfi");
+}
+
+void suspendSelf();
+
+struct IrqMutex {
```
avdgrinten, line 31: This class (and `StatelessIrqLock`) can be moved to generic code, maybe to a new `kernel-locks.hpp`.

kernel/thor/system/pci/pci_discover.cpp
```diff
@@ -2,7 +2,10 @@
 #include <frigg/debug.hpp>
 #include <hw.frigg_pb.hpp>
 #include <mbus.frigg_pb.hpp>
+#ifdef __x86_64__
+// ???
```
avdgrinten, line 6: ???

PR #215, opened 2020-08-27
protocols/fs: Turn assertions into error responses.

### LINKED ISSUE #209: fs: Turn assertions into error responses
The fs protocol uses assertions to check if the functions are available (like `setFileFlags()`), instead of asserting, and thus crashing posix if they are not implemented, we should send an error response indicating it isn't implemented

Dennisbonke:
This PR turns assertions into error responses in the fs code.

This PR closes #209


avdgrinten:
The documentation changes should go into their own PR.

avdgrinten:
Hmm, we should probably add an `ILLEGAL_OPERATION_TARGET` error for this purpose. `ILLEGAL_REQUEST` means that the request in general is malformed (e.g., could not be parsed).

PR #216, opened 2020-08-27
protocols/fs: Change assertions into error returns

### LINKED ISSUE #209: fs: Turn assertions into error responses

Dennisbonke:
This PR turns assertions into error responses in the fs code.

This PR closes #209


avdgrinten (approved):
LGTM.

PR #218, opened 2020-08-28
posix: Various additions to tmpfs

### LINKED ISSUE #217: posix: tmp_fs needs fixes in various error paths
- In `link()`, `unlink()`, `rename()`, `mkdir()`, `symlink()`, `mkdev()` and `mkfifo()` we assert that entries do not (or do) exist. We should return an error instead (`mksocket()` is correct).
- In `readSome()`, we have another assertion that should return 0 instead (and no error).

Dennisbonke:
This PR aims to implement `seek` and `rmdir` for the tmpfs `DirectoryNode` and `DirectoryFile`.
Python uses `rmdir()` and `rewinddir()` on directories in `/tmp` when running `python3 get-pip.py`, this PR implements the required functionality.
Furthermore, this PR changes various assertions in tmpfs to proper error returns. As such, this PR closes #217

### DIFF
--- a/posix/subsystem/src/main.cpp
+++ b/posix/subsystem/src/main.cpp
@@ -1536,8 +1536,9 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 			auto superblock = resolver.currentLink()->getTarget()->superblock();
 			auto directory = new_resolver.currentLink()->getTarget();
 			assert(superblock == directory->superblock());
-			co_await superblock->rename(resolver.currentLink().get(),
+			auto result = co_await superblock->rename(resolver.currentLink().get(),
 					directory.get(), new_resolver.nextComponent());
+			assert(result);
 
 			co_await sendErrorResponse(managarm::posix::Errors::SUCCESS);
 		}else if(preamble.id() == managarm::posix::FstatAtRequest::message_id) {
--- a/posix/subsystem/src/main.cpp
+++ b/posix/subsystem/src/main.cpp
@@ -1538,7 +1538,11 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 			assert(superblock == directory->superblock());
 			auto result = co_await superblock->rename(resolver.currentLink().get(),
 					directory.get(), new_resolver.nextComponent());
-			assert(result);
+			if(!result) {
+				assert(result.error() == Error::alreadyExists);
+				co_await sendErrorResponse(managarm::posix::Errors::ALREADY_EXISTS);
+				continue;
+			}
 
 			co_await sendErrorResponse(managarm::posix::Errors::SUCCESS);
 		}else if(preamble.id() == managarm::posix::FstatAtRequest::message_id) {
--- a/posix/subsystem/src/tmp_fs.cpp
+++ b/posix/subsystem/src/tmp_fs.cpp
@@ -237,6 +237,8 @@ public:
 
 	explicit DirectoryFile(std::shared_ptr<MountView> mount, std::shared_ptr<FsLink> link);
 
+	async::result<frg::expected<Error, off_t>> seek(off_t delta, VfsSeek whence) override;
+
 	void handleClose() override;
 
 	FutureMaybe<ReadEntriesResult> readEntries() override;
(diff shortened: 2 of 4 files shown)


posix/subsystem/src/tmp_fs.cpp
```diff
@@ -264,7 +264,7 @@ struct DirectoryNode final : Node, std::enable_shared_from_this<DirectoryNode> {
 
 private:
 	VfsType getType() override {
-		return VfsType::directory;
+  		return VfsType::directory;
```
avdgrinten, line 267: Strange indentation change.

posix/subsystem/src/tmp_fs.cpp
```diff
@@ -292,7 +292,8 @@ struct DirectoryNode final : Node, std::enable_shared_from_this<DirectoryNode> {
 
 	async::result<frg::expected<Error, std::shared_ptr<FsLink>>> link(std::string name,
 			std::shared_ptr<FsNode> target) override {
-		assert(_entries.find(name) == _entries.end());
+		if(!(_entries.find(name) == _entries.end()))
```
avdgrinten, line 295: We can use != here (same below).

posix/subsystem/src/tmp_fs.cpp
```diff
@@ -487,7 +489,9 @@ struct Superblock final : FsSuperblock {
 
 		auto src_dir = static_cast<DirectoryNode *>(src_link->getOwner().get());
 		auto it = src_dir->_entries.find(src_link->getName());
-		assert(it != src_dir->_entries.end() && it->get() == src_link);
+		// TODO: This should really return an error, frg::expected style.
+		if(it == src_dir->_entries.end() || it->get() != src_link)
+			co_return nullptr;
```
avdgrinten, line 494: We can do the conversion to `frg::expected` in this PR?

posix/subsystem/src/tmp_fs.cpp
Dennisbonke: This is the same as `mksocket()` uses, according to #217 that was correct

posix/subsystem/src/tmp_fs.cpp
Dennisbonke: Yes we can, done in 0699a50

posix/subsystem/src/main.cpp
```diff
@@ -1536,8 +1536,9 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 			auto superblock = resolver.currentLink()->getTarget()->superblock();
 			auto directory = new_resolver.currentLink()->getTarget();
 			assert(superblock == directory->superblock());
-			co_await superblock->rename(resolver.currentLink().get(),
+			auto result = co_await superblock->rename(resolver.currentLink().get(),
 					directory.get(), new_resolver.nextComponent());
+			assert(result);
```
avdgrinten, line 1541: Send an error to the client here.

posix/subsystem/src/tmp_fs.cpp
```diff
@@ -307,7 +310,8 @@ struct DirectoryNode final : Node, std::enable_shared_from_this<DirectoryNode> {
 
 	async::result<frg::expected<Error>> unlink(std::string name) override {
 		auto it = _entries.find(name);
-		assert(it != _entries.end());
+		if(it == _entries.end())
+			co_return Error::alreadyExists;
```
avdgrinten, line 314: The error should be `noSuchFile` here, right? (In contrast to mkdir() and friends, we are checking whether an entry *exists* and not whether it does not exist.)

avdgrinten (commented):
Since the seek part is dropped from the PR, please also drop the corresponding commit via rebase.

posix/subsystem/src/tmp_fs.cpp
```diff
@@ -316,6 +320,15 @@ struct DirectoryNode final : Node, std::enable_shared_from_this<DirectoryNode> {
 
 	async::result<frg::expected<Error, std::shared_ptr<FsLink>>> mksocket(std::string name) override;
 
+	async::result<frg::expected<Error>> rmdir(std::string name) override {
+		auto result = co_await unlink(name);
+		if(!result) {
+			std::cout << "tmpfs: Unexpected failure from unlink()" << std::endl;
+			co_return {};
```
avdgrinten, line 327: Shouldn't we return an error here?

PR #225, opened 2020-10-18
posix: Fix broken conditional in execute()

### LINKED ISSUE #224: thor: Panic in AddressSpace destructor code path
There are certain files, mostly shell scripts like `/usr/bin/nroff` and perl scripts like `/usr/bin/autoreconf` that panic frigg on execution.
The following is printed in the e9 log

`automake`
```
posix: Failed to read executable: /bin/automake
frg: Panic! ../../../src/managarm/subprojects/frigg/include/frg/optional.hpp:141: Assertion '_non_n
ll' failed!
```

`nroff`
```
posix: Failed to read executable: /bin/nroff
frg: Panic! ../../../src/managarm/subprojects/frigg/include/frg/optional.hpp:141: Assertion '_non_n
ll' failed!
```

Even a custom script that's only
```
#! /bin/bash

echo "lmaoyeet"
```
fails when executed as `lmaoyeet`. Were one to execute it as `bash /usr/bin/lmaoyeet` it would work fine. Same goes for `nroff` when executed as `bash /usr/bin/nroff` or `automake` when executed as `perl /usr/bin/automake` or whatever the scripts shebang may be

Dennisbonke:
This PR fixes a trigger of #224, but doesn't address the primary cause of this issue

### DIFF
--- a/posix/subsystem/src/exec.cpp
+++ b/posix/subsystem/src/exec.cpp
@@ -165,7 +165,7 @@ execute(ViewPath root, ViewPath workdir,
 
 			char buffer[128];
 			auto readResult = co_await execFile->readSome(nullptr, buffer, 128);
-			if(readResult) {
+			if(!readResult) {
 				std::cout << "posix: Failed to read executable" << std::endl;
 				co_return Error::badExecutable;
 			}


avdgrinten (approved):
LGTM.

PR #226, opened 2020-10-26
libblockfs: Small fixes for building under gcc-10

Dennisbonke:


avdgrinten (approved):
LGTM. In the long run, we should convert the fs protocol library to use `frg::expected` instead of `std::variant` but until that happens, the casts are fine.

Issue #205, opened 2020-07-31
thor: Boot failure due to ACPI error


Dennisbonke:
When trying to boot managarm on real hardware, I got the following error
![noacpi](https://user-images.githubusercontent.com/8709509/89034293-f4fbe680-d338-11ea-92a9-318e69b1256e.jpg)

This was produced on a laptop with an intel i7-8750H, on a NL5C motherboard (manufacturer Quanta)

Dennisbonke:
Some progress has been made, but the possible fix is stuck in testing until qword-os/lai#87 is resolved

avdgrinten:
Does thor detect ACPI now? If yes, we can close this and track progress in the LAI issue.

Dennisbonke:
Yes, acpi seems to be detected iirc. I'll retest that soon and close this issue if it is.

avdgrinten:
Closing this since it is solved.

PR #228, opened 2020-11-08
Add support for multiboot 2 handing us the acpi tables

Dennisbonke:
This PR adds support for the multiboot 2 acpi tags and uses it where available, lai is still used as a fallback.

Part of a series of fixes for #205

### DIFF
--- a/kernel/eir/arch/x86/multiboot2/multiboot2.cpp
+++ b/kernel/eir/arch/x86/multiboot2/multiboot2.cpp
@@ -4,6 +4,7 @@
 #include <eir-internal/arch.hpp>
 #include <eir-internal/generic.hpp>
 #include <eir-internal/debug.hpp>
+#include <acpispec/tables.h>
 
 namespace eir {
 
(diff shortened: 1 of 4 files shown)


kernel/eir/arch/x86/multiboot2/multiboot2.cpp
```diff
@@ -197,6 +215,28 @@ extern "C" void eirMultiboot2Main(uint32_t info, uint32_t magic){
 
 				break;
 			}
+
+			case kMb2TagAcpiOld: {
```
avdgrinten, line 219: If the structs are named `v1` and `v2`, why do we call the tags `Old`and `New`? Does this match the spec?

kernel/eir/arch/x86/multiboot2/multiboot2.cpp
```diff
@@ -213,6 +253,8 @@ extern "C" void eirMultiboot2Main(uint32_t info, uint32_t magic){
 	assert(cmdline.data()); // Make sure it at least exists
 	assert(framebuffer);
 
+	assert(rsdt);
+
```
avdgrinten, line 257: Is this necessary? Doesn't thor already complain when no ACPI table is found?

avdgrinten (commented):
Looks mostly good, some nitpicks.

avdgrinten:
LGTM now.

PR #229, opened 2020-11-20
ext2: Fix link counts

Dennisbonke:
This PR addresses several small bugs regarding error handling in the code for `isatty()` and link counts in the ext2 driver.

### DIFF
--- a/drivers/libblockfs/src/ext2fs.cpp
+++ b/drivers/libblockfs/src/ext2fs.cpp
@@ -737,7 +737,6 @@ async::detached FileSystem::initiateInode(std::shared_ptr<Inode> inode) {
 		abort();
 	}
 
-	inode->numLinks = disk_inode->linksCount;
 	// TODO: support large uid / gids
 	inode->uid = disk_inode->uid;
 	inode->gid = disk_inode->gid;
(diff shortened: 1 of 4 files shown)


drivers/libblockfs/src/libblockfs.cpp
```diff
@@ -290,7 +290,7 @@ getStats(std::shared_ptr<void> object) {
 	co_await self->readyJump.async_wait();
 
 	protocols::fs::FileStats stats;
-	stats.linkCount = self->numLinks;
+	stats.linkCount = self->diskInode()->linksCount;
```
avdgrinten, line 293: This is the right change but please remove the `numLinks` variable entirely instead of only changing it here.

drivers/libblockfs/src/ext2fs.cpp
```diff
@@ -656,6 +656,7 @@ async::result<std::shared_ptr<Inode>> FileSystem::createSymlink() {
 	disk_inode->atime = time.tv_sec;
 	disk_inode->ctime = time.tv_sec;
 	disk_inode->mtime = time.tv_sec;
+	disk_inode->linksCount++;
```
avdgrinten, line 659: Why is this change done?

avdgrinten (approved):
LGTM.

Issue #84, opened 2019-07-14
thor: Scheduler should support any number of threads


avdgrinten:
Due to the usage of fixed-point numbers, the number of concurrently running threads is limited to (1 << 6) = 64 threads. This needs to be fixed, probably by using more bits to store the scheduler's progress.

PR #232, opened 2020-12-06
Fixed issue #84 thor: Scheduler should support any number of threads

parthpatel1470:


kernel/thor/generic/schedule.cpp
```diff
@@ -5,7 +5,7 @@
 #include <thor-internal/core.hpp>
 #include <thor-internal/schedule.hpp>
 #include <thor-internal/timer.hpp>
-#ifdef __x86_64__
+#ifdef __x86_128__
```
avdgrinten, line 8: This should stay `__x86_64__`.

kernel/thor/generic/schedule.cpp
```diff
@@ -143,16 +143,16 @@ int64_t Scheduler::_liveRuntime(const ScheduleEntity *entity) {
 void Scheduler::update() {
 	// Returns the reciprocal in 0.8 fixed point format.
 	auto fixedInverse = [] (uint32_t x) -> uint32_t {
-		assert(x < (1 << 6));
-		return static_cast<uint32_t>(1 << 8) / x;
+		assert(x < (1 << 7));
+		return static_cast<uint32_t>(1 << 9) / x;
```
avdgrinten, line 147: This is a change from `0.8` fixed point (8 bits after the decimal point) to `0.9`. I would like to extend this even further, maybe to `0.32` (and maybe up to (1 << 24) tasks?). Note that the comments that mention `0.8` format should also be changed.

qookei:
Closing this due to inactivity

PR #234, opened 2020-12-20
Several non-blocking related fixes in POSIX and netserver

Dennisbonke:
This PR adds the ability to a FIFO pipe reader side to be non-blocking and implements `getFileFlags()` and `setFileFlags()` for a FIFO pipe reader side and tcp sockets. For the tcp sockets, this is a partial implementation, only working for `O_NONBLOCK`


avdgrinten (approved):
LGTM.

PR #241, opened 2020-12-30
posix: Implement pipe2 with O_NONBLOCK

Dennisbonke:
This PR should get rid of the message `posix: pipe2(O_NONBLOCK) is not implemented correctly` which is shown on boot.


avdgrinten (approved):
LGTM.

PR #243, opened 2021-01-11
posix: Fix an off by one bug in the signal handling code

Dennisbonke:
This PR fixes a signal handling bug found by the posix_spawn implementation found in managarm/mlibc#201.

This PR is not really blocked on managarm/mlibc#201, but should be merged together with it to avoid breaking changes


avdgrinten (approved):
LGTM.

PR #245, opened 2021-01-30
posix: Mark all ioctls on writer side fifo's as invalid

Dennisbonke:
This PR fixes certain applications crashing on calling ioctl on a writer side fifo.

After this, one should merge managarm/mlibc#224


avdgrinten (commented):
We should handle this for all files, not only for FIFOs (in file.cpp). Also, the default implementation should send an ILLEGAL_OPERATION_TARGET.

Dennisbonke:
Good point, I’ll modify it when I’m back home

avdgrinten (approved):
LGTM.

PR #248, opened 2021-02-19
Initial work on mknod

### LINKED ISSUE #237: posix: Implement mknod
With the necessary boilerplate in mlibc, we should implement the functionality required for `mknod()` in posix and the underlying file systems. This would allow the `coreutils` program `mknod` and the `runit` programs to function.

Dennisbonke:
This PR adds the infrastructure for mknod and various special files that it can create. It also adds error handling for finding devices in the device registry and adds O_PATH handling to the open request. This PR has (a partial) implementation for `/dev/zero`, `/dev/null` and `/dev/full`. Furthermore, a lot of refactoring with the write request has been done in order to be able to pass errors all the way back up to mlibc.

After this PR, managarm/mlibc#234 should be merged.
Fixes #237


posix/subsystem/src/main.cpp
```diff
@@ -2190,10 +2193,25 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 				}
 
 				auto target = resolver.currentLink()->getTarget();
-				auto fileResult = co_await target->open(resolver.currentView(), resolver.currentLink(),
-									semantic_flags);
-				assert(fileResult);
-				file = fileResult.value();
+
+				if(req->flags() & managarm::posix::OpenFlags::OF_PATH) {
+					auto fileTemp = smarter::make_shared<DummyFile>(resolver.currentView(), resolver.currentLink());
```
avdgrinten, line 2198: `fileTemp` -> `dummyFile`.

posix/subsystem/src/file.hpp
```diff
@@ -325,4 +328,27 @@ struct File : private smarter::crtp_counter<File, DisposeFileHandle> {
 	bool _isOpen;
 };
 
+struct DummyFile final : File {
+public:
+	static void serve(smarter::shared_ptr<DummyFile> file) {
+//TODO:		assert(!file->_passthrough);
+
```
avdgrinten, line 335: We can remove this comment.

posix/subsystem/src/file.hpp
```diff
@@ -325,4 +328,27 @@ struct File : private smarter::crtp_counter<File, DisposeFileHandle> {
...
+		helix::UniqueLane lane;
+		std::tie(lane, file->_passthrough) = helix::createStream();
+		async::detach(protocols::fs::servePassthrough(std::move(lane),
+				file, &fileOperations, file->_cancelServe));
+	}
+
+	DummyFile(std::shared_ptr<MountView> mount, std::shared_ptr<FsLink> link, DefaultOps default_ops = 0)
+	: File{StructName::get("DummyFile"), std::move(mount), std::move(link)} { }
```
avdgrinten, line 343: Isn't the convention for StructName all lower case with hyphens?

posix/subsystem/src/file.hpp
```diff
@@ -50,7 +50,10 @@ enum class Error {
 
 	notConnected,
 
-	alreadyExists
+	alreadyExists,
+
+	// Corresponds with ENXIO
+	specialDevice
```
avdgrinten, line 56: Rename this to noBackingDevice or something similar. specialDevice doesn't really sound like an error.

posix/subsystem/src/devices/null.cpp
```diff
@@ -0,0 +1,82 @@
...
+#include <experimental/coroutine>
+
+namespace {
+
+struct NullFile final : File {
+private:
+	async::result<frg::expected<Error, size_t>>
+	readSome(Process *, void *data, size_t max_length) override {
+		(void)data;
+		(void)max_length;
```
avdgrinten, line 15: You can drop the argument names from the definition to avoid the (void) casts.

posix/subsystem/src/main.cpp
```diff
@@ -3221,6 +3239,117 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+			} else if(S_ISSOCK(req->mode())) {
+				type = VfsType::socket;
+			} else {
+				// TODO: Add error handling for this case
+				type = VfsType::null;
+			}
+			dev.first = major(req->device());
+			dev.second = minor(req->device());
+
+			auto result = co_await parent->mkdev(resolver.nextComponent(), type, dev);
```
avdgrinten, line 3341: `mkdev` only works for char + block devices. For the other file types, we need to call other functions, such as mkdir. Some file types (such as sockets) are probably entirely forbidden.

Dennisbonke: Correct, `mknod` may only be used to make regular files, character special files, block special files, FIFO pipes or unix domain sockets. On linux, using `mknod` to make a directory is not allowed. I'll need to implement that.

posix/subsystem/src/devices/full.cpp
```diff
@@ -0,0 +1,69 @@
...
+	static void serve(smarter::shared_ptr<FullFile> file) {
+
+		helix::UniqueLane lane;
+		std::tie(lane, file->_passthrough) = helix::createStream();
+		async::detach(protocols::fs::servePassthrough(std::move(lane),
+				file, &fileOperations, file->_cancelServe));
+	}
+
+	FullFile(std::shared_ptr<MountView> mount, std::shared_ptr<FsLink> link)
+	: File{StructName::get("full"), std::move(mount), std::move(link)} { }
```
avdgrinten, line 42: Call the StructName `full-file` (similarly for the others).

posix/subsystem/src/devices/full.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+struct FullFile final : File {
+private:
+	async::result<frg::expected<Error, size_t>>
+	readSome(Process *, void *, size_t) override {
+		co_return 0;
+	}
+
+	async::result<frg::expected<Error, size_t>> writeAll(Process *, const void *, size_t) override {
+		co_return 0;
```
avdgrinten, line 18: Shouldn't writing to /dev/full return an error?

```
$ echo foo > /dev/full
bash: echo: write error: No space left on device
```

posix/subsystem/src/devices/full.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include <experimental/coroutine>
+
+namespace {
+
+struct FullFile final : File {
+private:
+	async::result<frg::expected<Error, size_t>>
+	readSome(Process *, void *, size_t) override {
+		co_return 0;
+	}
```
avdgrinten, line 15: What does Linux do when reading from /dev/full?

posix/subsystem/src/devices/zero.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include <experimental/coroutine>
+
+namespace {
+
+struct ZeroFile final : File {
+private:
+	async::result<frg::expected<Error, size_t>>
+	readSome(Process *, void *, size_t) override {
+		co_return 0;
+	}
```
avdgrinten, line 15: This should memset the input buffer to zero and return the number of bytes written.

posix/subsystem/src/devices/zero.cpp
```diff
@@ -0,0 +1,69 @@
...
+		return _passthrough;
+	}
+
+	helix::UniqueLane _passthrough;
+	async::cancellation_event _cancelServe;
+
+public:
+	static void serve(smarter::shared_ptr<ZeroFile> file) {
+
+		helix::UniqueLane lane;
```
avdgrinten, line 35: Unnecessary empty line.

avdgrinten (commented):
Looks very good but some things are not entirely done yet.

posix/subsystem/src/devices/null.cpp
```diff
@@ -0,0 +1,68 @@
...
+struct NullFile final : File {
+private:
+	async::result<frg::expected<Error, size_t>>
+	readSome(Process *, void *, size_t) override {
+		co_return 0;
+	}
+
+	async::result<frg::expected<Error, size_t>> writeAll(Process *, const void *, size_t) override {
+		co_return 0;
+	}
```
avdgrinten, line 19: Shouldn't we return the size here / pretend that all data was written successfully?

posix/subsystem/src/devices/zero.cpp
```diff
@@ -0,0 +1,69 @@
...
+private:
+	async::result<frg::expected<Error, size_t>>
+	readSome(Process *, void *data, size_t length) override {
+		memset(data, 0, length);
+		co_return length;
+	}
+
+	async::result<frg::expected<Error, size_t>> writeAll(Process *, const void *, size_t) override {
+		co_return 0;
+	}
```
avdgrinten, line 20: What about this function? Doesn't the same apply as for /dev/null?

avdgrinten (approved):

avdgrinten:
Thanks again for the PR!

Issue #235, opened 2020-12-30
posix: Implement sync()


Dennisbonke:
Now that the necessary boilerplate in mlibc is implemented, we should implement the functionality required for `sync()` in posix. This would allow the `coreutils` program `sync` to function.

PR #250, opened 2021-02-25
posix: Implement various process group related requests

Dennisbonke:
This PR will implement various requests related to process groups.

The following requests have been added:

- `getpgid`,
- `setpgid`,
- `setsid`,
- `getsid`.

Furthermore, the following `ioctls` have been implemented:

- `TIOCSCTTY` on the master tty,
- `TIOCGPGRP` on both the master and slave tty,
- `TIOCSPGRP` on both the master and slave tty,
- `TIOCGSID` on both the master and slave tty.

Finally, improved error handling was added in multiple places, tests for the new functionality was added and a duplicate number issue was resolved.

This PR has a mlibc counterpart (managarm/mlibc#235) that should be merged after this PR.


posix/subsystem/src/main.cpp
```diff
@@ -3252,6 +3252,65 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+			if (!req) {
+				std::cout << "posix: Rejecting request due to decoding failure" << std::endl;
+				break;
+			}
+
+			if(logRequests)
+				std::cout << "posix: SET_PGID" << std::endl;
+
+			std::shared_ptr<Process> target;
+			if(req->pid() != 0 && req->pid() != self->pid()) {
```
avdgrinten, line 3273: `if(foo != 0)` checks can be replaced just by `if(foo)`. Same for `if(foo != nullptr)`.

posix/subsystem/src/main.cpp
```diff
@@ -3252,6 +3252,65 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+				if(!target) {
+					co_await sendErrorResponse(managarm::posix::Errors::NO_SUCH_RESOURCE);
+					continue;
+				}
+			} else {
+				target = self;
+			}
+
+			pid_t pid;
+			if(req->pid() == 0) {
```
avdgrinten, line 3284: Same here.

posix/subsystem/src/main.cpp
```diff
@@ -3252,6 +3252,65 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+				pid = self->pid();
+			} else {
+				pid = req->pid();
+			}
+
+			if(req->pgid() == 0) {
```
avdgrinten, line 3290: Same here.

posix/subsystem/src/main.cpp
```diff
@@ -3252,6 +3252,65 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+			} else {
+				target = self;
+			}
+
+			pid_t pid;
+			if(req->pid() == 0) {
+				pid = self->pid();
+			} else {
+				pid = req->pid();
+			}
```
avdgrinten, line 3288: Move the PID calculation into the if branch where it is actually needed.

posix/subsystem/src/process.hpp
```diff
@@ -595,6 +595,8 @@ struct ProcessGroup : std::enable_shared_from_this<ProcessGroup> {
 		return hull_.get();
 	}
 
+	std::shared_ptr<TerminalSession> getSession() { return sessionPointer_; }
```
avdgrinten, line 598: This can return a `TerminalSession *` (obtained via `.get()`). Reason: the shared_ptr would sure that the `TerminalSession` does not get deallocated but since the session of a `ProcessGroup` can never change and `ProcessGroup` already owns a shared_ptr to the `TerminalSession`, that cannot happen anyway.

If we ever need to turn the plain pointer into a shared_ptr, we can use enable_shared_from_this.

posix/subsystem/src/main.cpp
```diff
@@ -3221,6 +3221,95 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+			if (!req) {
+				std::cout << "posix: Rejecting request due to decoding failure" << std::endl;
+				break;
+			}
+
+			if(logRequests)
+				std::cout << "posix: GET_PGID" << std::endl;
+
+			std::shared_ptr<Process> target;
+			if(req->pid() && req->pid() != self->pid()) {
```
avdgrinten, line 3242: The second half of the `if` is not really necessary (findProcess() can just return `self` as well).

posix/subsystem/src/main.cpp
```diff
@@ -3221,6 +3221,95 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+			if (!req) {
+				std::cout << "posix: Rejecting request due to decoding failure" << std::endl;
+				break;
+			}
+
+			if(logRequests)
+				std::cout << "posix: SET_PGID" << std::endl;
+
+			std::shared_ptr<Process> target;
+			if(req->pid() && req->pid() != self->pid()) {
```
avdgrinten, line 3273: Same here.

posix/subsystem/src/main.cpp
```diff
@@ -3221,6 +3221,95 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+				if(!group) {
+					pid_t pid;
+					if(req->pid() == 0) {
+						pid = self->pid();
+					} else {
+						pid = req->pid();
+					}
+					// Not found, making it if pgid and pid match?
+					if(pid == req->pgid()) {
+						target->pgPointer()->getSession()->spawnProcessGroup(target.get());
```
avdgrinten, line 3298: Where is this behavior specified? Linux' man page says:
> If pgid is zero, then the PGID of the process specified by pid is made the same as its process ID.
but that case is already implemented above.

Dennisbonke: To be honest I'm not entirely sure what I did here anymore and it's probably wrong. I'd need to take another look at it.

posix/subsystem/src/main.cpp
```diff
@@ -3402,6 +3402,37 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+				std::cout << "posix: Rejecting request due to decoding failure" << std::endl;
+				break;
+			}
+
+			if(logRequests)
+				std::cout << "posix: GET_PGID" << std::endl;
+
+			std::shared_ptr<Process> target;
+			if(req->pid() != 0 && req->pid() != self->pid()) {
+				target = Process::findProcess(req->pid());
```
avdgrinten, line 3437: We need to return EPERM if `target` is not in our session here (it's probably easiest to compare the session pointer both target and self).

posix/subsystem/src/main.cpp
```diff
@@ -3420,19 +3420,73 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
 			std::shared_ptr<Process> target;
-			if(req->pid() != 0 && req->pid() != self->pid()) {
+			if(req->pid() && req->pid() != self->pid()) {
 				target = Process::findProcess(req->pid());
 				if(!target) {
 					co_await sendErrorResponse(managarm::posix::Errors::NO_SUCH_RESOURCE);
 					continue;
 				}
-			} else {
-				target = self;
```
avdgrinten: Huh? Doesn't that mean that `target` is never set in this case?

posix/subsystem/src/main.cpp
```diff
@@ -3420,19 +3420,73 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+		}else if(preamble.id() == managarm::posix::SetPgidRequest::message_id) {
+			auto req = bragi::parse_head_only<managarm::posix::SetPgidRequest>(recv_head);
+
+			if (!req) {
+				std::cout << "posix: Rejecting request due to decoding failure" << std::endl;
+				break;
+			}
+
+			if(logRequests)
+				std::cout << "posix: SET_PGID" << std::endl;
```
avdgrinten, line 3465: Typo, this is SetPgid.

posix/subsystem/src/main.cpp
```diff
@@ -3420,19 +3420,73 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+
+			std::shared_ptr<Process> target;
+			if(req->pid() && req->pid() != self->pid()) {
+				target = Process::findProcess(req->pid());
+				if(!target) {
+					co_await sendErrorResponse(managarm::posix::Errors::NO_SUCH_RESOURCE);
+					continue;
+				}
+			}
+
```
avdgrinten, line 3482: There is no else here, and thus `target` is not always set.

posix/subsystem/src/main.cpp
```diff
@@ -3420,19 +3420,73 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+				std::cout << "posix: Rejecting request due to decoding failure" << std::endl;
+				break;
+			}
+
+			if(logRequests)
+				std::cout << "posix: SET_PGID" << std::endl;
+
+			std::shared_ptr<Process> target;
+			if(req->pid() && req->pid() != self->pid()) {
+				target = Process::findProcess(req->pid());
```
avdgrinten, line 3469: We have to return EPERM here if target is not in the same session.

posix/subsystem/src/main.cpp
avdgrinten: Also, we have to error out with EPERM if `target` is a session leader (we can check for that by comparing its PID to its session ID). This has to happen regardless of how we obtain `target` (whether it's `self` or not).

posix/subsystem/src/main.cpp
```diff
@@ -3420,19 +3420,73 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+				if(!target) {
+					co_await sendErrorResponse(managarm::posix::Errors::NO_SUCH_RESOURCE);
+					continue;
+				}
+			}
+
+			if(!req->pgid()) {
+				// Make a new process group with pgid == pid
+				target->pgPointer()->getSession()->spawnProcessGroup(target.get());
```
avdgrinten, line 3485: What if there is already a process group with PGID equal to `target`'s PID before? I think we should move the `if(!group)` part out of the else branch of this if to implement this.

posix/subsystem/src/main.cpp
avdgrinten: The Linux man page also mentions the case that `target` is a child of the current process, but setpgid() is only valid on such children if they did not execve() yet. We could add a `didExecute` flag to `Process` to implement that.

posix/subsystem/src/pts.cpp
```diff
@@ -689,6 +689,46 @@ async::result<void> SlaveFile::ioctl(Process *process, managarm::fs::CntRequest
...
+
+		auto [extractCreds] = co_await helix_ng::exchangeMsgs(
+			conversation,
+			helix_ng::extractCredentials()
+		);
+		HEL_CHECK(extractCreds.error());
+
+		auto process = findProcessWithCredentials(extractCreds.credentials());
+
+		resp.set_pid(process->pgPointer()->getHull()->getPid());
```
avdgrinten, line 709: This return the PGID of the current process, not the PGID of the foreground process group of the terminal.

posix/subsystem/src/pts.cpp
```diff
@@ -689,6 +689,46 @@ async::result<void> SlaveFile::ioctl(Process *process, managarm::fs::CntRequest
...
+
+		auto [extractCreds] = co_await helix_ng::exchangeMsgs(
+			conversation,
+			helix_ng::extractCredentials()
+		);
+		HEL_CHECK(extractCreds.error());
+
+		auto process = findProcessWithCredentials(extractCreds.credentials());
+
+		process->pgPointer()->getSession()->setForegroundGroup(process->pgPointer()->getSession()->getProcessGroupById(req.pgid()).get());
```
avdgrinten: This operates on `process->pgPointer()->getSession()`, but it should operate on the session associated with the terminal.

posix/subsystem/src/pts.cpp
avdgrinten: We also need to perform error checking as specified for tcgetpgrp()

posix/subsystem/src/pts.cpp
avdgrinten: Also, this needs to perform error checking.

posix/subsystem/src/pts.cpp
```diff
@@ -729,6 +729,27 @@ async::result<void> SlaveFile::ioctl(Process *process, managarm::fs::CntRequest
...
+
+		auto [extractCreds] = co_await helix_ng::exchangeMsgs(
+			conversation,
+			helix_ng::extractCredentials()
+		);
+		HEL_CHECK(extractCreds.error());
+
+		auto process = findProcessWithCredentials(extractCreds.credentials());
+
+		resp.set_pid(process->pgPointer()->getSession()->getSessionId());
```
avdgrinten, line 757: This operates on the current process but it needs to operate on the session associated with the terminal.

posix/subsystem/src/process.cpp
```diff
@@ -1123,9 +1123,14 @@ void TerminalSession::dropGroup(ProcessGroup *group) {
 	group->sessionPointer_ = nullptr;
 }
 
-void TerminalSession::setForegroundGroup(ProcessGroup *group) {
-	assert(group->sessionPointer_.get() == this);
+Error TerminalSession::setForegroundGroup(ProcessGroup *group) {
+	if(!group)
+		return Error::illegalArguments; // group is null, so pgrp is invalid?
```
avdgrinten, line 1128: We can `assert(group)`. This should be checked by the caller.

posix/subsystem/src/main.cpp
```diff
@@ -840,7 +840,14 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 			if(logRequests)
 				std::cout << "posix: WAIT" << std::endl;
 
-			assert(!(req.flags() & ~WNOHANG));
+			if(req.flags() & ~(WNOHANG | WUNTRACED)) {
+				if(req.flags() & WUNTRACED) {
+					std::cout << "posix: WAIT flag WUNTRACED is silently ignored" << std::endl;
```
avdgrinten, line 845: Move this case out of the outer if (it does not make sense here -- right now we only warn about unimplemented WUNTRACED if there are other unknown flags, too).

posix/subsystem/src/main.cpp
Dennisbonke: We return EPERM with the check a bit further down (line 3478 - 3481) but the children thing is indeed unimplemented.

posix/subsystem/src/main.cpp
Dennisbonke: > There is no else here, and thus `target` is not always set.

I believe that the else statement referred is found at line 3474 at the time of writing.

posix/subsystem/src/main.cpp
```diff
@@ -3429,9 +3458,51 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
 			}
+
+			if(target->pgPointer()->getSession() != self->pgPointer()->getSession() || target->pid() == target->pgPointer()->getSession()->getSessionId()) {
+				co_await sendErrorResponse(managarm::posix::Errors::INSUFFICIENT_PERMISSION);
+				continue;
+			}
+
+			if(target->getParent()->pid() == self->pid()) {
+				if(target->didExecute()) {
+					co_await sendErrorResponse(managarm::posix::Errors::ACCESS_DENIED);
```
avdgrinten, line 3488: Shouldn't this be INSUFFICIENT_PERMISSION? Or is it really EACCES?

posix/subsystem/src/main.cpp
```diff
@@ -3429,9 +3458,51 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 			} else {
 				target = self;
 			}
+
+			if(target->pgPointer()->getSession() != self->pgPointer()->getSession() || target->pid() == target->pgPointer()->getSession()->getSessionId()) {
```
avdgrinten, line 3481: What is the part right of the || good for?

posix/subsystem/src/main.cpp
```diff
@@ -3429,9 +3458,51 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+				}
+			}
+
+			pid_t pid;
+
+			if(!req->pid()) {
+				pid = self->pid();
+			} else {
+				pid = req->pid();
+			}
```
avdgrinten, line 3499: I think we don't need this variable: this is just `target->pid()`, right? We can replace the uses of ´pid` below by `target->pid()`.

posix/subsystem/src/pts.cpp
```diff
@@ -689,6 +689,59 @@ async::result<void> SlaveFile::ioctl(Process *process, managarm::fs::CntRequest
...
+			helix_ng::extractCredentials()
+		);
+		HEL_CHECK(extractCreds.error());
+
+		auto process = findProcessWithCredentials(extractCreds.credentials());
+		auto group = process->pgPointer()->getSession()->getProcessGroupById(req.pgid()).get();
+		if(!group) {
+			resp.set_error(managarm::fs::Errors::ILLEGAL_ARGUMENT);
+		} else {
+			Error ret = process->pgPointer()->getSession()->setForegroundGroup(process->pgPointer()->getSession()->getProcessGroupById(req.pgid()).get());
```
avdgrinten, line 842: This should operator on the session associated with the terminal (like TIOCGPGRP).

avdgrinten: Also, this does not use the `group` variable from above.

posix/subsystem/src/pts.cpp
```diff
@@ -689,6 +689,59 @@ async::result<void> SlaveFile::ioctl(Process *process, managarm::fs::CntRequest
...
+		managarm::fs::SvrResponse resp;
+
+		auto [extractCreds] = co_await helix_ng::exchangeMsgs(
+			conversation,
+			helix_ng::extractCredentials()
+		);
+		HEL_CHECK(extractCreds.error());
+
+		auto process = findProcessWithCredentials(extractCreds.credentials());
+		auto group = process->pgPointer()->getSession()->getProcessGroupById(req.pgid()).get();
```
avdgrinten, line 838: Store the shared_ptr in a local variable here (not the raw point returned by `.get()`) such that the group cannot be freed.

testsuites/posix-tests/src/processgroups.cpp
```diff
@@ -0,0 +1,86 @@
...
+ * We open /dev/ptmx to get a terminal,
+ * We spawn a new session with setsid,
+ * We make that opened terminal the controlling terminal,
+ * Now we can run the required tests,
+ * When done, exit the fork.
+ * Yes, this is tedious, but the only way to guarantee a proper environment.
+ */
+DEFINE_TEST(tcgetsid, ([] {
+	pid_t pid = fork();
+	if(!pid) {
```
avdgrinten, line 23: The parent should wait until the child exits (using wait/waitpid or a similar function).

posix/subsystem/src/main.cpp
Dennisbonke: From `man setpgid`
```
EACCES An attempt was made to change the process group ID of one of the children of the calling
              process and the child had already performed an execve(2) (setpgid(), setpgrp()).
```

posix/subsystem/src/pts.cpp
```diff
@@ -459,6 +571,9 @@ async::result<void> MasterFile::ioctl(Process *, managarm::fs::CntRequest req,
 			helix_ng::sendBuffer(ser.data(), ser.size())
 		);
 		HEL_CHECK(send_resp.error());
+	}else if(req.command() == TIOCSCTTY || req.command() == TIOCGPGRP
+			|| req.command() == TIOCSPGRP || req.command() == TIOCGSID) {
+		co_await _channel->commonIoctl(process, req, std::move(conversation));
```
avdgrinten, line 576: We can move `req` too, to avoid a copy.

avdgrinten:
Thanks again for the PR.

PR #253, opened 2021-03-20
netserver: Various bug fixes

### LINKED ISSUE #208: netserver: netserver hangs without any logs
After running any command which creates a connection 3 times (tested with `wget example.org`, but any command that uses networking would do), netserver seems to lock up. When running in weston, the offending terminal can be closed and the system can continue to operate, but all network functionality is lost.

The output of the e9 log of the last wget command is as follows
```
mlibc: tcgetpgrp() fails with ENOSYS
mlibc: tcgetpgrp() fails with ENOSYS
mlibc: tcgetpgrp() fails with ENOSYS
netserver:
 f020a0a
 8080808
    2500
    5687
    3500
    2500
     189
virtio-driver: sending frame
virtio-driver: sent frame
received udp datagram to port 8756
posix extern-socket.4626: Object does not implement handleClose()
mlibc: tcgetpgrp() fails with ENOSYS
mlibc: tcgetpgrp() fails with ENOSYS
mlibc: tcgetpgrp() fails with ENOSYS
virtio-driver: sending frame
virtio-driver: sent frame
```

Dennisbonke:
This PR originally held various bugfixes for the netstack. During bug hunting, the need to refactor the sendMsg code to handle partial writes (and as such also add this to the write code) came up. Furthermore, this PR fixes a formatting error, fixes 2 critical bugs in the TCP and UDP stack, adds additional error returns in several places and adds code to handle (but not yet send) FIN TCP packets.

This PR possibly fixes #208
After this PR, managarm/mlibc#242 should be merged

### DIFF
--- a/posix/subsystem/src/main.cpp
+++ b/posix/subsystem/src/main.cpp
@@ -2897,6 +2897,7 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 					co_await sendErrorResponse(managarm::posix::Errors::ALREADY_EXISTS);
 					continue;
 				}
+				assert(ret == Error::success);
 			}
 
 			struct epoll_event events[16];
@@ -2973,6 +2974,7 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 				co_await sendErrorResponse(managarm::posix::Errors::ALREADY_EXISTS);
 				continue;
 			}
+			assert(ret == Error::success);
 
 			managarm::posix::SvrResponse resp;
 			resp.set_error(managarm::posix::Errors::SUCCESS);
@@ -2998,6 +3000,7 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 				co_await sendErrorResponse(managarm::posix::Errors::FILE_NOT_FOUND);
 				continue;
 			}
+			assert(ret == Error::success);
 
 			managarm::posix::SvrResponse resp;
 			resp.set_error(managarm::posix::Errors::SUCCESS);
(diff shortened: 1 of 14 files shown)


protocols/fs/src/server.cpp
```diff
@@ -248,6 +248,15 @@ async::detached handlePassthrough(smarter::shared_ptr<void> file,
...
 
+				auto ser = resp.SerializeAsString();
+				auto [send_resp] = co_await helix_ng::exchangeMsgs(
+					conversation,
+					helix_ng::sendBuffer(ser.data(), ser.size())
+				);
+				HEL_CHECK(send_resp.error());
+			} else if(res.error() == Error::wouldBlock) {
+				resp.set_error(managarm::fs::Errors::WOULD_BLOCK);
+
```
avdgrinten, line 259: We can avoid duplicating the IPC code here: just set a different code using `set_error` but only have the helix_ng invocation once.

posix/subsystem/src/main.cpp
```diff
@@ -2891,8 +2891,12 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 				assert(file && "Illegal FD for EPOLL_ADD item");
 				auto locked = file->weakFile().lock();
 				assert(locked);
-				epoll::addItem(epfile.get(), self.get(), std::move(locked),
+				Error ret = epoll::addItem(epfile.get(), self.get(), std::move(locked),
 						req.events(i), i);
+				if(ret == Error::alreadyExists) {
```
avdgrinten, line 2896: In principle, this needs to handle other error as well. Since no other errors exist, we could also assert that the function succeeds.

avdgrinten:
LGTM now, thank you!

PR #254, opened 2021-03-23
Various aarch64 related changes

qookei:


kernel/thor/system/dtb/dtb.cpp
```diff
@@ -239,13 +290,61 @@ void DeviceTreeNode::finalizeInit() {
...
+	bool isPPI = prop->asU32(i);
+	uint32_t rawId = prop->asU32(i + 4);
+	uint32_t flags = prop->asU32(i + 8);
+
+	if (isPPI)
+		irq.id = rawId + 16;
+	else
+		irq.id = rawId + 32;
+
+	switch (flags & 0xF) {
```
avdgrinten, line 324: What if none of the cases below applies? (Maybe we should log an error?)

kernel/thor/system/pci/pci_discover.cpp
```diff
@@ -968,6 +969,305 @@ void configureBridges(PciBus *root, PciBus *bus, int &highestId) {
...
+					PciBusResource::prefMemory,
+					0, nullptr, bridge});
+		}
+	}
+
+	frg::insertion_sort(required.begin(), required.end(), [] (auto a, auto b) {
+		if (a.flags == b.flags)
+			return a.size < b.size;
+		return a.flags > b.flags;
+	});
```
avdgrinten, line 1071: Why do we sort the requirements here?

qookei: We sort them to allocate them out of resources in the most efficient way. Because they're sorted in descending order by size, all allocations of requirements with the same flags do not need to add any padding as the smaller alignment is always fulfilled by the previous, larger allocation (because all sizes are powers of 2).

avdgrinten: Alright, makes sense. Maybe add a comment to clarify that.

avdgrinten (approved):

avdgrinten:
LGTM now, thank you!

PR #260, opened 2021-04-02
thor: Add GDB server over serial I/O

avdgrinten:
We add a GDB server to `thor`. Whenever a server dies with a panic or segfault, we launch the server over the standard PC uart.


avdgrinten:
Adjusted some parts after discussion in Discord; this good be fine now.

PR #261, opened 2021-04-02
posix-init: Don't reset the environment in kmscon

Dennisbonke:
This is a hack, kmscon should invoke a login program which sets stuff like `HOME`, but as we don't have that yet, we should keep the environment as is to provide a better user experience.

### DIFF
--- a/posix/init/src/stage2.cpp
+++ b/posix/init/src/stage2.cpp
@@ -134,7 +134,10 @@ int main() {
 //		putenv("MESA_DEBUG=1");
 
 		if(launch == "kmscon") {
-			execl("/usr/bin/kmscon", "kmscon", nullptr);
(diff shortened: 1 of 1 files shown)


posix/init/src/stage2.cpp
```diff
@@ -134,7 +134,10 @@ int main() {
 //		putenv("MESA_DEBUG=1");
 
 		if(launch == "kmscon") {
-			execl("/usr/bin/kmscon", "kmscon", nullptr);
+			// FIXME: kmscon should invoke a login program which sets these environment vars.
```
avdgrinten, line 137: Amend this patch to replace `FIXME` by `TODO`. FIXME makes it should quite *strong* but this is not really that problematic.

avdgrinten (approved):

PR #266, opened 2021-04-10
AArch64 related changes

qookei:


posix/subsystem/src/process.cpp
```diff
@@ -583,14 +612,19 @@ async::result<void> SignalContext::raiseContext(SignalItem *item, Process *proce
...
 	};
 
 	// Store the current register stack on the stack.
+	nsp -= (simdState.size() + 15) & ~uintptr_t(15);
+	auto storeSimdFrame = co_await helix_ng::writeMemory(thread, nsp,
+			simdState.size(), simdState.data());
+	HEL_CHECK(storeSimdFrame.error());
+
 	assert(alignof(SignalFrame) == 8);
 	auto frame = alignFrame(sizeof(SignalFrame));
```
avdgrinten, line 629: I think we can handle this more elegantly by storing the SIMD state directly above `SignalFrame` (= do the computation as if the frame was `simdState.size() + sizeof(SignalFrame)` bytes and storing the SIMD state *after* alignFrame).

avdgrinten: This would also simply the code in the signal return path.

kernel/thor/system/pci/pci_discover.cpp
```diff
@@ -74,12 +74,11 @@ namespace {
...
-					assert(device->bars[k].offset == 0);
+				if(device->bars[k].hostType == PciBar::kBarIo) {
 					msg.set_io_type(managarm::hw::IoType::PORT);
 					msg.set_address(device->bars[k].address);
 					msg.set_length(device->bars[k].length);
-				}else if(device->bars[k].type == PciBar::kBarMemory) {
+				}else if(device->bars[k].hostType == PciBar::kBarMemory) {
 					msg.set_io_type(managarm::hw::IoType::MEMORY);
 					msg.set_address(device->bars[k].address);
 					msg.set_length(device->bars[k].length);
```
avdgrinten, line 84: I think we should not make the hw protocol lie about the BAR type (some drivers might need to change their behavior based on the "true" BAR type), instead, we should add `host_type` to the protocol.

avdgrinten (commented):
One comment regarding the hw protocl, we can merge once that is fixed.

avdgrinten (approved):
LGTM now.

Issue #75, opened 2019-05-04
drivers/ps2: Some keys are unimplemented


avdgrinten:
This affects at least:
- F1-F12
- Numpad
- Arrow keys

avdgrinten:
@qookei is this still an issue?

qookei:
The arrow keys are working now, the numpad and function keys still appear broken

PR #267, opened 2021-04-11
drivers: fix ps2 function and numpad keys

rizet:
Resolves issue #75 with function and numpad keys not working properly.


avdgrinten:
LGTM, thank you.

PR #268, opened 2021-04-13
posix: Fix poll/epoll FD duplication checks

64:
This allows Rust programs to run.

Note: I did not write this code, avdgrinten did :)


avdgrinten:
Alright. The code can still be simplified a bit and it also changes the timeout conditionals (in cases that previously led to a panic) but let's merge it now to enable Rust.

PR #271, opened 2021-04-15
Refactor HW protocol to bragi

qookei:


protocols/hw/hw.bragi
```diff
@@ -0,0 +1,103 @@
+namespace "managarm::hw";
+
+enum Errors {
+	SUCCESS = 0,
+	ILLEGAL_REQUEST,
```
avdgrinten, line 5: I think on syntactically broken requests, we should just close the lane. On requests that we don't understand, we should use the new `dismiss` item.

avdgrinten (commented):
Only one comment, remainder looks good.

avdgrinten (approved):
LGTM now.

PR #274, opened 2021-04-16
posix: Implement ioctl(FIOCLEX)

64:
This PR adds a new request type in `bragi` and `posix` to support `ioctl(FIOCLEX)`. This is used by Rust's `std` to set the `CLOEXEC` flag on file descriptors.

mlibc supported is added in managarm/mlibc#259.


avdgrinten:
Looks good but new requests should use specialized messages and not the generic `CntRequest` (see `VmMapRequest`).

avdgrinten:
LGTM now. Thank you for the PR! I will squash the two commits into one and merge.

PR #287, opened 2021-04-30
managarm-kernel: Fix various warnings

Dennisbonke:
This PR fixes several warnings found in the kernel after merging of #269.

### DIFF
--- a/kernel/thor/generic/hel.cpp
+++ b/kernel/thor/generic/hel.cpp
@@ -145,7 +145,7 @@ HelError helSubmitAsyncNop(HelHandle queueHandle, uintptr_t context) {
 
 	async::detach_with_allocator(*kernelAlloc, [] (
 			smarter::shared_ptr<IpcQueue> queue, uintptr_t context) -> coroutine<void> {
-		HelSimpleResult helResult{kHelErrNone};
+		HelSimpleResult helResult{.error = kHelErrNone};
 		QueueSource ipcSource{&helResult, sizeof(HelSimpleResult), nullptr};
 		co_await queue->submit(&ipcSource, context);
 	}(std::move(queue), context));
@@ -844,7 +844,7 @@ HelError helSubmitProtectMemory(HelHandle space_handle,
 				uint32_t protectFlags, uintptr_t context) -> coroutine<void> {
 			co_await space->protect(pointer, length, protectFlags);
 
-			HelSimpleResult helResult{kHelErrNone};
+			HelSimpleResult helResult{.error = kHelErrNone};
 			QueueSource ipcSource{&helResult, sizeof(HelSimpleResult), nullptr};
 			co_await queue->submit(&ipcSource, context);
 	}(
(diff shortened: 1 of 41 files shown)


kernel/thor/generic/hel.cpp
```diff
@@ -145,7 +145,7 @@ HelError helSubmitAsyncNop(HelHandle queueHandle, uintptr_t context) {
 
 	async::detach_with_allocator(*kernelAlloc, [] (
 			smarter::shared_ptr<IpcQueue> queue, uintptr_t context) -> coroutine<void> {
-		HelSimpleResult helResult{kHelErrNone};
+		HelSimpleResult helResult{kHelErrNone, 0};
```
avdgrinten, line 148: For `HelSimpleResult` and other `Hel...Result` types, it would be preferable to use designated initializers (something like `{.error = kHelErrNone}`. The `0` here initializes a padding word at the end of the struct; it is quite ugly if we have to know the exact number of padding words in each struct.

kernel/thor/generic/main.cpp
```diff
@@ -81,7 +81,7 @@ extern "C" void thorInitialize() {
 	SkeletalRegion::initialize();
 	physicalAllocator.initialize();
 	auto region = reinterpret_cast<EirRegion *>(thorBootInfoPtr->regionInfo);
-	for(int i = 0; i < thorBootInfoPtr->numRegions; i++)
+	for(unsigned long i = 0; i < thorBootInfoPtr->numRegions; i++)
```
avdgrinten, line 84: Is this `unsigned long` or `size_t`?

kernel/thor/generic/thor-internal/memory-view.hpp
```diff
@@ -750,7 +750,7 @@ inline auto copyToView(MemoryView *view, uintptr_t offset,
 	};
 
 	return async::let([=] {
-		return Node{view, offset, pointer, size, std::move(wq)};
+		return Node{view, offset, pointer, size, std::move(wq), 0, 0};
```
avdgrinten, line 753: Here we can use designated initializers.

avdgrinten: Same for the two other cases below.

kernel/thor/system/pci/pci_discover.cpp
```diff
@@ -1045,7 +1045,7 @@ void configureBridges(PciBus *root, PciBus *bus, int &highestId) {
 			}
 
 			if (parent) {
-				assert(highestId < parent->subordinateId);
+				assert(highestId < static_cast<int>(parent->subordinateId));
```
avdgrinten, line 1048: Maybe `highestId` should just be `uint32_t`? I don't think we need negative numbers here. Pinging @qookei to be sure.

kernel/thor/generic/main.cpp
Dennisbonke: The struct `EirInfo` defines `numRegions` as type `EirSize` which is defined as `uint64_t`

kernel/thor/system/pci/pci_discover.cpp
qookei: True, `highestId` could just be `uint32_t`

kernel/thor/generic/main.cpp
avdgrinten: I think then we should probably just use `size_t` here.

avdgrinten (commented):
LGTM except for one tiny nitpick.

avdgrinten:
LGTM now. Thank you for the PR, this was a lot of work.

PR #288, opened 2021-05-01
managarm-system: Fix several warnings

Dennisbonke:
This PR fixes several warnings found in the userspace parts after merging of #269.

### DIFF
--- a/core/drm/src/core.cpp
+++ b/core/drm/src/core.cpp
@@ -30,7 +30,7 @@ drm_core::Device::Device() {
 		SrcWProperty()
 		: drm_core::Property{drm_core::IntPropertyType{}} { }
 
-		bool validate(const Assignment& assignment) override {
+		bool validate(const Assignment&) override {
 			return true;
 		};
 	};
@@ -40,7 +40,7 @@ drm_core::Device::Device() {
 		SrcHProperty()
 		: drm_core::Property{drm_core::IntPropertyType{}} { }
 
-		bool validate(const Assignment& assignment) override {
+		bool validate(const Assignment&) override {
 			return true;
 		};
 	};
@@ -98,7 +98,7 @@ drm_core::Device::Device() {
 		CrtcXProperty()
 		: drm_core::Property{drm_core::IntPropertyType{}} { }
 
-		bool validate(const Assignment& assignment) override {
+		bool validate(const Assignment&) override {
 			return true;
 		};
 	};
@@ -108,7 +108,7 @@ drm_core::Device::Device() {
 		CrtcYProperty()
 		: drm_core::Property{drm_core::IntPropertyType{}} { }
 
-		bool validate(const Assignment& assignment) override {
+		bool validate(const Assignment&) override {
 			return true;
 		};
 	};
--- a/core/drm/src/core.cpp
+++ b/core/drm/src/core.cpp
@@ -590,22 +590,22 @@ drm_core::File::ioctl(void *object, managarm::fs::CntRequest req,
 		managarm::fs::SvrResponse resp;
 
 		auto &crtcs = self->_device->getCrtcs();
-		for(int i = 0; i < crtcs.size(); i++) {
+		for(size_t i = 0; i < crtcs.size(); i++) {
 			resp.add_drm_crtc_ids(crtcs[i]->id());
 		}
 			
 		auto &encoders = self->_device->getEncoders();
-		for(int i = 0; i < encoders.size(); i++) {
+		for(size_t i = 0; i < encoders.size(); i++) {
 			resp.add_drm_encoder_ids(encoders[i]->id());
 		}
 	
 		auto &connectors = self->_device->getConnectors();
-		for(int i = 0; i < connectors.size(); i++) {
+		for(size_t i = 0; i < connectors.size(); i++) {
 			resp.add_drm_connector_ids(connectors[i]->id());
 		}
 		
 		auto &fbs = self->getFrameBuffers();
-		for(int i = 0; i < fbs.size(); i++) {
+		for(size_t i = 0; i < fbs.size(); i++) {
 			resp.add_drm_fb_ids(fbs[i]->id());
 		}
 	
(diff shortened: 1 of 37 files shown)


drivers/kbd/src/main.cpp
```diff
@@ -202,7 +202,7 @@ bool Controller::processData(int port) {
 // --------------------------------------------------------------------
 
 Controller::Device::Device(Controller *controller, int port)
-: _controller{controller}, _port{port}, _deviceType{0}, _exists{false} {
+: _controller{controller}, _port{port}, _deviceType{0, 0, 0, 0}, _exists{false} {
```
avdgrinten, line 205: This should probably just use zero initialization by `{}`.

core/drm/src/core.cpp
```diff
@@ -590,22 +590,22 @@ drm_core::File::ioctl(void *object, managarm::fs::CntRequest req,
 		managarm::fs::SvrResponse resp;
 
 		auto &crtcs = self->_device->getCrtcs();
-		for(int i = 0; i < crtcs.size(); i++) {
+		for(unsigned long i = 0; i < crtcs.size(); i++) {
```
avdgrinten, line 593: This should be size_t, since std::vector::size() returns a size_t.

avdgrinten: Same below, whenever we compare against the size of a std::vector.

drivers/kernletcc/src/fafnir.cpp
```diff
@@ -301,7 +301,7 @@ std::vector<uint8_t> compileFafnir(const uint8_t *code, size_t size,
 			int nargs = extractUint();
 			int nrvs = extractUint();
 			auto function = extractString();
-			assert(comp->opstack.size() >= nargs);
+			assert(comp->opstack.size() >= static_cast<unsigned long>(nargs));
```
avdgrinten, line 304: size_t (?).

posix/subsystem/src/file.hpp
```diff
@@ -349,7 +349,7 @@ struct DummyFile final : File {
 				file, &fileOperations, file->_cancelServe));
 	}
 
-	DummyFile(std::shared_ptr<MountView> mount, std::shared_ptr<FsLink> link, DefaultOps default_ops = 0)
+	DummyFile(std::shared_ptr<MountView> mount, std::shared_ptr<FsLink> link)
```
avdgrinten, line 352: Why is this argument removed? Why did it exist in the first place? Do we need to pass it to the `File` ctor?

posix/subsystem/src/main.cpp
```diff
@@ -3613,7 +3613,7 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 				group->reassociateProcess(target.get());
 			} else {
 				// Not found, making it if pgid and pid match, or if pgid is 0, indicating that we should make one
-				if(target->pid() == req->pgid() || !req->pgid()) {
+				if(target->pid() == static_cast<int>(req->pgid()) || !req->pgid()) {
```
avdgrinten, line 3616: What type is `req->pgid()`?

posix/subsystem/src/main.cpp
```diff
@@ -187,7 +187,7 @@ async::result<void> observeThread(std::shared_ptr<Process> self,
 		HEL_CHECK(observe.error());
 		sequence = observe.sequence();
 
-		if(observe.observation() == kHelObserveSuperCall + 10) {
+		if(observe.observation() == static_cast<int>(kHelObserveSuperCall + 10)) {
```
avdgrinten, line 190: I think the proper fix here is to make `observer.obseration()` return an `unsigned int` (the HelObserveResult that the value is taken from is already `unsigned int`).

posix/subsystem/src/tmp_fs.cpp
```diff
@@ -170,17 +170,14 @@ struct FifoNode final : Node {
...
+	:Node{superblock} {
 		fifo::createNamedChannel(this);
 	}
 
 	~FifoNode() {
 		fifo::unlinkNamedChannel(this);
 	}
-
-private:
-	mode_t mode_;
```
avdgrinten, line 183: Instead of removing `mode_`, we should return it from `getStats()`.

posix/subsystem/src/file.hpp
Dennisbonke: Unsure, the compiler was complaining that it was unused, and removing it gave no issues.

posix/subsystem/src/tmp_fs.cpp
Dennisbonke: Agreed, but I figured that is a bit out of scope for this PR.

posix/subsystem/src/main.cpp
Dennisbonke: That is an uint64_t

posix/subsystem/src/file.hpp
avdgrinten: There is probably a reason why it was added. Should it be passed through to the `File` ctor?

posix/subsystem/src/tmp_fs.cpp
avdgrinten: Adding `getStats()` is pretty simple, that should only be a few SLoC.

Dennisbonke: Even then, it seems that this field is unused, looking at implementations from tmpfs `MemoryNode`.

posix/subsystem/src/file.hpp
avdgrinten: It makes a different if non-zero default_ops are passed. We should pass this parameter to the `File` constructor.

posix/subsystem/src/tmp_fs.cpp
avdgrinten: Well, then MemoryNode's getStats() is broken, too. getStats() should return the file mode.

posix/subsystem/src/main.cpp
avdgrinten: Why do we still have a cast here?

avdgrinten (approved):
LGTM now. Thank you very much for this PR!

PR #299, opened 2021-05-14

cflaviu:
- replaced "if, multipe else-if, else" sequence with switch
- replaced string comparison with switch
- fixed warnings
- small improvements


cflaviu:
a commit was missing -> closing

PR #300, opened 2021-05-14

cflaviu:
- replaced "if, multipe else-if, else" sequence with switch
- replaced multiple string comparisons with switch
- fixed warnings
- small improvements

### DIFF
--- a/core/drm/src/core.cpp
+++ b/core/drm/src/core.cpp
@@ -30,7 +30,7 @@ drm_core::Device::Device() {
 		SrcWProperty()
 		: drm_core::Property{drm_core::IntPropertyType{}} { }
 
-		bool validate(const Assignment& assignment) override {
+		bool validate(const Assignment&) override {
 			return true;
 		};
 	};
@@ -40,7 +40,7 @@ drm_core::Device::Device() {
 		SrcHProperty()
 		: drm_core::Property{drm_core::IntPropertyType{}} { }
 
-		bool validate(const Assignment& assignment) override {
+		bool validate(const Assignment&) override {
 			return true;
 		};
 	};
@@ -98,7 +98,7 @@ drm_core::Device::Device() {
 		CrtcXProperty()
 		: drm_core::Property{drm_core::IntPropertyType{}} { }
 
-		bool validate(const Assignment& assignment) override {
+		bool validate(const Assignment&) override {
 			return true;
 		};
 	};
@@ -108,7 +108,7 @@ drm_core::Device::Device() {
 		CrtcYProperty()
 		: drm_core::Property{drm_core::IntPropertyType{}} { }
 
-		bool validate(const Assignment& assignment) override {
+		bool validate(const Assignment&) override {
 			return true;
 		};
 	};
@@ -590,22 +590,22 @@ drm_core::File::ioctl(void *object, managarm::fs::CntRequest req,
 		managarm::fs::SvrResponse resp;
 
 		auto &crtcs = self->_device->getCrtcs();
-		for(int i = 0; i < crtcs.size(); i++) {
+		for(size_t i = 0; i < crtcs.size(); i++) {
 			resp.add_drm_crtc_ids(crtcs[i]->id());
 		}
 			
 		auto &encoders = self->_device->getEncoders();
-		for(int i = 0; i < encoders.size(); i++) {
+		for(size_t i = 0; i < encoders.size(); i++) {
 			resp.add_drm_encoder_ids(encoders[i]->id());
 		}
 	
 		auto &connectors = self->_device->getConnectors();
-		for(int i = 0; i < connectors.size(); i++) {
+		for(size_t i = 0; i < connectors.size(); i++) {
 			resp.add_drm_connector_ids(connectors[i]->id());
 		}
 		
 		auto &fbs = self->getFrameBuffers();
-		for(int i = 0; i < fbs.size(); i++) {
+		for(size_t i = 0; i < fbs.size(); i++) {
 			resp.add_drm_fb_ids(fbs[i]->id());
 		}
 	
@@ -631,7 +631,7 @@ drm_core::File::ioctl(void *object, managarm::fs::CntRequest req,
 		assert(conn);
 		
 		auto psbl_enc = conn->getPossibleEncoders();
-		for(int i = 0; i < psbl_enc.size(); i++) { 
+		for(size_t i = 0; i < psbl_enc.size(); i++) { 
 			resp.add_drm_encoders(psbl_enc[i]->id());
 		}
 
--- a/drivers/kbd/src/main.cpp
+++ b/drivers/kbd/src/main.cpp
@@ -131,21 +131,21 @@ std::optional<uint8_t> Controller::recvByte(uint64_t timeout) {
 	return _space.load(kbd_register::data);
 }
 
-void Controller::submitCommand(controller_cmd::DisablePort tag, int port) {
+void Controller::submitCommand(controller_cmd::DisablePort, int port) {
 	if (port == 0)
 		sendCommandByte(disable1stPort);
 	else if (port == 1)
 		sendCommandByte(disable2ndPort);
 }
 
-void Controller::submitCommand(controller_cmd::EnablePort tag, int port) {
+void Controller::submitCommand(controller_cmd::EnablePort, int port) {
 	if (port == 0)
 		sendCommandByte(enable1stPort);
 	else if (port == 1)
 		sendCommandByte(enable2ndPort);
 }
 
-uint8_t Controller::submitCommand(controller_cmd::GetByte0 tag) {
+uint8_t Controller::submitCommand(controller_cmd::GetByte0) {
 	sendCommandByte(readByte0);
 
 	auto result = recvByte(default_timeout);
@@ -155,12 +155,12 @@ uint8_t Controller::submitCommand(controller_cmd::GetByte0 tag) {
 	return result.value();
 }
 
-void Controller::submitCommand(controller_cmd::SetByte0 tag, uint8_t val) {
+void Controller::submitCommand(controller_cmd::SetByte0, uint8_t val) {
 	sendCommandByte(writeByte0);
 	sendDataByte(val);
 }
 
-void Controller::submitCommand(controller_cmd::SendBytePort2 tag) {
+void Controller::submitCommand(controller_cmd::SendBytePort2) {
 	sendCommandByte(0xD4); // TODO: define a constant?
 }
 
@@ -202,7 +202,7 @@ bool Controller::processData(int port) {
 // --------------------------------------------------------------------
 
 Controller::Device::Device(Controller *controller, int port)
-: _controller{controller}, _port{port}, _deviceType{0}, _exists{false} {
+: _controller{controller}, _port{port}, _deviceType{}, _exists{false} {
 }
 
 async::result<void> Controller::Device::init() {
--- a/drivers/libblockfs/src/ext2fs.cpp
+++ b/drivers/libblockfs/src/ext2fs.cpp
@@ -935,7 +935,7 @@ async::result<uint32_t> FileSystem::allocateBlock() {
 		// TODO: Update the block group descriptor table.
 
 		auto words = reinterpret_cast<uint32_t *>(bitmap_map.get());
-		for(int i = 0; i < (blocksPerGroup + 31) / 32; i++) {
+		for(unsigned int i = 0; i < (blocksPerGroup + 31) / 32; i++) {
 			if(words[i] == 0xFFFFFFFF)
 				continue;
 			for(int j = 0; j < 32; j++) {
@@ -978,7 +978,7 @@ async::result<uint32_t> FileSystem::allocateInode() {
 		// TODO: Update the block group descriptor table.
 
 		auto words = reinterpret_cast<uint32_t *>(bitmap_map.get());
-		for(int i = 0; i < (inodesPerGroup + 31) / 32; i++) {
+		for(unsigned int i = 0; i < (inodesPerGroup + 31) / 32; i++) {
 			if(words[i] == 0xFFFFFFFF)
 				continue;
 			for(int j = 0; j < 32; j++) {
--- a/drivers/usb/hcds/xhci/src/main.cpp
+++ b/drivers/usb/hcds/xhci/src/main.cpp
@@ -358,7 +358,7 @@ void Controller::ringDoorbell(uint8_t doorbell, uint8_t target, uint16_t stream_
 // ------------------------------------------------------------------------
 
 Controller::CommandRing::CommandRing(Controller *controller)
-:_commandRing{&controller->_memoryPool}, _dequeuePtr{0}, _enqueuePtr{0},
+:_commandRing{&controller->_memoryPool}, _enqueuePtr{0},
 	_controller{controller}, _pcs{true} {
 
 	for (uint32_t i = 0; i < commandRingSize; i++) {
@@ -832,7 +832,7 @@ async::detached Controller::Port::initPort() {
 
 Controller::TransferRing::TransferRing(Controller *controller)
 :_transferRing{&controller->_memoryPool}, _dequeuePtr{0}, _enqueuePtr{0},
-	_controller{controller}, _pcs{true} {
+	_pcs{true} {
 
 	for (uint32_t i = 0; i < transferRingSize; i++) {
 		_transferRing->ent[i] = {{0, 0, 0, 0}};
@@ -1209,8 +1209,8 @@ async::result<void> Controller::Device::setupEndpoint(int endpoint, PipeType dir
 // ------------------------------------------------------------------------
 
 Controller::ConfigurationState::ConfigurationState(Controller *controller, 
-		std::shared_ptr<Device> device, int number)
-:_controller{controller}, _device{device}, _number{number} {
+		std::shared_ptr<Device> device, int)
+:_controller{controller}, _device{device} {
 }
 
 async::result<Interface> Controller::ConfigurationState::useInterface(int number, int alternative) {
@@ -1223,8 +1223,8 @@ async::result<Interface> Controller::ConfigurationState::useInterface(int number
 // ------------------------------------------------------------------------
 
 Controller::InterfaceState::InterfaceState(Controller *controller, 
-		std::shared_ptr<Device> device, int interface)
-: _controller{controller}, _device{device}, _interface{interface} {
+		std::shared_ptr<Device> device, int)
+: _controller{controller}, _device{device} {
 }
 
 async::result<Endpoint> Controller::InterfaceState::getEndpoint(PipeType type, int number) {
@@ -1235,9 +1235,9 @@ async::result<Endpoint> Controller::InterfaceState::getEndpoint(PipeType type, i
 // Controller::EndpointState
 // ------------------------------------------------------------------------
 
-Controller::EndpointState::EndpointState(Controller *controller, 
+Controller::EndpointState::EndpointState(Controller *, 
 		std::shared_ptr<Device> device, int endpoint, PipeType type)
-: _controller{controller}, _device{device}, _endpoint{endpoint}, _type{type} {
+: _device{device}, _endpoint{endpoint}, _type{type} {
 }
 
 async::result<void> Controller::EndpointState::transfer(ControlTransfer info) {
--- a/posix/subsystem/src/main.cpp
+++ b/posix/subsystem/src/main.cpp
@@ -2962,7 +2962,7 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 			assert(req.fds_size() == req.events_size());
 
 			bool errorOut = false;
-			for(int i = 0; i < req.fds_size(); i++) {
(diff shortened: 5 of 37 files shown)


avdgrinten:
Thank you for the PR and for splitting up the commits. This is now a lot easier to review. I can review this within the next two days.

We will probably merge #288 first; this might cause some conflicts, but we can probably just drop the corresponding commits from this PR.

cflaviu:
Welcome. I solved the conflicts.

posix/subsystem/src/main.cpp
```diff
@@ -553,12 +572,12 @@ async::result<void> observeThread(std::shared_ptr<Process> self,
 			if(!self->checkSignalRaise())
 				std::cout << "\e[33m" "posix: Ignoring global signal flag "
 						"during synchronous SIGILL" "\e[39m" << std::endl;
-			bool killed;
 			co_await self->signalContext()->raiseContext(item, self.get(), killed);
 			if(killed)
```
cflaviu, line 576: missing !

core/drm/src/core.cpp
```diff
@@ -1164,12 +1205,10 @@ uint32_t drm_core::convertLegacyFormat(uint32_t bpp, uint32_t depth) {
...
 		case 32:
 			assert(depth == 24 || depth == 30 || depth == 32);
-			if(depth == 24) {
-				return DRM_FORMAT_XRGB8888;
-			}else if(depth == 30) {
-				return DRM_FORMAT_XRGB2101010;
-			}else {
-				return DRM_FORMAT_ARGB8888;
+			switch(depth) {
+			case 24: return DRM_FORMAT_XRGB8888;
```
Dennisbonke, line 1209: Conform the [coding style](https://docs.managarm.org/handbook/contributing/coding-style.html), we should put the return statements on their own line.

core/virtio/src/core.cpp
```diff
@@ -595,15 +595,17 @@ discover(protocols::hw::Device hw_device, DiscoverMode mode) {
 			auto bar = co_await hw_device.accessBar(bir);
 			Mapping mapping{std::move(bar), info.barInfo[bir].offset + offset, length};
 
-			if(subtype == 1) {
-				common_mapping = std::move(mapping);
-			}else if(subtype == 2) {
-				notify_mapping = std::move(mapping);
+			switch(subtype) {
+			case 1: common_mapping = std::move(mapping);
```
Dennisbonke, line 599: Same here, statements are on their own line.

drivers/kbd/src/main.cpp
```diff
@@ -634,17 +631,21 @@ async::detached Controller::KbdDevice::processReports() {
 
 		byte0 = (co_await _port->pullByte()).value();
 
-		if (byte0 == 0xE0) {
+		switch (byte0)
+		{
```
Dennisbonke, line 635: This brace should not be on it's own line.

drivers/libevbackend/src/libevbackend.cpp
```diff
@@ -390,18 +408,29 @@ void EventDevice::emitEvent(int type, int code, int value) {
...
+		break;
+	case EV_REL:
+		if (!value)
+			return;
+		break;
+	case EV_ABS:
+		if (value == _absoluteSlots[code].value)
+			return;
+		break;
+	default:;
```
Dennisbonke, line 424: What is this `;` doing here?

drivers/libterminal/src/libterminal.cpp
```diff
@@ -281,23 +313,29 @@ void Emulator::printChar(char character) {
...
+	case kStatusCsi: {
 		handleCsi(character);
 	}
+	default:;
+	}
 }
 
-void Emulator::printString(std::string string){
-	for(unsigned int i = 0; i < string.size(); i++) {
+void Emulator::printString(const std::string& text){
```
Dennisbonke, line 331: `const std::string& text` should be `const std::string &text`.

drivers/libblockfs/src/ext2fs.cpp
```diff
@@ -761,13 +761,14 @@ async::detached FileSystem::initiateInode(std::shared_ptr<Inode> inode) {
...
-	if((disk_inode->mode & EXT2_S_IFMT) == EXT2_S_IFREG) {
-		inode->fileType = kTypeRegular;
-	}else if((disk_inode->mode & EXT2_S_IFMT) == EXT2_S_IFLNK) {
-		inode->fileType = kTypeSymlink;
-	}else if((disk_inode->mode & EXT2_S_IFMT) == EXT2_S_IFDIR) {
-		inode->fileType = kTypeDirectory;
-	}else{
+	switch(disk_inode->mode & EXT2_S_IFMT) {
+	case EXT2_S_IFREG:
+		inode->fileType = kTypeRegular; break;
```
Dennisbonke, line 766: Every statement should be on it's own line, so the `break;` would be it's own line.

drivers/libterminal/src/libterminal.cpp
```diff
@@ -130,40 +147,50 @@ void Emulator::handleControlSeq(char character) {
...
+	case 'K': {
 		int n = 0;
 		if(!params.empty())
 			n = params[0];
 
 		Attribute attribute;
-		if(n == 0) {
-			for(int i = cursorX; i < width; i++) {
+		switch(n) {
+		case 0:	for(int i = cursorX; i < width; i++) {
```
Dennisbonke, line 170: Statements should be on their own line here too.

drivers/usb/hcds/xhci/src/main.cpp
```diff
@@ -384,10 +384,11 @@ void Controller::CommandRing::pushRawCommand(RawTrb cmd,
 	assert(_enqueuePtr < 127 && "ring aspect of the command ring not yet supported");
 	_commandRing->ent[_enqueuePtr] = cmd;
 	_commandEvents[_enqueuePtr] = ev;
+	auto& value = _commandRing->ent[_enqueuePtr].val[3];
```
Dennisbonke, line 387: `auto& value` should be `auto &value`.

drivers/usb/hcds/xhci/src/main.cpp
```diff
@@ -860,10 +861,11 @@ void Controller::TransferRing::pushRawTransfer(RawTrb cmd,
 
 	_transferRing->ent[_enqueuePtr] = cmd;
 	_transferEvents[_enqueuePtr] = ev;
+	auto& value = _transferRing->ent[_enqueuePtr].val[3];
```
Dennisbonke, line 864: Same here.

kernel/eir/arch/arm/arch.cpp
```diff
@@ -96,12 +96,15 @@ void mapSingle4kPage(address_t address, address_t physical, uint32_t flags,
...
 
-	if (caching_mode == CachingMode::writeCombine) {
-		new_entry |= kPageGRE | kPageOuterSh;
-	} else if (caching_mode == CachingMode::mmio) {
-		new_entry |= kPagenGnRnE | kPageOuterSh;
-	} else {
-		assert(caching_mode == CachingMode::null);
+	switch (caching_mode) {
+	case CachingMode::writeCombine:
+		new_entry |= kPageGRE | kPageOuterSh; break;
```
Dennisbonke, line 101: One statement per line.

kernel/eir/arch/arm/raspi4/raspi4.cpp
```diff
@@ -387,12 +387,11 @@ extern "C" void eirRaspi4Main(uintptr_t deviceTreePtr) {
...
 	uintptr_t initrd = 0;
 	if (auto p = chosenNode.findProperty("linux,initrd-start"); p) {
-		if (p->size() == 4)
-			initrd = p->asU32();
-		else if (p->size() == 8)
-			initrd = p->asU64();
-		else
-			assert(!"Invalid linux,initrd-start size");
+		switch (p->size()) {
+		case 4:	initrd = p->asU32(); break;
```
Dennisbonke, line 391: Same here, only one statement per line.

kernel/thor/arch/x86/thor-internal/arch/cpu.hpp
```diff
@@ -160,24 +163,22 @@ struct IrqImageAccessor {
 
 	bool inThreadDomain() {
 		assert(inPreemptibleDomain());
-		if(*cs() == kSelExecutorFaultCode
-				|| *cs() == kSelExecutorSyscallCode
-				|| *cs() == kSelClientUserCompat
-				|| *cs() == kSelClientUserCode) {
+		switch(*cs())
+		{
```
Dennisbonke, line 167: This brace is wrong.

kernel/thor/generic/hel.cpp
```diff
@@ -2363,7 +2407,9 @@ HelError helSubmitAsync(HelHandle handle, const HelAction *actions, size_t count
 		smarter::weak_ptr<Universe> weakUniverse;
 		smarter::shared_ptr<IpcQueue> ipcQueue;
 		frg::dyn_array<Item, KernelAlloc> items;
-	} *closure = frg::construct<Closure>(*kernelAlloc, std::move(items));
+	}
```
Dennisbonke, line 2410: Seems like an extra white space was added here.

kernel/thor/generic/servers.cpp
```diff
@@ -462,15 +471,17 @@ coroutine<Error> handleReq(LaneHandle boundLane) {
...
+	case managarm::svrctl::CntReqType::FILE_UPLOAD_DATA: {
 		auto [dataError, dataBuffer] = co_await RecvBufferSender{lane};
 		if(dataError != Error::success)
 			co_return dataError;
 		MfsRegular *file;
 		if(!(co_await createMfsFile(req.name(), dataBuffer.data(), dataBuffer.size(), &file))) {
 			// TODO: Verify that the file data matches. This is somewhat expensive because
-			//       we would have to map the file's memory. Hence, we do not implement
-			//       it for now.
+			//	   we would have to map the file's memory. Hence, we do not implement
```
Dennisbonke, line 483: This entire comment block should just be refactored to be C style comments, as should all multiline comments. Imo the expected alignment would probably look like this
```
/*
 * TODO: Verify that the file ....
 * we would have to ....
 * it for now.
 */
 ```

File: kernel/thor/system/pci/pci_discover.cpp [hunk 16]
```diff
@@ -470,97 +518,43 @@ namespace {
 	// mbus object creation and management.
 	// ------------------------------------------------------------------------
 
+	void addPropertyValue(managarm::mbus::CntRequest<KernelAlloc>& req, const char* name,
+		auto&& value) {
```
Dennisbonke, line 522: Pointers and references are declared like `const char *name`, not `const char* name`.

kernel/thor/system/pci/pci_discover.cpp
```diff
@@ -470,97 +518,43 @@ namespace {
...
+		managarm::mbus::Property<KernelAlloc> property(*kernelAlloc);
+		property.set_name(frg::string<KernelAlloc>(*kernelAlloc, name));
+		auto &mutable_item = property.mutable_item().mutable_string_item();
+		mutable_item.set_value(std::move(value));
+		req.add_properties(std::move(property));
+	}
+
+	void addProperty(managarm::mbus::CntRequest<KernelAlloc>& req, const char* name, auto... args) {
```
Dennisbonke, line 530: I think `auto... args` should be `auto ... args` or something like that?

posix/subsystem/src/main.cpp
```diff
@@ -169,10 +169,7 @@ async::result<void> observeThread(std::shared_ptr<Process> self,
 	auto thread = self->threadDescriptor();
 
 	uint64_t sequence = 1;
-	while(true) {
-		if(generation->inTermination)
-			break;
-
+	for(bool killed = false; !killed && !generation->inTermination; ) {
```
Dennisbonke, line 172: what is the advantage of the `for` loop over the `while` loop? Imo the `while` loop makes it very clear that this loops indefinitely until broken out of and that can not be said for the `for` loop.

utils/runsvr/src/main.cpp
```diff
@@ -155,84 +155,109 @@ async::result<void> uploadFile(const char *name) {
 // Freestanding mbus functions.
 // ----------------------------------------------------------------
 
-async::detached asyncMain(const char **args) {
+enum class Action: uint8_t
+{
```
Dennisbonke, line 159: The brace should not be on it's own line.

utils/runsvr/src/main.cpp
```diff
@@ -155,84 +155,109 @@ async::result<void> uploadFile(const char *name) {
...
+	runsvr,
+	run,
+	bind,
+	upload,
+};
+
+static const std::unordered_map<std::string_view, Action> actionMap
+{
```
Dennisbonke, line 167: Same here.

utils/runsvr/src/main.cpp
```diff
@@ -155,84 +155,109 @@ async::result<void> uploadFile(const char *name) {
...
+		}
+	else
 		throw std::runtime_error("Unexpected command for svrctl utility");
-	}
+
+	exit(0);
 }
 
-int main(int, const char **argv) {
+int main(int argc, const char *argv[]) {
```
Dennisbonke, line 260: `argc` is probably unused, thus triggering a warning. We don't want warnings so this should be removed.

Dennisbonke (changes requested):
In general, a few places have the weird `default:;` thing that I'm not entirely sure about. Furthermore, I'm not sure if refactoring all `if`, `else-if` statements to `switch` statements is needed, in some places it gets a bit ugly. For all files I want to comment that the indentation is off, we indent the `case` statements one level relative to the `switch` statement. Also, we only do one statement per line, which is not followed in several places. Lastly, as a general note, I didn't see any major issues with the commit messages, but when in doubt follow [this](https://docs.managarm.org/handbook/contributing/commit-messages.html). As a reference, [this](https://docs.managarm.org/handbook/contributing/coding-style.html) is the coding style I'm referencing to.

utils/runsvr/src/main.cpp
qookei: I also don't see why the `**argv` to `*argv[]` change was done

core/drm/src/core.cpp
```diff
@@ -564,48 +565,57 @@ drm_core::File::ioctl(void *object, managarm::fs::CntRequest req,
...
 		HEL_CHECK(send_resp.error());
-	}else if(req.command() == DRM_IOCTL_GET_CAP) {
+		break;
+	}
+	case DRM_IOCTL_GET_CAP: {
 		helix::SendBuffer send_resp;
 		managarm::fs::SvrResponse resp;
 		
-		if(req.drm_capability() == DRM_CAP_TIMESTAMP_MONOTONIC) {
+		switch(req.drm_capability()) {
```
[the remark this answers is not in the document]
cflaviu: Inner switches should be functions but that is another step.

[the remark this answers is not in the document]
Dennisbonke: Why would that be the next step?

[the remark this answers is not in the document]
qookei: I don't see the benefit of moving it to it's own function, especially since it only has one user.

core/drm/src/core.cpp
```diff
@@ -564,48 +565,57 @@ drm_core::File::ioctl(void *object, managarm::fs::CntRequest req,
...
-	}else if(req.command() == DRM_IOCTL_MODE_GETRESOURCES) {
+		break;
+	}
+	case DRM_IOCTL_MODE_GETRESOURCES: {
 		helix::SendBuffer send_resp;
 		managarm::fs::SvrResponse resp;
 
 		auto &crtcs = self->_device->getCrtcs();
-		for(size_t i = 0; i < crtcs.size(); i++) {
+		for(size_t i = 0; i != crtcs.size(); i++) {
```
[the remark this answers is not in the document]
cflaviu: Just semantics: < has sense when step is not 1 (eg. greater than 1 or variable). In almost all cases (I would say 99%) we are just checking if limit is reached using step 1.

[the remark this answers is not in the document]
Dennisbonke: I still don't see the need for the change but that may just be me.

Dennisbonke:
Also, I'd like to note that this PR should pass the compilation with all warnings at meson warning level 2 enabled without disabling any specific warning and without introducing new ones. At this time, the github ci still has 3 warnings disabled but to my knowledge, (almost) all of those have been fixed by work in #288 and #287, and I'd like to keep it that way.

core/drm/src/core.cpp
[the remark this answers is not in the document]
cflaviu: Guidelines says a function should not exceed one view page. Readability is a benefit.

core/drm/src/core.cpp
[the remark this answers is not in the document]
qookei: Yeah I'd say that's just personal preference, and I don't see the point of changing it in existing files

core/drm/src/core.cpp
cflaviu: Clang-format would fix it instantly.

core/virtio/src/core.cpp
cflaviu: This is why clang-format would be useful.

drivers/libevbackend/src/libevbackend.cpp
cflaviu: Static code analyzers are complaining if no default clause is defined and it helps on code robustness. 
Specifying default:; you simply say other cases will not be taken into account.

Dennisbonke: The default clause I get and agree on, I'm just not familiar with adding a `;` directly behind it, because I always see a separate `break;` or an `assert` or actual code in the default statement, which is why I asked.

qookei: I'd say for default cases that should not be reached we should do something like `assert(!"Unexpected value");` (or maybe throw an exception).

drivers/usb/hcds/xhci/src/main.cpp
cflaviu: Yeah, i used my coding style instinctively :)

drivers/libevbackend/src/libevbackend.cpp
Dennisbonke: Agreed, and even the coding style strongly suggests it.

drivers/usb/hcds/xhci/src/main.cpp
cflaviu: Yeah, i used my coding style instinctively :)

kernel/thor/arch/x86/thor-internal/arch/cpu.hpp
cflaviu: instinctively :)

kernel/thor/generic/servers.cpp
cflaviu: Why using C style in C++ code?

posix/subsystem/src/main.cpp
cflaviu: You know breaks inside while-switch constructs will not break the loop. Using a boolean controlling the loop is a solution to that issue. If you have a better solution please let me know.

utils/runsvr/src/main.cpp
cflaviu: k

utils/runsvr/src/main.cpp
cflaviu: k

utils/runsvr/src/main.cpp
cflaviu: argc should be checked before accessing argv. **argv is useful iterating trough argument values using a pointer. I never seen that.

cflaviu:
> Also, I'd like to note that this PR should pass the compilation with all warnings at meson warning level 2 enabled without disabling any specific warning and without introducing new ones. At this time, the github ci still has 3 warnings disabled but to my knowledge, (almost) all of those have been fixed by work in #288 and #287, and I'd like to keep it that way.

If new warnings were introduced on this PR, they should be fixed. Old warnings should be fixed in other PR.

utils/runsvr/src/main.cpp
qookei: `**argv` and `*argv[]` are fully equivalent in this case, as seen by the following code compiling without errors:
```cpp
#include <type_traits>

void foo(char **a, char *b[]) {
    static_assert(std::is_same_v<decltype(a), decltype(b)>);
}
```

cflaviu: I didn't say those types are not equivalent. But are we using argv incrementing the pointer? Or always we use it with an index. This is the point.

drivers/libevbackend/src/libevbackend.cpp
cflaviu: This is a lesson learned after fixing QA C++ warnings : )

drivers/kbd/src/main.cpp
cflaviu: ok

utils/runsvr/src/main.cpp
qookei: Doesn't matter, you can index pointers as well.

PR #303, opened 2021-05-18
block/ahci: Various fixes

64:
Fixes various AHCI issues to allow us to run on real hardware. Fixes managarm#323.

Draft until I can fix all the remaining problems.

### DIFF
--- a/drivers/block/ahci/src/command.cpp
+++ b/drivers/block/ahci/src/command.cpp
@@ -7,10 +7,6 @@
 
 #include "command.hpp"
 
-namespace {
-	constexpr bool logCommands = false;
-}
-
 Command::Command(uint64_t sector, size_t numSectors, size_t numBytes, void *buffer,
 		CommandType type) : sector_{sector}, numSectors_{numSectors}, numBytes_{numBytes},
 	buffer_{buffer}, type_{type}, event_{} {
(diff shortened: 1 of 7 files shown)


avdgrinten:
What is the status of this PR?

64:
When using queue depth 1, we hit the asynchronous nack issue which we believe to be an XHCI bug, but I’d be hesitant to merge until we've figured that out.

When using queue depth > 1 we hit some AHCI interface error (which may or may not be related to the XHCI issue).

64:
I've rebased this PR but we still have the same IRQ NACK issues, at least on my hardware (potentially still related to XHCI). Maybe someone else could test on a spare disk

avdgrinten:
Does AHCI support MSIs? If yes, you could add MSI support to the driver (which should be only a few lines of code changes); this would isolate AHCI's IRQs from XHCI's.

64:
This PR should be mostly good to go, but there are some remaining issues:

1. `libarch` asserts that `alignof(commandTable) <= alignof(max_align_t)` [here](https://github.com/managarm/libarch/blob/923eb13477621d97503d1d2e8947801aa60ed6bf/include/arch/dma_structs.hpp#L329) which we deliberately break
2. https://github.com/managarm/managarm/issues/322 prevents booting in a lot of setups

Also I'm not sure if hardcoding MSI index 0 will be a problem in practice. I guess we'll have to see

64:
Okay, this is ready to merge now

avdgrinten (approved):
LGTM, thank you!

PR #304, opened 2021-05-21
Add NVMe block driver

Jimx-:


drivers/block/nvme/src/command.cpp
```diff
@@ -0,0 +1,84 @@
...
+        command_.readWrite.dataPtr.prp2 = 0;
+        return;
+    }
+    size -= pageSize - offset;
+    virtStart += pageSize - offset;
+    HEL_CHECK(helPointerPhysical((void*)virtStart, &dmaAddr));
+
+    if (size <= pageSize) {
+        command_.readWrite.dataPtr.prp1 = convert_endian<endian::little, endian::native>(prp1);
+        command_.readWrite.dataPtr.prp1 = convert_endian<endian::little, endian::native>(dmaAddr);
```
thomtl, line 49: You're assigning to thesame thing twice here?

drivers/block/nvme/src/command.cpp
```diff
@@ -0,0 +1,84 @@
...
+        prpList[i++] = convert_endian<endian::little, endian::native>(dmaAddr);
+        virtStart += pageSize;
+        HEL_CHECK(helPointerPhysical((void*)virtStart, &dmaAddr));
+
+        if (size <= pageSize) break;
+        size -= pageSize;
+    }
+
+    command_.readWrite.dataPtr.prp1 = convert_endian<endian::little, endian::native>(prp1);
+    command_.readWrite.dataPtr.prp1 = convert_endian<endian::little, endian::native>(prp2);
```
thomtl, line 83: Once again assigning to thesame thing twice? Shouldn't the second one be dataPtr.prp2

avdgrinten:
Thanks a lot for the PR! An NVMe driver is a very welcome addition to Managarm. You can expect the PR get a review soon. :smile: 

As a minor comment: the indentation is off in some places (tabs vs. spaces). It'd be nice to fix that.

drivers/block/nvme/src/controller.cpp
```diff
@@ -0,0 +1,283 @@
...
+    if ((co_await identifyController(idCtrl)).first != 0)
+        co_return;
+
+    nn = convert_endian<endian::little>(idCtrl.nn);
+
+    auto nsList = arch::dma_array<uint32_t>{nullptr, 1024};
+    int numLists = (nn + 1023) >> 10;
+    unsigned int prev = 0;
+    for (auto i = 0; i < numLists; i++) {
+        if ((co_await identifyNamespaceList(prev, nsList.data())).first != 0)
```
thomtl, line 251: The identify command for namespaces actually wasn't included on NVMe 1.0 controllers (Like the one VMWare emulates), only since 1.1.0 has it been supported, so this will break on 1.0 machines

Jimx-:
I have pushed some new commits to fix the issues. Do I need to replace all spaces with tabs for indentation in the code?

64:
It might also be nice to add the required qemu options in vm-utils.py to boot from NVMe (this will require a PR to bootstrap-managarm)

Jimx-:
> It might also be nice to add the required qemu options in vm-utils.py to boot from NVMe (this will require a PR to bootstrap-managarm)

I tried that but qemu's bios does not seem to support booting from nvme. If I have only one nvme drive in the qemu options (like                 qemu_args += ['-device', 'nvme,drive=boot-drive,serial=1234']) then qemu won't boot at all, complaining there is no bootable disk. When testing the driver, I need to set up both the nvme disk and another virtio-blk device for booting (it will crash later in device node creation because both the virtio driver and the nvme driver advertise the same sda0 disk but blockfs can read the FS information from nvme so the driver part should be ok). Don't know if I did something wrong here.

qookei:
> Do I need to replace all spaces with tabs for indentation in the code?

Yes, we use tabs for indentation. Also, pointers and references should be next to the name, not the type (`T *x` and `T &x`, not `T* x` and `T& x`).

drivers/block/nvme/src/command.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+    static size_t pageSize = getpagesize();
+
+    uintptr_t virtStart = reinterpret_cast<uintptr_t>(view.data());
+    auto offset = virtStart % pageSize;
+
+    if (offset + view.size() <= pageSize * 2) {
+        // Inline
+        uintptr_t physical;
+        HEL_CHECK(helPointerPhysical(view.data(), &physical));
```
qookei, line 19: We have convenience wrappers for this (`helix::ptrToPhysical` and `helix::addressToPhysical`), and it could be rewritten as:
```cpp
command_.common.dataPtr.prp1 = convert_endian<endian::little, endian::native>(
                helix::ptrToPhysical(view.data()));
```
Same applies to other places where `helPointerPhysical` is used.

drivers/block/nvme/src/controller.cpp
```diff
@@ -0,0 +1,310 @@
...
+
+namespace cc {
+constexpr int enable       = 1 << 0;
+constexpr int iosqes_shift = 16;
+constexpr int iocqes_shift = 20;
+}
+
+namespace csts {
+constexpr int ready       = 1 << 0;
+}
```
qookei, line 40: For small private namespaces we should indent their contents by 1 level.

Some of these could be defined using `arch::field`s instead. To do that, turn the associated register into a `arch::bit_register`, and add an `arch::field` for each field of a register you care about, for example:
```cpp
// first argument is the bit offset, second argument is the bit width
constexpr arch::field<uint64_t, uint16_t> mqes{0, 16};

constexpr arch::field<uint64_t, uint8_t> iocqes{20, 4};
constexpr arch::field<uint64_t, uint8_t> iosqes{16, 4};
constexpr arch::field<uint64_t, bool> enable{0, 1};
```
Then, to get the value of the field from the register, you'd do something like:
```cpp
auto cap = regs_.load(regs::cap);
auto val = cap & cap::mqes;
```
and to update the field values you'd do:
```cpp
auto cc = regs_.load(regs::cc);
// This preserves the other values in cc and only replaces the fields we specify
// Use | instead of / if you don't want to clear out the bits used by the fields we modify
auto val = cc / cc::iosqes(6) / cc::iocqes(4) / cc::enable(true);
regs_.store(regs::cc, val);
```

drivers/block/nvme/src/controller.cpp
```diff
@@ -0,0 +1,310 @@
...
+
+async::result<void> Controller::enable() {
+    auto cc = regs_.load(regs::cc);
+    cc |= 6 << flags::cc::iosqes_shift;
+    cc |= 4 << flags::cc::iocqes_shift;
+    cc |= flags::cc::enable;
+    regs_.store(regs::cc, cc);
+
+    co_await waitStatus(true);
+    co_return;
```
qookei, line 102: Explicit `co_return` is not needed if the function returns a `async::result<void>` and is already a coroutine (already has a `co_await` somewhere).

Also, these functions don't need to be coroutines, you can just turn this function, `waitStatus`, and other functions that only read from/write to registers into regular functions (drop the `async::result<>` from the return type and drop all the `co_await`s), since they don't actually perform any async operations.

Dennisbonke:
In general, try to follow [this](https://docs.managarm.org/handbook/contributing/coding-style.html) for the coding style. I didn’t see many issues apart from what others have already pointed out but I didn’t do a thorough review to be honest.

drivers/block/nvme/src/namespace.cpp
```diff
@@ -0,0 +1,46 @@
+#include <arch/bit.hpp>
+
+#include "namespace.hpp"
+#include "controller.hpp"
+
+Namespace::Namespace(Controller* controller, unsigned int nsid, int lbaShift) : BlockDevice{(size_t)1 << lbaShift}, controller_(controller),
+                                                                        nsid_(nsid), lbaShift_(lbaShift) {
+}
```
qookei, line 8: Put the member initializer list on it's own line, on the same level as `Namespace::Namespace`. Same applies to other places where this is on the same line.

drivers/block/nvme/src/queue.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+    cqes_ = reinterpret_cast<spec::CompletionEntry *>(window);
+    memset(cqes_, 0, cqSize);
+
+    HEL_CHECK(helAllocateMemory(sqSize, kHelAllocContinuous, nullptr, &memory));
+    HEL_CHECK(helMapMemory(memory, kHelNullHandle, nullptr,
+            0, sqSize, kHelMapProtRead | kHelMapProtWrite, &window));
+    HEL_CHECK(helCloseDescriptor(kHelThisUniverse, memory));
+
+    sqCmds_ = reinterpret_cast<spec::CompletionEntry *>(window);
```
qookei, line 32: Here using a `arch::dma_array<spec::CompletionEntry>` (or at least a `arch::dma_buffer`) would be better than manually allocating the memory and having the `reinterpret_cast`.

drivers/block/nvme/src/queue.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+    co_return;
+}
+
+int Queue::handleIrq() {
+    using arch::convert_endian;
+    using arch::endian;
+
+    int found = 0;
+    volatile spec::CompletionEntry* cqe = &cqes_[cqHead_];
```
qookei, line 49: Why do we need `volatile` here?

drivers/block/nvme/src/queue.hpp
```diff
@@ -0,0 +1,58 @@
...
+        struct stl_allocator {
+        void *allocate(size_t size) {
+            return operator new(size);
+        }
+
+        void deallocate(void *p, size_t) {
+            return operator delete(p);
+        }
+    };
+    async::queue<std::unique_ptr<Command>, stl_allocator> pendingCmdQueue_;
```
qookei, line 48: Include `<frg/std_compat.hpp>` and use `frg::stl_allocator` instead.

drivers/block/nvme/src/controller.hpp
```diff
@@ -0,0 +1,56 @@
...
+#include <async/result.hpp>
+#include <helix/memory.hpp>
+#include <protocols/hw/client.hpp>
+
+#include "queue.hpp"
+#include "namespace.hpp"
+
+class Controller {
+
+public:
```
qookei, line 13: Prefer `struct Foo {` over `class Foo { public:`

qookei (changes requested):
Over all looks good, apart from the comments (most of which are mainly code-style related).

drivers/block/nvme/src/queue.cpp
Jimx-: But dma_array does not guarantee the allocated buffer is page-aligned, right?

drivers/block/nvme/src/queue.cpp
Jimx-: Not sure if the cpu can read stale values if volatile is dropped because the device can be posting CQEs at the same time.

drivers/block/nvme/src/controller.hpp
Jimx-: But Controller also has private members.

qookei: `struct`s can also have private members just like classes.

drivers/block/nvme/src/queue.cpp
qookei: Hm that is a fair point, does NVMe require that the queues are page aligned?

Jimx-: Yes, if I use dma_array then the create CQ command fails with base address misaligned error.

qookei: Alright, I think for now this will be fine then until we have some way to allocate memory with a specified alignment.

qookei (approved):
LGTM now

64:
someone needs to approve the CI run

avdgrinten:
Alright, I think we can merge this now. I will squash the commits into one. Again, thank you very much for the PR!

PR #305, opened 2021-05-23
Add simple stack trace, fix lai bug and block device changes

Geertiebear:


drivers/libblockfs/src/libblockfs.cpp
```diff
@@ -600,6 +601,26 @@ async::detached runDevice(BlockDevice *device) {
...
+
+		mbus::Properties descriptor {
+			{"unix.devtype", mbus::StringItem{"block"}},
+			{"unix.blocktype", mbus::StringItem{"disk"}}
+		};
+
+		auto handler = mbus::ObjectHandler{}
+		.withBind([] () -> async::result<helix::UniqueDescriptor> {
+			async::promise<helix::UniqueDescriptor> promise;
+			return promise.async_get();
```
avdgrinten, line 616: This would hang forever. We should at least print an error message.

kernel/thor/generic/debug.cpp
```diff
@@ -177,6 +178,10 @@ void PanicSink::operator() (const char *msg) {
 		logProcessor.print(msg);
 		logProcessor.print('\n');
 	}
+	urgentLogger() << "STACKTRACE:" << frg::endlog;
```
avdgrinten, line 181: Why all caps?

drivers/libblockfs/src/libblockfs.cpp
Geertiebear: Ah okay, that I can do.

kernel/thor/generic/debug.cpp
Geertiebear: For no particular reason, I can also change it to normal caps.

Geertiebear:
Fixed the comments.

avdgrinten (approved):
LGTM now, thank you for the PR!

PR #312, opened 2021-05-25
meson.build: add -Dkernel_frame_pointers

Geertiebear:


kernel/thor/meson.build
```diff
@@ -111,6 +111,10 @@ if log_alloc
 	args = [ '-fno-omit-frame-pointer', '-DKERNEL_LOG_ALLOCATIONS' ]
```
avdgrinten, line 111: We should also set `-DTHOR_HAS_FRAME_POINTERS` here, as we get that for free in this case.

Geertiebear: Fixed.

avdgrinten:
LGTM, thank you.

PR #316, opened 2021-05-27
posix: Neofetch fixes

64:
- posix/exec: Fix order of arguments in shebang
- posix: Ignore illegal truncate targets


avdgrinten (approved):
LGTM, thanks for the PR.

PR #317, opened 2021-05-28
kernel: Refactored scrubStack arm and x86 functions; improvements

cflaviu:


kernel/thor/arch/x86/cpu.cpp
```diff
@@ -35,6 +35,18 @@ bool FaultImageAccessor::allowUserPages() {
...
 // --------------------------------------------------------
 
+constexpr uint16_t fcw_mask =
+	(1 << 0) |	  // IM
+	(1 << 1) |	  // DM
+	(1 << 2) |    // ZM
+	(1 << 3) |	  // OM
+	(1 << 4) |	  // UM
+	(1 << 5) |	  // PM
+	(0b11 << 8);  // PC
```
Dennisbonke, line 45: Indentation seems off for `ZM` and `PC`?

Dennisbonke (commented):
Looks good to me apart from 2 small things.

kernel/thor/arch/x86/cpu.cpp
cflaviu: I guess it's due tabs used. Opening the file in text editor it looks fine.

kernel/thor/arch/arm/cpu.cpp
```diff
@@ -133,44 +133,33 @@ void workOnExecutor(Executor *executor) {
...
 	assert(top >= bottom);
 	cleanKasanShadow(cont.sp, top - bottom);
 	// Perform some sanity checking.
-	validateKasanClean(reinterpret_cast<void *>(bottom & ~(kPageSize - 1)),
-			bottom & (kPageSize - 1));
+	validateKasanClean(reinterpret_cast<void *>(bottom & ~(kPageSize - 1)), bottom & (kPageSize - 1));
+}
+
+template <typename Accessor>
+void scrubStackAccessor(const Accessor& accessor, const Continuation& cont) {
```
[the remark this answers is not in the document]
cflaviu: solved

kernel/thor/arch/x86/cpu.cpp
64: There are some tabs in the middle of the line (after `|`)

cflaviu: Well but coding style says "Indentation uses tabs" instead of spaces. 
I would prefer only spaces for indentation avoiding such cases but this is against coding style.

Dennisbonke: While this is indeed true, in cases like this we can and probably should align the comments with spaces

64: Correct me if I'm wrong but I always assumed that tabs should happen at the start of the line and then if things need to be aligned in the middle of the line we do that with spaces

cflaviu: There are two indentations, one before ( and another before //. Perhaps the coding style should say use tabs only on the first indentation but it's weird because when you press tab key to indent, either tab char, either several spaces are inserted. Tab insertion is not contextual sensitive switching when necessary between tab char and spaces.

cflaviu: I added spaces (https://pasteboard.co/K4MZhUg.png) but here the rendering is still wrong.

kernel/thor/arch/arm/cpu.cpp
```diff
@@ -133,44 +133,33 @@ void workOnExecutor(Executor *executor) {
...
 	// Perform some sanity checking.
-	validateKasanClean(reinterpret_cast<void *>(bottom & ~(kPageSize - 1)),
-			bottom & (kPageSize - 1));
+	validateKasanClean(reinterpret_cast<void *>(bottom & ~(kPageSize - 1)), bottom & (kPageSize - 1));
+}
+
+template <typename Accessor>
+void scrubStackAccessor(const Accessor &accessor, const Continuation &cont) {
+	scrubStackInternal(reinterpret_cast<uintptr_t>(accessor.frameBase()), cont);
+}
```
avdgrinten, line 147: I think this function is unnecessary, it's probably less mental overhead to just put the `scrubStackInternal` call into each call site (the reinterpret_cast is not too bad.

kernel/thor/arch/arm/cpu.cpp
```diff
@@ -133,44 +133,33 @@ void workOnExecutor(Executor *executor) {
...
-	auto top = reinterpret_cast<uintptr_t>(accessor.frameBase());
+void scrubStackInternal(uintptr_t top, Continuation cont) {
 	auto bottom = reinterpret_cast<uintptr_t>(cont.sp);
 	assert(top >= bottom);
 	cleanKasanShadow(cont.sp, top - bottom);
 	// Perform some sanity checking.
-	validateKasanClean(reinterpret_cast<void *>(bottom & ~(kPageSize - 1)),
-			bottom & (kPageSize - 1));
+	validateKasanClean(reinterpret_cast<void *>(bottom & ~(kPageSize - 1)), bottom & (kPageSize - 1));
+}
```
avdgrinten, line 142: This function is now identical on x86_64 and aarch64, right? I think we could move it to `kasan.hpp` (as an inline function) since it only deals with KASAN-related things.

avdgrinten: We can also rename it to `scrubStackFrom` to give it a more descriptive name.

kernel/thor/arch/x86/cpu.cpp
```diff
@@ -35,6 +35,18 @@ bool FaultImageAccessor::allowUserPages() {
 // Executor
 // --------------------------------------------------------
 
+constexpr uint16_t fcw_mask =
```
avdgrinten, line 38: Rename to `fcwInitializer`. Same for `mxcsr_mask` -> `mxcsrInitializer`.

avdgrinten: These variables can be `static`.

avdgrinten (commented):
A few comments, this can be merged afterwards.

kernel/thor/arch/x86/cpu.cpp
cflaviu: OK

kernel/thor/arch/arm/cpu.cpp
cflaviu: I prefer following DRY principle.

kernel/thor/arch/arm/cpu.cpp
cflaviu: OK

kernel/thor/arch/arm/cpu.cpp
avdgrinten: There is no violation of that principle here because no logic is duplicated, only a function call.

cflaviu: The logic is to reinterpret the result of a specific function and this logic was the same in 3 cases.

avdgrinten: That's only a tiny bit of logic, and duplicating it is favorable to introducing another (mental) indirection; the code is clearer if we avoid the indirection.

cflaviu: I disagree. Inheritance introduces mental overhead and indirection but it is preferable than duplicating code.

cflaviu: DRY is very very helpful. 

Similar to his, I had to add 2 boolean flags in two modules - a minor change. In one module, there was a base class I refactored from duplicated code. Some colleagues complained that the resulted code is harder to follow (as you said mental overhead and indirection).

 In that base class I added those 2 flags covering 4 test cases. 

 In another module there was 7 classes with similar code but no base class. So I have to add 14 new flags. The bad thing  was I had to cover all cases with unit tests with painfully 28 test cases instead of just 4 test cases. Wasted time...

avdgrinten (approved):
LGTM now, thank you!

PR #319, opened 2021-05-28
build: Enable several warnings

Dennisbonke:
This PR is basically the culmination of work done in PR #287 and #288, and enables the warnings the PRs mentioned fixed. There might be a few more cases that produce warnings now in code introduced after the merging of #287 and #288, or in the aarch64 port, as I did not build that. If that turns out to be the case, a follow-up PR will fix those warnings.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -67,13 +67,6 @@ if not build_kernel and not build_drivers and not build_tools
 	subdir_done()
 endif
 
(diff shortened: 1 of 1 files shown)


avdgrinten (approved):
LGTM, thank you for resolving all the warnings.

PR #325, opened 2021-06-01
drivers/ps2-hid: Send port 2 byte properly

64:


64:
Closed as this turns out to be some race condition and not a PS/2 driver issue.

PR #273, opened 2021-04-15
drivers/libblockfs: Fix reading long symlinks on ext2 filesystems

Dennisbonke:
This PR fixes #161

PR #336, opened 2021-06-06
Various fixes for hexchat

Dennisbonke:
This PR adds various fixes in order to run `hexchat` on Managarm. For now, this consist of better error handling in `sendMsg` and handling unknown address families gracefully in the socket request. I fully expect more to come and as such, this PR is a draft.

The mlibc counterpart to this PR is managarm/mlibc#273

### DIFF
--- a/protocols/fs/src/server.cpp
+++ b/protocols/fs/src/server.cpp
@@ -886,6 +886,51 @@ async::detached handlePassthrough(smarter::shared_ptr<void> file,
 					helix_ng::sendBuffer(ser.data(), ser.size())
 				);
 				HEL_CHECK(send_resp.error());
+			} else if(res.error() == Error::hostUnreachable) {
+				resp.set_error(managarm::fs::Errors::HOST_UNREACHABLE);
+
+				auto ser = resp.SerializeAsString();
+				auto [send_resp] = co_await helix_ng::exchangeMsgs(
+					conversation,
+					helix_ng::sendBuffer(ser.data(), ser.size())
+				);
+				HEL_CHECK(send_resp.error());
+			} else if(res.error() == Error::accessDenied) {
+				resp.set_error(managarm::fs::Errors::ACCESS_DENIED);
+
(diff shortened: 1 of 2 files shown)


protocols/fs/src/server.cpp
```diff
@@ -889,6 +889,51 @@ async::detached handlePassthrough(smarter::shared_ptr<void> file,
...
+
+				auto ser = resp.SerializeAsString();
+				auto [send_resp] = co_await helix_ng::exchangeMsgs(
+					conversation,
+					helix_ng::sendBuffer(ser.data(), ser.size())
+				);
+				HEL_CHECK(send_resp.error());
+			} else if(res.error() == Error::addressNotAvailable) {
+				resp.set_error(managarm::fs::Errors::ADDRESS_NOT_AVAILABLE);
+
```
cflaviu, line 936: I would suggest creating a mapping function Error::* -> managarm::fs::Errors::* that would return an optional or a boolean. If mapping is succeesful, error response is set and exchangeMsgs is called avoiding code duplication.

avdgrinten: :+1: for the mapping function.

But let's make it map one type to the other and panic (= fail an assertion) if the argument is invalid; there is no reasonable way to handle errors inside the mapping function.

cflaviu: Just return false or {} in case of mismatching.

avdgrinten: But what do we send to the client in that case? There is no meaningful way to handle that (if we send a "internal error" message to the client, the client will just crash). Passing an invalid value is a contract violation, not a runtime error.

cflaviu: There is already and else handling the mismatch.

```
managarm::fs::Errors resultedError;
if (map(inputError, resultedError)) {
   // set error, exchange msgs
} else {
  // reuse existing "else" code
}
```

no92 (approved):

PR #339, opened 2021-06-06

cflaviu:


cflaviu:
I noticed lot of "Debug fault at ...." messages running this build but it doesn't hang. I tried to identify what file changed causes this but I could not find a specific file.

Dennisbonke:
In several places the [coding style](https://docs.managarm.org/handbook/contributing/coding-style.html) is not followed. Please fix the inconsistencies. If you want, I can point them out manually but that would be an ungodly amount of work.

cflaviu:
> In several places the [coding style](https://docs.managarm.org/handbook/contributing/coding-style.html) is not followed. Please fix the inconsistencies. If you want, I can point them out manually but that would be an ungodly amount of work.

Because it's too much to change just for coding style I really need a .clang-format file to use. Could you review & adapt the .clang-format file I sent in one thread from Discussions?

PR #351, opened 2021-06-11
Remove illegalOperationTarget as much as possible

Dennisbonke:
This PR is the managarm counterpart of managarm/mlibc#250. It can't be merged without merging the other.

This PR removes almost all uses of `illegalOperationTarget` and adds a `isDirectory` error return.

Blocked on managarm/mlibc#250, but should be merged at the same time to avoid build failures.


posix/subsystem/src/main.cpp
```diff
@@ -2172,8 +2125,9 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 				auto resolveResult = co_await resolver.resolve(
 						resolvePrefix | resolveNoTrailingSlash);
 				if(!resolveResult) {
-					if(resolveResult.error() == protocols::fs::Error::illegalOperationTarget) {
-						co_await sendErrorResponse(managarm::posix::Errors::ILLEGAL_OPERATION_TARGET);
+					if(resolveResult.error() == protocols::fs::Error::isDirectory) {
+						// TODO: Verify additional constraints for sending EISDIR.
```
avdgrinten, line 2129: What additional constraints?

posix/subsystem/src/main.cpp
```diff
@@ -2224,8 +2178,9 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 			}else{
 				auto resolveResult = co_await resolver.resolve();
 				if(!resolveResult) {
-					if(resolveResult.error() == protocols::fs::Error::illegalOperationTarget) {
-						co_await sendErrorResponse(managarm::posix::Errors::ILLEGAL_OPERATION_TARGET);
+					if(resolveResult.error() == protocols::fs::Error::isDirectory) {
+						// TODO: Verify additional constraints for sending EISDIR.
```
avdgrinten, line 2182: See above.

protocols/fs/include/protocols/fs/common.hpp
```diff
@@ -29,8 +29,9 @@ enum class Error {
 	notConnected = 17,
 	alreadyExists = 18,
 	illegalOperationTarget = 19,
-	noSpaceLeft = 20,
+	noSpaceLeft = 21,
```
avdgrinten, line 32: Why is this changed?

posix/subsystem/src/main.cpp
```diff
@@ -3483,14 +3430,15 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
 			if(type == VfsType::charDevice || type == VfsType::blockDevice) {
 				dev.first = major(req->device());
 				dev.second = minor(req->device());
 
 				auto result = co_await parent->mkdev(resolver.nextComponent(), type, dev);
 				if(!result) {
 					if(result.error() == Error::illegalOperationTarget) {
-						co_await sendErrorResponse(managarm::posix::Errors::ILLEGAL_OPERATION_TARGET);
+						co_await sendErrorResponse(managarm::posix::Errors::ILLEGAL_ARGUMENTS);
 						continue;
```
avdgrinten, line 3442: Why do we change this? Under what conditions does `mkdev` emit `illegalOperationTarget`?

avdgrinten: Same for `mkfifo` / `mksocket` below.

avdgrinten (commented):
Why is the `isDirectory` case added to only some cases? Doesn't it need to be added to all cases where `resolveNoTrailingSlash` is specified?

posix/subsystem/src/main.cpp
Dennisbonke: From `man 2 open`

```
EISDIR
pathname refers to a directory and the access requested involved writing (that is, O_WRONLY or O_RDWR is set).
```

So it seems that we need to verify that too before returning EISDIR. The same goes for the other comment related to this below.

protocols/fs/include/protocols/fs/common.hpp
Dennisbonke: `noSpaceLeft` and `notDirectory` have the same value, which is wrong. This fixes that.

posix/subsystem/src/main.cpp
Dennisbonke: My initial guess would be that all three of these functions return `illegalOperationTarget` when called on file systems that don’t implement it.

Dennisbonke:
It has been a while since I looked at this but I believe that the one place I didn’t add the EISDIR error return is where it’s actually valid to get a directory back?

avdgrinten:
Well, if `resolveNoTrailingSlash` is specified, `resolve()` fails with `isDirectory` if the end of the path has a slash. So we need to add handling for the error in all these cases (there aren't many, though). Or is `resolveNoTrailingSlash` specified incorrectly somewhere?

posix/subsystem/src/main.cpp
avdgrinten: Hm, then we should apply `resolveNoTrailingSlash` only in these cases.

Dennisbonke:
Seems like the one for `OPENAT` might be wrong, `MKFIFOAT` and `SYMLINK` I’m not sure about.

posix/subsystem/src/main.cpp
Dennisbonke: That will be difficult, considering we don't do modes in `OPENAT` and friends at all yet. It's not wired up at all. We should revisit these errors once we actually do have modes wired up but I'd like to not let this PR (and more notably the complementing mlibc PR) wait on that.

avdgrinten:
Okay, I think we can merge this as-is and deal with the edge cases in future PRs.

PR #370, opened 2021-07-18
Do not copy libraries from system-gcc to the initrd

Dennisbonke:
This PR removes the usage of `libstdc++.so.6` and `libgcc_s.so.1` provided by the `system-gcc` tool in the initrd generation script, and replaces them with the one provided in the system-root, as installed by the `gcc` package. Furthermore, it removes 2 duplicate lines related to zlib.

Blocked on managarm/bootstrap-managarm#126


avdgrinten:
LGTM.

PR #371, opened 2021-07-22
sysfs and libblockfs changes to support udev's by-uuid

Geertiebear:


drivers/libblockfs/src/libblockfs.cpp
```diff
@@ -480,10 +482,98 @@ constexpr protocols::fs::NodeOperations nodeOperations{
...
+		rsp.set_error(managarm::fs::Errors::NOT_A_TERMINAL);
+
+		auto ser = rsp.SerializeAsString();
+		auto [send_resp] = co_await helix_ng::exchangeMsgs(
+				conversation,
+				helix_ng::sendBuffer(ser.data(), ser.size())
+		);
+		HEL_CHECK(send_resp.error());
+	} else {
+		throw std::runtime_error("Unknown ioctl() with ID " + std::to_string(req.command()));
```
avdgrinten, line 560: This should dismiss the request instead of killing the server.

drivers/libblockfs/src/raw.hpp
```diff
@@ -0,0 +1,31 @@
...
+	async::result<void> init();
+
+	async::detached manageMapping();
+
+	uint64_t offset;
+	BlockDevice *device;
+	HelHandle backingMemory;
+	HelHandle frontalMemory;
+	helix::Mapping fileMapping;
+	Flock flock;
```
avdgrinten, line 26: The `Flock` should be a per-open-file and not a per-device object, right? There could be two open files that refer to the same device. Do we handle that situation at all right now?

avdgrinten:
I think we discussed in Discord that the `requestSize` mechanism is premature optimization here. Can we get rid of that in favor of a virtual `getSize`?

Geertiebear:
Ah, yeah sure.

avdgrinten (approved):
LGTM.

avdgrinten:
This has conflicts that need to be resolved. While resolving these, you can also remove the debugging `printf` from the raw device.

PR #389, opened 2021-08-18
meta: add git mailmap for changed emails

ArsenArsen:
this also applies to other repos, if we conclude it's a fine addition


avdgrinten:
We found the consensus to merge this on Discord.

PR #390, opened 2021-08-20
Refactor RENAME request to proper bragi

Dennisbonke:
This PR changes the RENAME request from `posix-subsystem` to `libblockfs` to use a seperate bragi message, which allows for a tail and longer paths to be passed in it to `libblockfs`. This fixes a long standing issue with multiple programs when they call a rename from a long path to another long path, most notably `python3` on certain operations.


qookei (approved):
LGTM

PR #393, opened 2021-09-02
posix: Create /dev/shm and mount a tmpfs to it

Dennisbonke:
This allows functions like `shm_open()` to work, a requirement for `gtk+-3`.


avdgrinten (approved):
LGTM.

PR #320, opened 2021-05-28
posix: Fix unused parameter warning

Dennisbonke:
Should be merged before #319 to minimize the amount of warnings given.

PR #397, opened 2021-10-03
Make open() return error codes instead of fail on assert.

ikbenlike:
The primary purpose of these changes is to make `open()` return (informative) error codes instead of fail on assert. This means that an application won't unexpectedly hang if it opens something if it attempts to open a file with unsupported flags, for example. 

This PR is accompanied by [mlibc PR #320.](https://github.com/managarm/mlibc/pull/320)


posix/subsystem/src/main.cpp
```diff
@@ -60,12 +60,12 @@
...
 
 namespace {
-	constexpr bool logRequests = false;
+	constexpr bool logRequests = true;
 	constexpr bool logPaths = false;
 	constexpr bool logSignals = false;
 	constexpr bool logCleanup = false;
 
-	constexpr bool debugFaults = false;
+	constexpr bool debugFaults = true;
```
avdgrinten, line 68: These debugging flags should be kept off on the main branch since they cause quite a bit of log spam.

posix/subsystem/src/subsystem/input.cpp
```diff
@@ -70,7 +70,8 @@ struct Device final : UnixDevice, drvcore::ClassDevice {
 
 async::result<std::string> CapabilityAttribute::show(sysfs::Object *object) {
 	auto device = static_cast<Device *>(object);
-	auto file = co_await device->open(nullptr, nullptr, 0);
+	auto fileMaybe = co_await device->open(nullptr, nullptr, 0);
+	auto file = fileMaybe.value(); //TODO: properly handle error returns here
```
avdgrinten, line 74: The proper way to handle this is letting `show()` return a frg::expected. That's probably not *that* much work since there are only few sysfs attributes defined.

avdgrinten (commented):
Is the first commit still WIP? If yes, it should not be PRd (yet) ;).

posix/subsystem/src/main.cpp
ikbenlike: Whoops, yeah, I forgot about those since I was still working on something else.

posix/subsystem/src/subsystem/input.cpp
ikbenlike: I'll change that asap

ikbenlike:
> Is the first commit still WIP? If yes, it should not be PRd (yet) ;).

It's no longer WIP, I perhaps should have explicitly mentioned this in a later commit message

robertlipe:
Just as a fly-by, I think if you use GitHub's Squash & Merge button, it'll remove the "WIP" flag, so it'll no longer show as a draft for the reviewer. It looks like one of the comments indicates it's no longer WIP, but the flag is still set because at least one of the commits contains WIP. Squashing those gives you the opportunity to remove that word. Hopefully that'll get this unstuck in the reviewer queue. Good luck!

borrrden:
FYI I rebased these changes onto the current master (c022c53a1f74bed656a4769296d35a66988f3d20) in my fork.  If you think this PR is ok as-is for review then I will open a new one but I think there is still some outstanding feedback here.

ikbenlike:
I also rebased the commits on the latest master. Everything seems to work fine so I'll try to complete work on this & clean up the commit history as soon as possible.

ikbenlike:
This PR is now ready for review. I've not yet made it so errors are returned in `posix/subsystem/src/subsystem/input.cpp`, as I don't really know how the code in there works. At the moment that just freezes if an error is returned, but that should probably be improved. I can do that if it's deemed necessary for this PR to be merged, but it may be better done by someone who knows a bit about the relevant code. The commit history should probably also be cleaned up, but I'll wait until I have gotten feedback before I do that.

posix/subsystem/src/main.cpp
```diff
@@ -2472,7 +2475,23 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+			
+			auto ttynameResult = co_await file->ttyname();
+			if(!ttynameResult) {
+			    assert(ttynameResult.error() == Error::notTerminal);
+			    co_await sendErrorResponse(managarm::posix::Errors::NOT_A_TTY);
+			    continue;
+			}
+			resp.set_path(ttynameResult.value());
+
+			std::cout << "\e[31mposix: ttynameResult: " << ttynameResult.value() << "\e[39m" << std::endl;
```
qookei, line 2493: Is this a debug print? Should we get rid of it?

qookei (commented):
LGTM, just that one small comment

posix/subsystem/src/main.cpp
ikbenlike: that is a debug print yes, I'll remove it

avdgrinten (approved):

qookei (approved):
LGTM

PR #400, opened 2021-10-22
Remove uses of run_queue

qookei:
Depends on managarm/libasync#9

### DIFF
--- a/kernel/thor/generic/service.cpp
+++ b/kernel/thor/generic/service.cpp
@@ -574,9 +574,6 @@ namespace posix {
 					continue;
 				}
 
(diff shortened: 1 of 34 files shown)


kernel/thor/generic/service.cpp
```diff
@@ -574,16 +574,17 @@ namespace posix {
 					continue;
 				}
 
-				if(!(req->flags() & 4)) // MAP_FIXED.
-					assert(!"TODO: implement non-fixed mappings");
```
avdgrinten, line 578: It's good to get rid of the assertion here, but shouldn't we return an error for unsupported flags?

PR #405, opened 2021-12-10
Procfs improvements

Geertiebear:
Added ``/proc/self``, ``/proc/*/exe`` and ``/proc/*/maps``.

### DIFF
--- a/posix/subsystem/src/procfs.cpp
+++ b/posix/subsystem/src/procfs.cpp
@@ -237,8 +237,24 @@ std::shared_ptr<Link> DirectoryNode::directMkdir(std::string name) {
 	return link;
 }
 
+std::shared_ptr<Link> DirectoryNode::createProcDirectory(std::string name,
+		Process *process) {
+	auto link = directMkdir(name);
+	auto proc_dir = link->getTarget();
+
+	auto exe_link = std::make_shared<Link>(proc_dir, "exe", std::make_shared<ExeLink>(process));
+	static_cast<DirectoryNode*>(proc_dir.get())->_entries.insert(exe_link);
+
+	return link;
+}
+
 VfsType DirectoryNode::getType() {
 	return VfsType::directory;
+}
+
+async::result<frg::expected<Error, std::shared_ptr<FsLink>>> DirectoryNode::link(std::string name,
+		std::shared_ptr<FsNode> target) {
+	co_return Error::noSuchFile;
 }
 
 async::result<frg::expected<Error, FileStats>> DirectoryNode::getStats() {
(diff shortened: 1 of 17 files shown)


posix/subsystem/src/procfs.hpp
```diff
@@ -123,12 +136,21 @@ struct DirectoryNode final : FsNode, std::enable_shared_from_this<DirectoryNode>
...
 	async::result<frg::expected<Error, std::shared_ptr<FsLink>>> getLink(std::string name) override;
+	async::result<frg::expected<Error>> unlink(std::string name) override;
 
 private:
 	Link *_treeLink;
 	std::set<std::shared_ptr<Link>, LinkCompare> _entries;
 };
 
+struct SelfLink final : FsNode, std::enable_shared_from_this<SelfLink> {
+	SelfLink();
```
qookei, line 153: Using `= default;` here is most likely preferable to having an empty constructor in the source file.

posix/subsystem/src/procfs.cpp
```diff
@@ -293,6 +309,22 @@ async::result<frg::expected<Error, FileStats>> SelfLink::getStats() {
 	co_return FileStats{};
 }
 
+ExeLink::ExeLink(Process *process) : _process(process)
+{ }
```
qookei, line 322: This could probably be defined in the header, inside of the class body.

posix/subsystem/src/procfs.cpp
```diff
@@ -325,6 +334,53 @@ async::result<frg::expected<Error, FileStats>> ExeLink::getStats() {
...
+			auto fileStats = co_await fsNode->getStats();
+			DeviceId deviceId{};
+			if (fsNode->getType() == VfsType::charDevice || fsNode->getType() == VfsType::blockDevice)
+				deviceId = fsNode->readDevice();
+			assert(fileStats);
+
+			stream << std::dec << std::setfill('0') << std::setw(2) << deviceId.first << ":" << deviceId.second;
+			stream << " ";
+			stream << std::setw(0) << fileStats.value().inodeNumber;
+			stream << "\t";
```
qookei, line 371: On linux this seems to be padded with spaces. Does this matter?

posix/subsystem/src/procfs.cpp
```diff
@@ -325,6 +334,53 @@ async::result<frg::expected<Error, FileStats>> ExeLink::getStats() {
 	co_return FileStats{};
 }
 
+MapNode::MapNode(Process *process)
+: _process(process)
+{ }
```
qookei, line 339: This constructor could also be defined in the header in the class body.

qookei (commented):
Overall looks good, just some nitpicks.

posix/subsystem/src/procfs.cpp
Geertiebear: I'd hope it doesn't but there is probably some crazy program out there that matches spaces instead of logical whitespace. I'll pad it with 4 spaces ig.

qookei (approved):
LGTM now.

PR #272, opened 2021-04-15
docs: Fix some broken links and typo's and update building instructions

Dennisbonke:
Some small issues slipped through QA on the last docs PR, most notably the broken links on the commit messages page. In addition to that, the name of the `gui` metapackage has changed to `weston-desktop`, so I've adjusted the docs to that as well.

PR #406, opened 2021-12-14
Implement atomic modesetting in DRM

no92:
This PR aims to land atomic modesetting support in DRM and wiring up `vmware` and `plainfb` drivers for it. Atomic modesetting has been tested with `weston` and the [`drm-howto`](https://github.com/dvdhrm/docs/). `kmscon` does not utilize atomic modesetting.

These changes depend on managarm/mlibc#272.

### DIFF
--- a/core/drm/include/core/drm/core.hpp
+++ b/core/drm/include/core/drm/core.hpp
@@ -380,27 +380,18 @@ struct CrtcState {
 
 	std::weak_ptr<Crtc> crtc(void);
 
-	std::shared_ptr<Blob> mode();
-	void mode(std::shared_ptr<Blob> mode);
-
-	bool active();
-	void active(bool active);
-
-	bool mode_changed();
-
-private:
 	std::weak_ptr<Crtc> _crtc;
-	bool _active;
+	bool active;
 
-	bool _planesChanged;
-	bool _modeChanged;
-	bool _activeChanged;
-	bool _connectorsChanged;
-	uint32_t _planeMask;
-	uint32_t _connectorMask;
-	uint32_t _encoderMask;
+	bool planesChanged;
+	bool modeChanged;
+	bool activeChanged;
+	bool connectorsChanged;
+	uint32_t planeMask;
+	uint32_t connectorMask;
+	uint32_t encoderMask;
 
-	std::shared_ptr<Blob> _mode;
+	std::shared_ptr<Blob> mode;
 };
 
 struct Crtc : ModeObject {
@@ -452,20 +443,12 @@ private:
 };
 
 struct ConnectorState {
-	ConnectorState(std::weak_ptr<Connector> connector) : _connector(connector) {};
+	ConnectorState(std::weak_ptr<Connector> connector) : connector(connector) {};
 
-	std::shared_ptr<Connector> connector();
-	std::shared_ptr<Crtc> crtc();
-	void crtc(std::shared_ptr<Crtc> crtc);
-	std::shared_ptr<Encoder> encoder();
-	uint32_t dpms();
-	void dpms(uint32_t val);
-
-private:
-	std::shared_ptr<Connector> _connector;
-	std::shared_ptr<Crtc> _crtc;
-	std::shared_ptr<Encoder> _encoder;
-	uint32_t _dpms;
+	std::shared_ptr<Connector> connector;
+	std::shared_ptr<Crtc> crtc;
+	std::shared_ptr<Encoder> encoder;
+	uint32_t dpms;
 };
 
 /**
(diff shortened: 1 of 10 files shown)


no92:
Added DRM states for better atomic modesetting and avoiding data duplication, should be ready for review now.

core/drm/include/core/drm/core.hpp
```diff
@@ -225,6 +312,9 @@ struct File {
 	async::recurring_event _eventBell;
 
 	protocols::fs::StatusPageProvider _statusPage;
+
+	bool universal_planes;
```
avdgrinten, line 324: camelCase.

core/drm/include/core/drm/core.hpp
```diff
@@ -52,28 +52,70 @@ struct BlobPropertyType {
...
 	virtual ~Property() = default;
 
 	virtual bool validate(const Assignment& assignment);
+
+	PropertyId id();
+	uint32_t flags();
 	PropertyType propertyType();
+	std::string name();
+	void add_enum_info(uint64_t value, std::string name);
+	const std::unordered_map<uint64_t, std::string>& enum_info();
```
avdgrinten, line 113: What do these methods do?

avdgrinten (commented):
Quick review of the first two commits only. I still have to look at the new infrastructure in detail.

core/drm/include/core/drm/core.hpp
```diff
@@ -262,10 +358,16 @@ struct ModeObject {
 	void setupWeakPtr(std::weak_ptr<ModeObject> self);
 	std::shared_ptr<ModeObject> sharedModeObject();
 
+	bool setAssignment(const Assignment a);
```
avdgrinten: How does `setAssignment` interact with `Configuration`? As far as I can see it does not, at least not in the first commit?

core/drm/include/core/drm/core.hpp
no92: Querying property values returns an integer value for enums, where associating that numeric value with a enums member requires retrieving the mapping of numeric value to enum. These methods are for adding and retrieving those mappings.

core/drm/include/core/drm/core.hpp
```diff
@@ -52,28 +52,70 @@ struct BlobPropertyType {
...
+	CrtcY,
+	PlaneType,
+	Dpms,
+	CrtcId,
+	Active, /* 10 */
+	SrcX,
+	SrcY,
+	CrtcW,
+	CrtcH,
+};
```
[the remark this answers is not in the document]
no92: This served debugging purposes when logging DRM requests to decode property IDs and is probably unnecessary now, as property names can be retrieved by calling `name()` on the property.

core/drm/include/core/drm/core.hpp
no92: `setAssignment` is a relic of the old way of setting properties and gets removed in later commits.

no92:
Addressed the issues raised above.

core/drm/src/core.cpp
```diff
@@ -1003,6 +1550,249 @@ drm_core::File::ioctl(void *object, managarm::fs::CntRequest req,
...
+		auto &&transmit = helix::submitAsync(conversation, helix::Dispatcher::global(),
+			helix::action(&send_resp, ser.data(), ser.size()));
+		co_await transmit.async_wait();
+		HEL_CHECK(send_resp.error());
+	}else if(req.command() == DRM_IOCTL_SET_CLIENT_CAP) {
+		helix::SendBuffer send_resp;
+		managarm::fs::SvrResponse resp;
+
+		if(req.drm_capability() == DRM_CLIENT_CAP_STEREO_3D) {
+			std::cout << "\e[31mcore/drm: DRM client cap for stereo 3D unsupported\e[39m" << std::endl;
```
qookei, line 1563: Missing setting the error code here?

core/drm/src/core.cpp
```diff
@@ -1003,6 +1550,249 @@ drm_core::File::ioctl(void *object, managarm::fs::CntRequest req,
...
+		co_await transmit.async_wait();
+		HEL_CHECK(send_resp.error());
+	}else if(req.command() == DRM_IOCTL_MODE_SETPROPERTY) {
+		helix::SendBuffer send_resp;
+		managarm::fs::SvrResponse resp;
+
+		std::cout << "\e[31mcore/drm: DRM_IOCTL_MODE_SETPROPERTY is a no-op\e[39m"
+				<< std::endl;
+
+		auto ser = resp.SerializeAsString();
```
qookei, line 1640: No error code set here either?

core/drm/src/core.cpp
```diff
@@ -208,19 +367,149 @@ drm_core::Property *drm_core::Device::crtcXProperty() {
...
+
+const std::unordered_map<uint32_t, std::shared_ptr<drm_core::Property>>& drm_core::Device::getProperties() {
+	return _properties;
+}
+
 // ----------------------------------------------------------------
 // Property
 // ----------------------------------------------------------------
 
+std::vector<drm_core::Assignment> drm_core::ModeObject::getAssignments(std::shared_ptr<Device> dev) {	switch(this->type()) {
```
qookei, line 425: Missing newline?

core/drm/src/core.cpp
```diff
@@ -1003,6 +1550,249 @@ drm_core::File::ioctl(void *object, managarm::fs::CntRequest req,
...
+				}
+			}
+
+			prop_count += req.drm_prop_counts(i);
+		}
+
+		{
+			auto valid = config->capture(assignments, state);
+			assert(valid);
+		}
```
qookei, line 1781: Why is this in it's own block? Isn't `valid` just a bool?

qookei (commented):
Overall seems good, apart from the comments. Also one general thing is that some member variables and methods use snake case and some don't.

core/drm/src/core.cpp
no92: This is its own block due to the `goto` in line 1744.

no92:
Addressed the issues raised above, removed one unused function and adjusted code style to match the guidelines. However, the accessors for DRM properties are spelled the way as in the original, that is including underscores, but in lower case.

no92:
Removed redundant getters and setters, as discussed on Discord.

avdgrinten:
LGTM now, thanks a lot! Atomic modeset enables quite a lot of useful stuff.

PR #409, opened 2022-01-21
PCI MSI support and minor fixes

qookei:
Fixes race condition in MSI vector allocation and the possibility of entering an IRQ handler between printing the panic message and the stack trace.

### DIFF
--- a/kernel/thor/arch/arm/gic.cpp
+++ b/kernel/thor/arch/arm/gic.cpp
@@ -117,6 +117,7 @@ frg::string<KernelAlloc> GicDistributor::buildPinName(uint32_t irq) {
 }
 
 extern frg::manual_box<IrqSlot> globalIrqSlots[numIrqSlots];
+extern IrqSpinlock globalIrqSlotsLock;
 
 auto GicDistributor::setupIrq(uint32_t irq, TriggerMode trigger) -> Pin * {
 	if (irq >= irqPins_.size())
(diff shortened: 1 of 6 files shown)


kernel/thor/arch/arm/gic.cpp
```diff
@@ -129,6 +130,8 @@ auto GicDistributor::setupIrq(uint32_t irq, TriggerMode trigger) -> Pin * {
 }
 
 IrqStrategy GicDistributor::Pin::program(TriggerMode mode, Polarity polarity) {
+	auto guard = frg::guarf(&globalIrqSlotsLock);
```
avdgrinten, line 133: Typo (does not break GitHub CI because it's aarch64 only).

kernel/thor/system/pci/pci_discover.cpp
```diff
@@ -1110,6 +1175,17 @@ void checkPciFunction(PciBus *bus, uint32_t slot, uint32_t function,
 				space.store(msixVectorControl,
 						space.load(msixVectorControl) | 1);
 			}
+		} else if (device->msiIndex >= 0) {
+			auto offset = device->caps[device->msiIndex].offset;
+
+			auto msgControl = io->readConfigHalf(bus, slot, function, offset + 2);
+			device->numMsis = 1; // TODO(qookie): 1 << ((msgControl >> 1) & 0b111)
```
avdgrinten, line 1182: Hm? What about this TODO?

qookei: Related to the other one about supporting non-zero indices (MSI cap always guarantees 1 message is available).

PR #412, opened 2022-01-26
meta: prepare for Werror and Wimplicit-fallthrough

no92:
These changes enable compilation with `-Wimplicit-fallthrough` and, while not currently enabling it, work with `-Werror` if desired.

### DIFF
--- a/kernel/thor/generic/address-space.cpp
+++ b/kernel/thor/generic/address-space.cpp
@@ -15,6 +15,7 @@ namespace {
 	constexpr bool logCleanup = false;
 	constexpr bool logUsage = false;
 
+	[[maybe_unused]]
 	void logRss(VirtualSpace *space) {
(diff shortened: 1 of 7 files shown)


kernel/thor/generic/address-space.cpp
```diff
@@ -15,6 +15,7 @@ namespace {
 	constexpr bool logCleanup = false;
 	constexpr bool logUsage = false;
 
+	__attribute__((unused))
```
qookei, line 18: You could use a C++-style attribute here (`[[maybe_unused]]` is standard and would be the best fit).

no92:
Addressed the comment above. CI fails because managarm/frigg#18 is not merged yet.

avdgrinten:
LGTM, thanks for the PR!

PR #413, opened 2022-01-30
posix/init: Launch native X by default (DO NOT MERGE)

Dennisbonke:
This PR changes the weston option to launch Xorg, which should launch a default config bspwm. Do not merge, this is for testing only.


ArsenArsen:
shooting this one down chief, definitely shouldn't change this default, add a cmdline for it

Dennisbonke:
@ArsenArsen it was at the request of leo and never to be merged, obviously a cmdline switch would be made for this

PR #415, opened 2022-01-31
Bugfixes to DRM

no92:


no92:
ksmcon, sway and weston 10 work with these changes. This PR will stay a draft for now, until we're sure that xorg-server works with DRM.

avdgrinten:
Doesn't it make more sense to add doxygen comments to the headers (and not the sources)?

avdgrinten (approved):
LGTM.

PR #417, opened 2022-02-04
netserver/tcp: Add getsockname and getpeername

Dennisbonke:
`OpenSSH` makes use of `sockname()` and `peername()` on tcp sockets. Implement this functionality so `OpenSSH` at least somewhat knows who it is talking to.

### DIFF
--- a/servers/netserver/src/ip/tcp4.cpp
+++ b/servers/netserver/src/ip/tcp4.cpp
@@ -410,6 +410,41 @@ struct Tcp4Socket {
 		co_return 0;
 	}
 
+	static async::result<size_t> sockname(void *obj, void *addrPtr, size_t maxAddrLength) {
+		sockaddr_in sa;
+		in_addr addr;
+		auto self = static_cast<Tcp4Socket *>(obj);
+
+		memset(&sa, 0, sizeof(struct sockaddr_in));
+		memset(&addr, 0, sizeof(struct in_addr));
+		sa.sin_family = AF_INET;
+		addr.s_addr = htonl(self->localEp_.ipAddress);
+		sa.sin_addr = addr;
+		sa.sin_port = self->localEp_.port;
+
+		auto destSize = std::min(sizeof(sockaddr_in), maxAddrLength);
+		memcpy(addrPtr, &sa, destSize);
+		co_return destSize;
+	}
+
+	static async::result<frg::expected<protocols::fs::Error, size_t>> 
+	peername(void *obj, void *addrPtr, size_t maxAddrLength) {
+		sockaddr_in sa;
+		in_addr addr;
+		auto self = static_cast<Tcp4Socket *>(obj);
+
+		memset(&sa, 0, sizeof(struct sockaddr_in));
+		memset(&addr, 0, sizeof(struct in_addr));
+		sa.sin_family = AF_INET;
+		addr.s_addr = htonl(self->remoteEp_.ipAddress);
+		sa.sin_addr = addr;
+		sa.sin_port = self->remoteEp_.port;
+
+		auto destSize = std::min(sizeof(sockaddr_in), maxAddrLength);
+		memcpy(addrPtr, &sa, destSize);
+		co_return destSize;
+	}
+
 	constexpr static protocols::fs::FileOperations ops {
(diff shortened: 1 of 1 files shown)


servers/netserver/src/ip/tcp4.cpp
```diff
@@ -410,17 +410,54 @@ struct Tcp4Socket {
...
+		memset(&sa, 0, sizeof(struct sockaddr_in));
+		memset(&addr, 0, sizeof(struct in_addr));
+		sa.sin_family = AF_INET;
+		addr.s_addr = htonl(self->remoteEp_.ipAddress);
+		sa.sin_addr = addr;
+		sa.sin_port = self->remoteEp_.port;
+
+		auto destSize = std::min(sizeof(sockaddr_in), maxAddrLength);
+		memcpy(addrPtr, &sa, destSize);
+		co_return 128;
```
ArsenArsen, line 445: Incorrect return value?

servers/netserver/src/ip/tcp4.cpp
```diff
@@ -410,17 +410,54 @@ struct Tcp4Socket {
...
+		memset(&sa, 0, sizeof(struct sockaddr_in));
+		memset(&addr, 0, sizeof(struct in_addr));
+		sa.sin_family = AF_INET;
+		addr.s_addr = htonl(self->localEp_.ipAddress);
+		sa.sin_addr = addr;
+		sa.sin_port = self->localEp_.port;
+
+		auto destSize = std::min(sizeof(sockaddr_in), maxAddrLength);
+		memcpy(addrPtr, &sa, destSize);
+		co_return 128;
```
ArsenArsen, line 427: Incorrect return value?

These functions should return zero even if the address is trimmed. I could be forgetting how we utilize these return values, though,

ArsenArsen (changes requested):

Dennisbonke:
I believe I've addressed the comments made, ready for another round.

no92 (approved):

servers/netserver/src/ip/tcp4.cpp
```diff
@@ -410,17 +410,54 @@ struct Tcp4Socket {
...
+		memset(&sa, 0, sizeof(struct sockaddr_in));
+		memset(&addr, 0, sizeof(struct in_addr));
+		sa.sin_family = AF_INET;
+		addr.s_addr = htonl(self->localEp_.ipAddress);
+		sa.sin_addr = addr;
+		sa.sin_port = self->localEp_.port;
+
+		auto destSize = std::min(sizeof(sockaddr_in), maxAddrLength);
+		memcpy(addrPtr, &sa, destSize);
+		co_return destSize;
```
ArsenArsen, line 427:
```
DESCRIPTION
       getsockname() returns the current address to which the socket sockfd is bound, in the buffer pointed to by
       addr.  The addrlen argument should be initialized to indicate the amount of space (in bytes) pointed to by
       addr.  On return it contains the actual size of the socket address.

       The returned address is truncated if the buffer provided is too small; in this case, addrlen will return a
       value greater than was supplied to the call.

RETURN VALUE
       On success, zero is returned.  On error, -1 is returned, and errno is set to indicate the error.
```

ArsenArsen (changes requested):

servers/netserver/src/ip/tcp4.cpp
```diff
@@ -410,17 +410,54 @@ struct Tcp4Socket {
...
+		memset(&sa, 0, sizeof(struct sockaddr_in));
+		memset(&addr, 0, sizeof(struct in_addr));
+		sa.sin_family = AF_INET;
+		addr.s_addr = htonl(self->remoteEp_.ipAddress);
+		sa.sin_addr = addr;
+		sa.sin_port = self->remoteEp_.port;
+
+		auto destSize = std::min(sizeof(sockaddr_in), maxAddrLength);
+		memcpy(addrPtr, &sa, destSize);
+		co_return destSize;
```
ArsenArsen, line 445: ditto

ArsenArsen (changes requested):

Dennisbonke:
Superseeded by #595

PR #418, opened 2022-02-04
posix: implement memfd and file sealing

no92:
The `MemoryFile` used here is an adapted version of the same struct in tmpfs, stripped of some functionality not yet needed.

### DIFF
--- a/posix/init/src/stage2.cpp
+++ b/posix/init/src/stage2.cpp
@@ -158,6 +158,7 @@ int main() {
 		setenv("XDG_RUNTIME_DIR", "/run", 1);
 		setenv("MESA_GLSL_CACHE_DISABLE", "1", 1);
 //		setenv("MESA_DEBUG", "1", 1);
+		setenv("SHELL", "/bin/bash", 1);
 
 		if(launch == "kmscon") {
 			// TODO: kmscon should invoke a login program which sets these environment vars.
--- a/posix/subsystem/src/file.hpp
+++ b/posix/subsystem/src/file.hpp
@@ -121,7 +121,7 @@ public:
 	static async::result<frg::expected<protocols::fs::Error>>
 	ptTruncate(void *object, size_t size);
 
-	static async::result<void>
+	static async::result<frg::expected<protocols::fs::Error>>
 	ptAllocate(void *object, int64_t offset, size_t size);
 
 	static async::result<int>
@@ -169,6 +169,8 @@ public:
 	static async::result<frg::expected<protocols::fs::Error, size_t>>
 	ptPeername(void *object, void *addr_ptr, size_t max_addr_length);
 
+	static async::result<frg::expected<protocols::fs::Error, int>> ptGetSeals(void *object);
+	static async::result<frg::expected<protocols::fs::Error, int>> ptAddSeals(void *object, int seals);
 
 	static constexpr auto fileOperations = protocols::fs::FileOperations{
 		.seekAbs = &ptSeekAbs,
@@ -191,6 +193,8 @@ public:
 		.recvMsg = &ptRecvMsg,
 		.sendMsg = &ptSendMsg,
 		.peername = &ptPeername,
+		.getSeals = &ptGetSeals,
+		.addSeals = &ptAddSeals,
 	};
 
 	// ------------------------------------------------------------------------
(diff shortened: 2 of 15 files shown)


no92:
CI is failing right now because it depends on a definition of `F_SEAL_SEAL`, which is included in the mlibc PR.

posix/subsystem/src/file.hpp
```diff
@@ -340,6 +346,9 @@ struct File : private smarter::crtp_counter<File, DisposeFileHandle> {
 	DefaultOps _defaultOps;
 
 	bool _isOpen;
+
+protected:
+	int _seals;
```
avdgrinten, line 351: Seals should be handled by the memfd `File` implementation and not by the general `File` class. If we make them generic, the chance is very high that we miss a few places where we would need to check for seals.

posix/subsystem/src/memfd.hpp
```diff
@@ -0,0 +1,53 @@
+#pragma once
+
+#include <fcntl.h>
+
+#include "file.hpp"
+
+struct MemoryFile final : File {
+public:
+	static void serve(smarter::shared_ptr<MemoryFile> file) {
+// TODO:		assert(!file->_passthrough);
```
avdgrinten, line 10: Hm?

no92:
Addressed the issues raised above.

no92:
Dropped some noops that are better served by `File`.

no92:
Blocked on mesa not compiling with memfd present, due to it making direct Linux syscalls.

Dennisbonke:
> Blocked on mesa not compiling with memfd present, due to it making direct Linux syscalls.

I think I might be able to patch that out but with weston 10 crashing consistently I'm unable to properly test that. Feel free to ping me on discord at a later date so I can look at it.

posix/subsystem/src/main.cpp
```diff
@@ -3679,6 +3680,47 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+
+			if (!req) {
+				std::cout << "posix: Rejecting request due to decoding failure" << std::endl;
+				break;
+			}
+
+			if(logRequests)
+				std::cout << "posix: MEMFD_CREATE " << req->name() << std::endl;
+
+			std::cout << "posix: MEMFD_CREATE ignores some flags" << std::endl;
```
avdgrinten, line 3701: Shouldn't we return EINVAL for unsupported flags?

posix/subsystem/src/process.cpp
```diff
@@ -1124,7 +1124,9 @@ async::result<void> Process::terminate(TerminationState state) {
 	_fileContext = nullptr;
 	//_signalContext = nullptr; // TODO: Migrate the notifications to PID 1.
 	_currentGeneration = nullptr;
-	assert(co_await _procfs_dir->getOwner()->unlink(_procfs_dir->getName()));
+	if(_procfs_dir) {
+		assert(co_await _procfs_dir->getOwner()->unlink(_procfs_dir->getName()));
+	}
```
avdgrinten, line 1129: The code needs to be moved out of `assert()`. We can still `assert` that it succeeds, but this snippet does not run in release mode. Replace by something like:
```cpp
auto unlinkResult = co_await ...;
assert(unlinkResult);
(void)unlinkResult;
```

avdgrinten (commented):
See below. Also, this PR contains commits that are not related to memfd and file sealing; this should be avoided.

posix/init/src/stage2.cpp
```diff
@@ -172,6 +172,7 @@ int main() {
 			if(mkdir("/tmp/.X11-unix", 1777))
 				throw std::runtime_error("mkdir() failed");
 			//execl("/usr/bin/weston", "weston", nullptr);
+			setenv("SHELL", "/bin/bash", 1);
```
Dennisbonke, line 175: Why not put this `setenv` call a bit further up with the other calls? All environments could benefit from having `SHELL` set.

no92:
Addressed the issues raised above.

posix/subsystem/src/main.cpp
Dennisbonke: According to the [man page](https://man7.org/linux/man-pages/man2/memfd_create.2.html) we should yes.

posix/subsystem/src/memfd.hpp
```diff
@@ -0,0 +1,52 @@
...
+	helix::UniqueLane _passthrough;
+	async::cancellation_event _cancelServe;
+
+	uint64_t _offset;
+
+	helix::UniqueDescriptor _memory;
+	helix::Mapping _mapping;
+	size_t _areaSize;
+	size_t _fileSize;
+	int _seals;
```
avdgrinten, line 51: Make sure all members are initialized. `_areaSize`, `_fileSize` and `_seals` do not seem to be initialized in the ctor. You can simply add `= 0` here.

avdgrinten (commented):
One small comment, can be merged afterwards.

Issue #367, opened 2021-07-13
posix: Hang in WAIT request


Dennisbonke:
While executing `man man` in kmscon or weston, the program eventually hangs in the WAIT request, thus freezing the process, or in kmscon the entire system, as there is no way to kill `man`. From what I can see, both kernel and posix-subsystem continue to function.

EDIT: This might be the `less` pager that is hanging, investigating.

Dennisbonke:
And turns out the pager is indeed at fault, disregard this issue.

PR #419, opened 2022-02-04
posix: implement PRIME for fd <-> handle conversion

no92:
This is rough code for making the PRIME ioctls work for sway. The code is not optimal and the naming is weird, but I couldn't come up with something better.

Posting as draft, as I don't expect this to survive review unchanged.

mlibc counterpart is managarm/mlibc#367.

### DIFF
--- a/core/drm/include/core/drm/core.hpp
+++ b/core/drm/include/core/drm/core.hpp
@@ -5,6 +5,7 @@
 #include <map>
 #include <unordered_map>
 #include <optional>
+#include <variant>
 
 #include <arch/mem_space.hpp>
 #include <async/cancellation.hpp>
@@ -210,6 +211,7 @@ public:
 	uint32_t getMinHeight();
 	uint32_t getMaxHeight();
 
+	helix::UniqueDescriptor _posixLane;
 private:
 	std::vector<Crtc *> _crtcs;
 	std::vector<Encoder *> _encoders;
@@ -248,7 +250,12 @@ private:
 	std::shared_ptr<Property> _crtcWProperty;
 	std::shared_ptr<Property> _crtcHProperty;
 
+	std::map<std::array<char, 16>, std::shared_ptr<drm_core::BufferObject>> _exportedBufferObjects;
+
 public:
+	void registerBufferObject(std::shared_ptr<drm_core::BufferObject> obj, std::array<char, 16> creds);
+	std::shared_ptr<drm_core::BufferObject> findBufferObject(std::array<char, 16> creds);
+
 	id_allocator<uint32_t> allocator;
 
 	/**
@@ -331,6 +338,12 @@ struct File {
 	 */
 	uint32_t createHandle(std::shared_ptr<BufferObject> bo);
 	BufferObject *resolveHandle(uint32_t handle);
+	uint32_t getHandle(std::shared_ptr<drm_core::BufferObject> bo);
+
+	bool exportBufferObject(uint32_t handle, std::array<char, 16> creds);
+
+	std::pair<std::shared_ptr<drm_core::BufferObject>, uint32_t>
+	importBufferObject(std::array<char, 16> creds);
 
 	/**
 	 * Add an @p event to the queue of pending events to be read by userspace.
@@ -368,6 +381,21 @@ private:
 
 	bool universalPlanes;
 	bool atomic;
+};
+
+struct PrimeFile {
+	PrimeFile(helix::BorrowedDescriptor handle, size_t size);
+
+	static async::result<helix::BorrowedDescriptor> accessMemory(void *object);
+
+	static async::result<protocols::fs::SeekResult> seekAbs(void *object, int64_t offset);
+	static async::result<protocols::fs::SeekResult> seekRel(void *object, int64_t offset);
+	static async::result<protocols::fs::SeekResult> seekEof(void *object, int64_t offset);
+
+	helix::BorrowedDescriptor _memory;
+
+	size_t offset;
+	size_t size;
 };
 
 struct Configuration {
--- a/core/drm/src/core.cpp
+++ b/core/drm/src/core.cpp
@@ -19,6 +19,7 @@
 #include <libdrm/drm_fourcc.h>
 
 #include "fs.bragi.hpp"
+#include "posix.bragi.hpp"
 #include "core/drm/core.hpp"
 
 constexpr bool logDrmRequests = false;
@@ -26,6 +27,17 @@ constexpr bool logDrmRequests = false;
 // ----------------------------------------------------------------
 // Device
 // ----------------------------------------------------------------
+
+namespace drm_core {
+
+static constexpr auto primeFileOperations = protocols::fs::FileOperations{
+	.seekAbs = &drm_core::PrimeFile::seekAbs,
+	.seekRel = &drm_core::PrimeFile::seekRel,
+	.seekEof = &drm_core::PrimeFile::seekEof,
+	.accessMemory = &drm_core::PrimeFile::accessMemory,
+};
+
+}
 
 drm_core::Device::Device() {
 	struct SrcWProperty : drm_core::Property {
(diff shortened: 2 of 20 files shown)


core/drm/include/core/drm/core.hpp
```diff
@@ -210,6 +211,8 @@ struct Device {
 	uint32_t getMinHeight();
 	uint32_t getMaxHeight();
 
+	helix::UniqueDescriptor _posixLane;
+	std::map<uint64_t, uint32_t> _fd_mappings;
```
64, line 215: This should be camel case.

no92: This should be removed as it is unused.

kernel/thor/generic/thor-internal/thread.hpp
```diff
@@ -37,7 +37,18 @@ smarter::borrowed_ptr<Thread> getCurrentThread();
 
 struct ActiveHandle { };
 
-struct Thread final : smarter::crtp_counter<Thread, ActiveHandle>, ScheduleEntity {
+struct Credentials {
+	Credentials(void);
+	void allocateCredentials(void);
```
avdgrinten, line 42: `void` as argument doesn't do anything in C++. Also: why is `allocateCredentials` its own function? Why should one call it over the constructor?

core/drm/src/core.cpp
```diff
@@ -930,6 +963,30 @@ drm_core::BufferObject *drm_core::File::resolveHandle(uint32_t handle) {
 	return it->second.get();
 };
 
+void drm_core::File::exportBufferObject(uint32_t handle, std::array<char, 16> creds) {
+	auto bo = resolveHandle(handle);
+	assert(bo);
```
avdgrinten, line 968: What if the handle is illegal?

core/drm/src/core.cpp
```diff
@@ -930,6 +963,30 @@ drm_core::BufferObject *drm_core::File::resolveHandle(uint32_t handle) {
...
+	auto buffer = bo->sharedBufferObject();
+
+	_device->exportBufferObject(buffer, creds);
+	_primeBufferMap.insert({creds, handle});
+}
+
+std::pair<std::shared_ptr<drm_core::BufferObject>, uint32_t>
+drm_core::File::importBufferObject(std::array<char, 16> creds) {
+	auto bo = _device->importBufferObject(creds);
+	assert(bo);
```
avdgrinten, line 978: What if the credentials are illegal?

core/drm/src/core.cpp
```diff
@@ -930,6 +963,30 @@ drm_core::BufferObject *drm_core::File::resolveHandle(uint32_t handle) {
...
+
+	auto it = _primeBufferMap.find(creds);
+	if(it != _primeBufferMap.end()) {
+		return {bo, it->second};
+	} else {
+		uint32_t handle = createHandle(bo);
+		_primeBufferMap.insert({creds, handle});
+		return {bo, handle};
+	}
```
avdgrinten, line 987: Why do we insert something in the `import` function? Isn't exportBufferObject supposed to do that?

core/drm/include/core/drm/core.hpp
```diff
@@ -248,7 +250,12 @@ struct Device {
 	std::shared_ptr<Property> _crtcWProperty;
 	std::shared_ptr<Property> _crtcHProperty;
 
+	std::map<std::array<char, 16>, std::shared_ptr<drm_core::BufferObject>> _exportedBufferObjects;
+
 public:
+	void exportBufferObject(std::shared_ptr<drm_core::BufferObject> obj, std::array<char, 16> creds);
+	std::shared_ptr<drm_core::BufferObject> importBufferObject(std::array<char, 16> creds);
```
avdgrinten, line 257: I don't understand what the difference between these functions in `File` and `Device` is. It doesn't help that they have the same name.

core/drm/src/core.cpp
```diff
@@ -202,6 +216,10 @@ drm_core::Device::Device() {
 		bool validate(const Assignment&) override {
 			return true;
 		};
+
+		void writeToState(const Assignment assignment, std::unique_ptr<AtomicState> &state) override {
+			state->crtc(assignment.object->id())->active = assignment.intValue;
+		}
```
avdgrinten, line 222: Unrelated change? Seems that this function is never called.

core/drm/src/core.cpp
```diff
@@ -1733,6 +1816,88 @@ drm_core::File::ioctl(void *object, managarm::fs::CntRequest req,
...
+		auto [creds] = co_await helix_ng::exchangeMsgs(conversation, helix_ng::extractCredentials());
+		HEL_CHECK(creds.error());
+
+		// 'import' the BufferObject while returning or creating the DRM handle that references it
+		std::array<char, 16> credentials;
+		std::copy_n(creds.credentials(), 16, std::begin(credentials));
+		auto [bo, handle] = self->importBufferObject(credentials);
+
+		assert(handle);
+		assert(bo);
```
avdgrinten, line 1890: This is missing error handling.

avdgrinten (commented):
Architecturally, the PR is good now. Some minor issues still need to be fixed.

core/drm/src/core.cpp
no92: While unused now, I have some patches to drivers that make this necessary as some piece of software relied on this property.

core/drm/src/core.cpp
no92: If the handle is invalid/unknown, `resolveHandle` returns a nullptr. The assert is supposed to catch that.

core/drm/include/core/drm/core.hpp
no92: In `Device`, (credential, BO) pairs are registered so they can be imported back into `File`s. Each `File` has their own mappings for DRM handles, and that's what the functions in `File` set up. It might however be better to rename the functions in `Device` to something like `registerBufferObject` and `findBufferObject` to avoid confusion.

core/drm/src/core.cpp
```diff
@@ -913,7 +942,11 @@ const std::vector<std::shared_ptr<drm_core::FrameBuffer>> &drm_core::File::getFr
 
 uint32_t drm_core::File::createHandle(std::shared_ptr<BufferObject> bo) {
 	auto handle = _allocator.allocate();
-	_buffers.insert({handle, bo});
+	auto ret = _buffers.insert({handle, bo});
+	assert(ret.second);
```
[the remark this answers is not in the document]
no92: Hopefully not, but this is a sanity check to catch the (unlikely) case of duplicate handles.

core/drm/src/core.cpp
no92: If the credentials are illegal, the assert would trip due to `importBufferObject` returning a nullptr.

core/drm/src/core.cpp
no92: The `_primeBufferMap` is the `File`-local mapping between credentials and DRM handles, which can change between `File`s. This fact is the entire reason for PRIME using fds and not the pre-existing DRM handles. The `File::importBufferObject` function returns a pair of (BO, DRM handle) read from the `_primeBufferMap`, which is used in the PRIME ioctls.

core/drm/include/core/drm/core.hpp
```diff
@@ -357,6 +369,8 @@ struct File {
 	std::unordered_map<uint32_t, std::shared_ptr<BufferObject>> _buffers;
 	// id allocator for mapping BufferObjects
 	id_allocator<uint32_t> _allocator;
+	// a mapping of credentials to DRM handles for PRIME
+	std::map<std::array<char, 16>, uint32_t> _primeBufferMap;
```
avdgrinten, line 373: Why doesn't this simply map the credentials to a shared_ptr that points to the buffer? The redirection over handles seems unnecessary (also handles can be deallocated and reallocated).

no92: We do not currently free ids in `DESTROY_DUMB` as it turns out, so handle reassignment does not happen with current code.

no92:
Removed `_primeBufferMap`, squashed the fixups, dropped one commit as requested.

avdgrinten (approved):
LGTM, thank you for the PR!

PR #420, opened 2022-02-05
docs: Fix minor link issue

a-andreyev:


a-andreyev:
The related link is inside `docs/src/getting-started/supported/index.md`:

`A quick overview of various categories can be found at [package list](package-index.md).`

avdgrinten:
For clarification: what does this fix? A broken link?

If the file is renamed, we also need to fix the index here: https://github.com/managarm/managarm/blob/master/docs/src/SUMMARY.md?plain=1#L8

a-andreyev:
> For clarification: what does this fix? A broken link?

Yes. I found a broken link in `docs/src/getting-started/supported/index.md`. Firstly I've tried just to update the link in markdown according to the existing `package_index.md` file. Then I've discovered that all the other filenames are in kebab-case, so it looks like it's more convenient to update the filename as `package-index.md`.

> If the file is renamed, we also need to fix the index here: https://github.com/managarm/managarm/blob/master/docs/src/SUMMARY.md?plain=1#L8

Thank you! Also updated the summary file.

avdgrinten (approved):
LGTM now, thanks for the PR!

PR #421, opened 2022-02-10
Misc. fixes and SIGWINCH support.

### LINKED ISSUE #166: posix: Implement SIGWINCH
This allows programs like nano and vim to be resized.

qookei:
Fixes #166.

### DIFF
--- a/kernel/eir/generic/main.cpp
+++ b/kernel/eir/generic/main.cpp
@@ -104,37 +104,54 @@ address_t cutFromRegion(size_t size) {
 			continue;
 
 		regions[i].size -= size;
+
+		// Discard this region if it's smaller than alignment
+		if(regions[i].size < 0x200000)
+			regions[i].regionType = RegionType::null;
+
 		return regions[i].address + regions[i].size;
 	}
-	
+
 	eir::panicLogger() << "Eir: Unable to cut memory from a region" << frg::endlog;
 	__builtin_unreachable();
 }
 
 void setupRegionStructs() {
+	for(size_t j = numRegions; j > 0; j--) {
+		size_t i = j - 1;
+
+		if(regions[i].regionType != RegionType::allocatable)
+			continue;
+
+		// Setup a buddy allocator.
+		auto order = BuddyAccessor::suitableOrder(regions[i].size >> pageShift);
+		auto preRoots = regions[i].size >> (pageShift + order);
+		auto overhead = BuddyAccessor::determineSize(preRoots, order);
+		overhead = (overhead + address_t(pageSize - 1)) & ~address_t(pageSize - 1);
+
+		assert(overhead >= preRoots * (1 << (order + 1)));
+
+		regions[i].buddyTree = cutFromRegion(overhead);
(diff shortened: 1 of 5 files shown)


kernel/eir/generic/main.cpp
```diff
@@ -104,37 +104,61 @@ address_t cutFromRegion(size_t size) {
...
-	
+
 	eir::panicLogger() << "Eir: Unable to cut memory from a region" << frg::endlog;
 	__builtin_unreachable();
 }
 
 void setupRegionStructs() {
-	for(size_t i = 0; i < numRegions; ++i) {
+	address_t buddyTreeAddrs[numRegions] = {};
+	address_t buddyTreeOverheads[numRegions] = {};
```
avdgrinten, line 121: Can't we just store this info within the `regions` array?

posix/subsystem/src/main.cpp
```diff
@@ -1052,7 +1052,6 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 
 			void *address;
 			if(req->flags() & MAP_ANONYMOUS) {
-				assert(req->fd() == -1);
```
avdgrinten, line 1055: Why is that removed? The assertion is broken, but shouldn't we return an error?

posix/subsystem/src/process.cpp
```diff
@@ -466,6 +466,12 @@ struct CompileSignalInfo {
 	siginfo_t *si;
 };
 
+constexpr SignalDisposition defaultDispositions[65] = {
+	[SIGCHLD] = SignalDisposition::ignore,
+	[SIGURG] = SignalDisposition::ignore,
+	[SIGWINCH] = SignalDisposition::ignore
+};
```
avdgrinten, line 473: What about all other signals?

qookei: All other signals are zero initialized (`SignalDisposition::none`), and we treat `handler.disposition == none && defaultDispositions[sn] == none` as terminate.

posix/subsystem/src/main.cpp
qookei: As per the Linux man page, fd is ignored if MAP_ANONYMOUS is specified:
```
       MAP_ANONYMOUS
              The mapping is not backed by any file; its contents are initial‐
              ized to zero.  The fd argument is ignored; ...
```

posix/subsystem/src/process.cpp
ArsenArsen: Other than TSTP, STOP and CONT, whose use-case we don't support yet

avdgrinten: Hm, one issue with this implementation is that there are default behaviors that that differ from all possible user specified behavior (for example for SIGCHLD, SIGSTOP, etc.). Maybe this is better expressed by a switch in the handler function?

Dennisbonke:
What is the status of this PR?

avdgrinten (approved):
LGTM.

Issue #379, opened 2021-07-28
thor: Implement generational LRU


avdgrinten:
Thor currently uses an LRU algorithm to reclaim memory. However, due to hardware constraints and runtime overhead, we cannot determine the least recently used pages exactly (and we need to rely on accessed bits instead). Solve this in a way similar to Linux: store multiple generations (Linux uses two, namely the "active" and the "inactive" one) of pages in the memory reclaimer. Bumping a page moves it to the youngest generation, while each unsuccessful scan moves the page to the next older generation. Reclaim always frees the oldest generation first.

avdgrinten:
Fixed in https://github.com/managarm/managarm/pull/1307

PR #424, opened 2022-02-12
posix: implement O_NONBLOCK for netlink

64:
This PR fixes a hang in `kmscon` caused by the use of a `O_NONBLOCK` netlink socket in `eudev`. Since we don't support that yet, the call to `recvmsg` never returns.

We implement the required functionality and also modify init to block on the call to `udev_monitor_receive_device`, as would have happened before.

See also https://github.com/managarm/mlibc/pull/379 which fixes some further boot failures with this patch.

### DIFF
--- a/posix/init/src/stage2.cpp
+++ b/posix/init/src/stage2.cpp
@@ -93,7 +93,8 @@ int main() {
 	// Set the into blocking mode, so that we don't have to poll it.
 	auto flags = fcntl(udev_monitor_get_fd(init_udev_mon), F_GETFL, 0);
 	assert(flags >= 0);
-	assert(!fcntl(udev_monitor_get_fd(init_udev_mon), F_SETFL, flags & ~O_NONBLOCK));
+	auto ret = fcntl(udev_monitor_get_fd(init_udev_mon), F_SETFL, flags & ~O_NONBLOCK);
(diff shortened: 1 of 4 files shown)


posix/init/src/stage2.cpp
```diff
@@ -90,6 +90,11 @@ int main() {
 
 	waitpid(udev_settle, nullptr, 0);
 
+	// Set the into blocking mode, so that we don't have to poll it.
```
avdgrinten, line 93: Why is the socket non-blocking anyway?

posix/init/src/stage2.cpp
```diff
@@ -90,6 +90,11 @@ int main() {
...
+	auto flags = fcntl(udev_monitor_get_fd(init_udev_mon), F_GETFL, 0);
+	assert(flags >= 0);
+	assert(!fcntl(udev_monitor_get_fd(init_udev_mon), F_SETFL, flags & ~O_NONBLOCK));
```
avdgrinten, line 96: Logic in assertions is broken since assertions are removed entirely if NDEBUG is set.

avdgrinten (commented):
LGTM except for some comments below.

avdgrinten:
LGTM now.

PR #430, opened 2022-03-10
Implement pwrite

Dennisbonke:
While we've had `pread()` for a while now, `pwrite()` wasn't implemented. This PR adds `pwrite()` support for ext2 and tmpfs, which pleases `gtk3`, specifically `dconf`.

A corresponding mlibc PR will be provided shortly.

### DIFF
--- a/drivers/libblockfs/src/libblockfs.cpp
+++ b/drivers/libblockfs/src/libblockfs.cpp
@@ -167,6 +167,17 @@ async::result<frg::expected<protocols::fs::Error, size_t>> write(void *object, c
 	co_return length;
 }
 
+async::result<frg::expected<protocols::fs::Error, size_t>> pwrite(void *object, int64_t offset, const char *credentials,
+			const void *buffer, size_t length) {
+	if(!length) {
+		co_return 0;
+	}
+
+	auto self = static_cast<ext2fs::OpenFile *>(object);
+	co_await self->inode->fs.write(self->inode.get(), offset, buffer, length);
+	co_return length;
+}
+
 async::result<helix::BorrowedDescriptor>
 accessMemory(void *object) {
 	auto self = static_cast<ext2fs::OpenFile *>(object);
@@ -208,6 +219,7 @@ constexpr protocols::fs::FileOperations fileOperations {
 	.read         = &read,
 	.pread        = &pread,
 	.write        = &write,
+	.pwrite       = &pwrite,
 	.readEntries  = &readEntries,
 	.accessMemory = &accessMemory,
 	.truncate     = &truncate,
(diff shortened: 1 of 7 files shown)


drivers/libblockfs/src/libblockfs.cpp
```diff
@@ -167,6 +167,15 @@ async::result<frg::expected<protocols::fs::Error, size_t>> write(void *object, c
 	co_return length;
 }
 
+async::result<frg::expected<protocols::fs::Error, size_t>> pwrite(void *object, int64_t offset, const char *credentials,
+			const void *buffer, size_t length) {
+	assert(length);
```
avdgrinten, line 172: This assertion should be fixed.

protocols/fs/src/server.cpp
```diff
@@ -267,6 +267,61 @@ async::detached handlePassthrough(smarter::shared_ptr<void> file,
...
+
+		managarm::fs::SvrResponse resp;
+		if(!res) {
+			if(res.error() == Error::noSpaceLeft) {
+				resp.set_error(managarm::fs::Errors::NO_SPACE_LEFT);
+			} else if(res.error() == Error::wouldBlock) {
+				resp.set_error(managarm::fs::Errors::WOULD_BLOCK);
+			} else {
+				std::cout << "Unknown error from pwrite()" << std::endl;
+				co_return;
```
avdgrinten, line 313: Is this how we also handle other errors of this kind?

posix/subsystem/src/file.cpp
```diff
@@ -266,6 +283,12 @@ async::result<frg::expected<Error, size_t>> File::writeAll(Process *, const void
 	throw std::runtime_error("posix: Object has no File::writeAll()");
 }
 
+async::result<frg::expected<Error, size_t>> File::pwrite(Process *, int64_t, const void *, size_t) {
+	std::cout << "posix \e[1;34m" << structName()
+			<< "\e[0m: Object does not implement pwrite()" << std::endl;
+	throw std::runtime_error("posix: Object has no File::pwrite()");
+}
```
avdgrinten, line 290: Let's return an error instead of dying here. I guess the default error should be `ESPIPE`?

protocols/fs/src/server.cpp
Dennisbonke: Yes, this would catch instances where a new error is introduced and returned by libblockfs (or any other implementation), but not yet propagated through protocols. I believe `sendMsg()` has it too.

drivers/libblockfs/src/libblockfs.cpp
Dennisbonke: What would you like to see then? The `write()` function in libblockfs has the same assert, I assume for a reason. I couldn't find any type of error I'd like to return from the man pages.

posix/subsystem/src/file.cpp
Dennisbonke: Done as requested.

drivers/libblockfs/src/libblockfs.cpp
Dennisbonke: Done as requested, also for `write()`.

avdgrinten (approved):
LGTM.

avdgrinten:
Actually, this PR has conflicts.

Issue #359, opened 2021-06-23
usb-hid: Unexpected token: 0xb0


borrrden:
Related to #357 in that I am trying to reproduce it on qemu.  This was stalled for a while with a bug in qemu USB passthrough but now I've gotten it working as noted by the following:

> usb-hid: Detected USB device
usb-hid: Ignoring secondary HID interface: 1, alternative: 0
usb-hid: Ignoring secondary HID interface: 2, alternative: 0
usb-hid: Detected HID device. Interface: 0, alternative: 0
usb-hid: Using endpoint number 1
usb-hid: Parsing report descriptor

Unfortunately it seems there is a token in there it doesn't like and it ends up aborting:

> Unexpected token: 0xb0

Issue #357, opened 2021-06-16
ehci: Crash on `Assertion '_state == State::null' failed!`


borrrden:
Happens on real hardware with a Microsoft Sculpt wireless keyboard attached

[putty.log](https://github.com/managarm/managarm/files/6663200/putty.log)

avdgrinten:
This seems to be caused by a disconnect/re-connect cycle that EHCI does not handle correctly.

borrrden:
I compiled with all EHCI logging enabled, and the issue doesn't appear to happen in the same way.  It also makes serial logging stop working so I took a video and transcribed the following ehci statements:


> ehci: Starting driver (compiler on Oct 3 2021 01:44:58)
ehci: Detected controller
ehci: 8 ports
ehci: Controller uses 64-bit pointers
ehci: Extended capability: 1
ehci: OS access to the EHCI is already requested
ehci: Acquired OS <-> BIOS semaphore
ehci: Controller reset.
ehci: Awaiting IRQ event
ehci: IRQ event fired (sequence: 2), bits: 4
ehci: Checking ports
ehci: Device on port 1 is low-speed
ehci: Connect on port 7
ehci: Awaiting IRQ event
ehci: IRQ event fired (sequence: 3), bits: 4
ehci: Checking ports
ehci: Disconnect on port 1
ehci: Awaiting IRQ event
ehci: Port 7 was enabled
ehci: Setting device address
ehci: Linking in _linkTransaction
ehci: IRQ event fired (sequence: 4), bits: 1
ehci: Processing transfers
ehci: Transfer complete!
/var/lib/managarm-buildenv/build/system-root/usr/include/frg/expected.hpp:199: Assertion 'indicates
_
error(e)' failed!
thor: Panic in server /sbin/ehci

borrrden:
Also in case it is useful, here is `lsusb -v` from today on the same machine running Manjaro:

https://gist.github.com/borrrden/fe32bec312b2b791883e730799777082

borrrden:
I have figured out quite a bit more about this and the more I look at it, the more I am convinced that this state machine with its assertions is flawed and destined to fail.  These connections and such happen for multiple ports on the same controller and asynchronously so I don't think you are going to get a nice defined path of states.  If two ports connect at around the same time they are both going to be competing here.  I simply removed the assertions and the state variable all together and let it do its thing and this problem goes away.

PR #433, opened 2022-04-01
Fix multiple issues regarding EHCI stack and HID

borrrden:
- The assertions happening in the EHCI controller regarding clean state transition appear to not be reliable for more than one device being connected to the controller.  The port connections will be happening simultaneously and there is no guaranteed starting state for them to be in.
- Implemented TODO for handing off port to UHCI in the event that it is not a high speed device
- Added the FEATURE (0xB0) token to the list of recognized tokens for HID


borrrden:
This addresses #357 and #359

PR #434, opened 2022-04-03
UHCI: Fixed a modify while iterating bug

borrrden:
The progressSchedule function runs queue entries which potentially add more entries into the list of things that progressSchedule is iterating.  If the vector allocates more memory to accomdate the addition then the C++ spec says that all iterators are invalidated and thus the function will iterate in a bad way, probably right off the end and into undefined memory causing a GPF


avdgrinten (approved):
LGTM. Thank you for the PR!

borrrden:
Can this get merged then?

avdgrinten:
I was waiting for the CI, let's restart it...

borrrden:
Cool all green!

PR #435, opened 2022-04-03
core/virtio: Fix use-after-free in Queue

Dennisbonke:
This makes `netserver` not die randomly anymore, and is required for `webkitgtk` to load webpages.


avdgrinten (approved):
LGTM.

PR #437, opened 2022-04-09
drivers: nic: Add stub rtl8139 driver

YusufKhan-gamedev:
Adds untested, uncompiled, unanything rtl8139 driver that I made in a hurry. I dont really have plans to improve this in any capacity that would require using my brain so im just putting this PR out for someone to improve.

### DIFF
--- a/drivers/nic/rtl8139/src/rtl8139.cpp
+++ b/drivers/nic/rtl8139/src/rtl8139.cpp
@@ -1,23 +1,3 @@
-// SPDX-License-Identifier:  LGPL-3.0-or-later
-/*
- * Copyright 2022 Yusuf Khan
- *
- * This library is free software; you can redistribute it and/or
- * modify it under the terms of the GNU Lesser General Public
- * License as published by the Free Software Foundation; either
- * version 3.0 of the License, or (at your option) any later version.
- *
- * This library is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
- * Lesser General Public License for more details.
- *
- * You can find a copy of the GNU Lesser General Public License at
- * https://www.gnu.org/licenses/ or write to the Free Software
- * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
- *
- */
-
 #include <nic/rtl8139/rtl8139.hpp>
 #include <arch/io_space.hpp>
 #include <arch/dma_pool.hpp>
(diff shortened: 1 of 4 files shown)


YusufKhan-gamedev:
No, I dont know how to mark this as a draft.

YusufKhan-gamedev:
If anyone can, add this commit and fix any compile issues that may come up

avdgrinten:
This PR is not really *that* useful without any testing etc. If you want to try out if it actually works, feel free to sync with us on Discord or IRC.

Dennisbonke:
Apart from the aforementioned things, I see a LGPL license specifier. Not only do we not have those kinds of license headers in the code base, I don't know if we can mix LGPL and MIT freely. Anyone that can give some insight in that?

YusufKhan-gamedev:
> Apart from the aforementioned things, I see a LGPL license specifier. Not only do we not have those kinds of license headers in the code base, I don't know if we can mix LGPL and MIT freely. Anyone that can give some insight in that?

Pretty sure you can, see https://www.gnu.org/licenses/lgpl-3.0.en.html sections 4 and 5.

drivers/nic/rtl8139/src/rtl8139.cpp
```diff
@@ -0,0 +1,140 @@
...
+#include <protocols/fs/server.hpp>
+#include <protocols/hw/client.hpp>
+#include <protocols/mbus/client.hpp>
+#include <protocols/svrctl/server.hpp>
+
+namespace {
+
+uint16_t readPointer;
+/* pci port address */
+auto port;
```
Dennisbonke, line 33: Is auto here really needed? Same for some of the other types, I have a feeling this is a bit liberal with the auto use, which for me IMO makes the code less readable.

pitust: I am pretty sure this auto is not allowed too.

YusufKhan-gamedev: > I am pretty sure this auto is not allowed too.

Its not

YusufKhan-gamedev:
Errors Left:

_rtl8139.cpp.o -c ../../../src/managarm/drivers/nic/rtl8139/src/rtl8139.cpp
../../../src/managarm/drivers/nic/rtl8139/src/rtl8139.cpp:125:15: error: expression is not assignable
        frame.size() =  8208;
        ~~~~~~~~~~~~ ^
../../../src/managarm/drivers/nic/rtl8139/src/rtl8139.cpp:127:20: error: expression is not assignable
        frame.byte_data() = receiveOut;
        ~~~~~~~~~~~~~~~~~ ^
2 errors generated.

Working on a different project rn(was supposed to do it yesterday) that will improve my live in the immediate time so...bye for now

YusufKhan-gamedev:
If you want to pcik this up see https://github.com/openbsd/src/blob/master/sys/dev/ic/rtl81x9.c#L192 and fix the compile errors.

Dennisbonke:
Yeah that is not how this works, @qookei or @avdgrinten at this point just close it, you can reopen it when you're able to do some work on it that isn't pointing us at the openbsd drivers.

YusufKhan-gamedev:
Ah, and it doesnt update the pull request...great

PR #438, opened 2022-04-12
ehci: Fix multiple issues regarding controller stack and HID 

borrrden:
- The assertions happening in the EHCI controller regarding clean state transition appear to not be reliable for more than one device being connected to the controller.  The port connections will be happening simultaneously and there is no guaranteed starting state for them to be in.
- Implemented TODO for handing off port to UHCI in the event that it is not a high speed device
- Added the FEATURE (0xB0) token to the list of recognized tokens for HID

### DIFF
--- a/drivers/usb/devices/hid/src/main.cpp
+++ b/drivers/usb/devices/hid/src/main.cpp
@@ -412,6 +412,11 @@ void HidDevice::parseReportDescriptor(Device, uint8_t *p, uint8_t* limit) {
 			local = LocalState();
 			break;
 
+		case 0xB0:
+			if(logDescriptorParser)
+				printf("usb-hid:      Feature: 0x%x\n", data);
+			break;
+
 		case 0x80:
 			if(logDescriptorParser)
 				printf("usb-hid:     Input: 0x%x\n", data);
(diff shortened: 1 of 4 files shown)


Dennisbonke (approved):
It looks good to me but needs another review.

Dennisbonke:
@avdgrinten or @qookei can we review this PR?

ArsenArsen:
please split up this commit and reword it to align with the rest of the repository

drivers/usb/hcds/ehci/src/ehci.hpp
```diff
@@ -27,16 +27,14 @@ struct Enumerator {
 	// Called by the USB hub driver once a device completes reset.
 	void enablePort(int port);
 
-private:
-	enum class State {
-		null, resetting, probing
-	};
+	// Called by the USB hub driver if a port fails to enable after connection
+	void portDisabled(int port);
```
ArsenArsen, line 31: This function could be named better

drivers/usb/devices/hid/src/main.cpp
```diff
@@ -412,6 +412,11 @@ void HidDevice::parseReportDescriptor(Device, uint8_t *p, uint8_t* limit) {
 			local = LocalState();
 			break;
 
+		case 0xB0:
+			if(logDescriptorParser)
+				printf("usb-hid:      Feature: 0x%x\n", data);
```
ArsenArsen, line 417: there's an extra space here in the string

ArsenArsen (changes requested):

no92:
Superseded by #545.

PR #440, opened 2022-04-15
posix: Fix a duplicated bragi message ID

borrrden:
Between getpgid and setpgid

### DIFF
--- a/protocols/posix/posix.bragi
+++ b/protocols/posix/posix.bragi
@@ -465,7 +465,7 @@ head(128):
 	int64 pid;
 }
 
-message SetPgidRequest 81 {
+message SetPgidRequest 85 {
 head(128):
 	int64 pid;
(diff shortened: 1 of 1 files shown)


avdgrinten:
We should only renumber the duplicated ID here and keep the remaining IDs.

borrrden:
Interesting. I didn’t think it mattered since everything gets the updated number anyway but this means that either the numbers will no longer be in order in the file or getpgid and setpgid won’t appear next to each other in the list. Do you have a preference?

avdgrinten:
AFAICS the messages are not numbered by IDs anyway. So we can keep the order but change the ID.

borrrden:
Alright, this PR just got really tiny I love it 😄

avdgrinten:
LGTM now, thanks!

PR #441, opened 2022-04-16
nic: add support for Intel 82XX Gigabit Ethernet family

cleanbaja:
I am working on adding this family of well known and common Ethernet cards, and this pull request shows my progress in doing so.


YusufKhan-improves-shit:
Hi! Reading the BSD driver I can see that it is quite portable. Couldnt this be out of tree with patches to e1000_osdep.c. etc?

Just a thought

Dennisbonke:
> Hi! Reading the BSD driver I can see that it is quite portable. Couldnt this be out of tree with patches to e1000_osdep.c. etc?
> 
> Just a thought

Why would it be out of tree? We want these drivers in tree. So I don't see the usecase for that.

YusufKhan-improves-shit:
> Why would it be out of tree? We want these drivers in tree. So I don't see the usecase for that.

The BSD driver is licensed under the BSD 3-clause license, which is not allowed as managarm is strictly MIT.

ArsenArsen:
We can't really use an architecturally incompatible driver here. Wherever you decide to pull "`e1000_osdep.c`" (that doesn't sound like a whole driver) from will have been built for an entirely unrelated codebase, and unless we reuse an entire subsystem of an existing kernel, that makes no sense, as a "porting layer" would already far exceed such a simple driver.

cleanbaja:
to be continued, in the form of a realtek driver (intel e1000 is too legacy for today's usecase)

PR #442, opened 2022-04-17
posix: Set the controlling terminal of the process on open

borrrden:
If a session leader opens a terminal device without using O_NOCTTY then it should become the controlling terminal if the controlling terminal is not present.

Note: This depends on https://github.com/managarm/mlibc/pull/409

### DIFF
--- a/posix/subsystem/src/file.hpp
+++ b/posix/subsystem/src/file.hpp
@@ -15,6 +15,7 @@ struct File;
 struct MountView;
 struct FsLink;
 struct Process;
+struct ControllingTerminalState;
 
 struct FileHandle { };
 
@@ -274,6 +275,9 @@ public:
 
 	virtual async::result<frg::expected<Error, size_t>>
 	writeAll(Process *process, const void *data, size_t length);
+
+	virtual async::result<frg::expected<Error, ControllingTerminalState *>>
+	getControllingTerminal();
 
 	virtual FutureMaybe<ReadEntriesResult> readEntries();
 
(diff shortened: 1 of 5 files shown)


borrrden:
Oh, this also depends on #397 and I didn't realize that.  That PR added a new error code that I use.

ArsenArsen:
please hard-wrap your commit message, and reword them so that they explain what change was made precisely (i.e. `posix: add notTerminal error` rather that `add in a missing error`); the commits are also out of order, the first one does not build

borrrden:
Alright, updated!

posix/subsystem/src/file.hpp
```diff
@@ -275,6 +275,9 @@ struct File : private smarter::crtp_counter<File, DisposeFileHandle> {
 	virtual async::result<frg::expected<Error, size_t>>
 	writeAll(Process *process, const void *data, size_t length);
 
+	virtual async::result<frg::expected<Error>>
+	setControllingTerminal(Process *process);
```
avdgrinten, line 279: Isn't it easier to simply have a function that returns a `frg::expected<Error, std::shared_ptr<ControllingTerminalState>>`? The remainder of the logic should remain identical for all terminals.

borrrden: I don't think that this structure can actually be a shared pointer because it is not a standalone object, but rather a pointer inside of another struct (Channel), but I take your point otherwise about returning the value instead of having the file object perform the set.

borrrden: Alright I changed SET to GET

avdgrinten (approved):
LGTM.

PR #444, opened 2022-05-05
protocols/hw: Add new bus device protocol

borrrden:
For non-PCI bus devices (example ATA)

### DIFF
--- a/protocols/hw/hw.bragi
+++ b/protocols/hw/hw.bragi
@@ -27,6 +27,13 @@ struct PciCapability {
 	uint64 length;
 }
 
+struct BusDeviceMemoryInfo {
+	string tag;
+	IoType type;
+	uint64 address;
+	uint64 length;
+}
+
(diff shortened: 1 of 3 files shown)


Dennisbonke:
@borrrden can you rebase this PR so we can rerun ci?

borrrden:
Sure, I'm currently traveling for the next week and a half but after that I can get to it.

qookei:
Personally I'd prefer if we merged PCI and system bus devices into one thing in the hw protocol. We could for example used named memory regions for PCI BARs and MMIO/PIO (`bar0`-`bar5` for PCI and `mmio0`-`mmioN` and `pio0`-`pioN` for system bus). We'd still need some PCI-specific functions (MSI-related and whatnot). but those could just return errors for system bus devices.

Also worth noting is that having a good way of representing system bus devices is also needed for the AArch64 and RISC-V ports, for publishing devices described by the device tree.

PR #448, opened 2022-05-15
kernel: rename this_thread to thisThread

64:
Fixes a crash we see with alacritty.

Also renamed `this_thread` to `thisThread` for consistency with the other functions.


kernel/thor/generic/thread.cpp
```diff
@@ -100,24 +100,27 @@ void Thread::blockCurrent() {
 
 // FIXME: This function does not save the state! It needs to be given a ImageAccessor parameter!
```
avdgrinten, line 101: This comment can now be removed. Other than that, it LGTM.

PR #449, opened 2022-05-15
posix: handle sway launch option

no92:


posix/init/src/stage2.cpp
```diff
@@ -189,6 +189,12 @@ int main() {
 				throw std::runtime_error("dup2() failed");
 			close(fd);
 			execl(command.c_str(), command.c_str(), nullptr);
+		}else if(launch == "sway") {
+			setenv("WLR_DRM_NO_MODIFIERS", "1", 1);
+			setenv("WLR_BACKENDS", "drm,libinput", 1);
+			setenv("WLR_RENDERER_ALLOW_SOFTWARE", "1", 1);
+			setenv("WLR_XWAYLAND", "/usr/bin/Xwayland", 1);
```
ArsenArsen, line 196: why is this needed?

no92: I can't put my finger on why, but sometimes sway picked up different settings for those. This sets what works best for managarm.

avdgrinten (approved):
LGTM but we should wait until Sway is merged before adding this.

Dennisbonke (approved):
LGTM

PR #451, opened 2022-05-16
drivers/{xhci,virtio}: enableMsi after installMsi

64:
Since `installMsi` masks and `enableMsi` unmasks, it should really be done in this order.


avdgrinten:
AFAICS, the two functions are unrelated for MSI-X: one enables MSIs in general, the other one installs one MSI vector (there can be multiple vectors for MSI-X).

For plain MSI, this does not seem to be the case, right? Pinging @qookei since he added plain MSI support. Is there any way to unify the MSI and MSI-X behavior?

Dennisbonke:
Any consensus on if this is correct?

64:
Closing in favour of #507.

PR #453, opened 2022-05-20
meson: declare frigg as a dependency everywhere

64:
Some userspace programs didn't do this correctly. This is needed when we move the frigg headers from `/usr/include/frg` to `/usr/share/frigg/include/frg`, since the latter isn't picked up automatically.


avdgrinten (approved):
LGTM.

PR #458, opened 2022-06-03
kernel: increase log buffer from 100 to 256

64:
This should make some assertion failures more readable.

### DIFF
--- a/kernel/thor/generic/hel.cpp
+++ b/kernel/thor/generic/hel.cpp
@@ -117,15 +117,15 @@ HelError translateError(Error error) {
 HelError helLog(const char *string, size_t length) {
 	size_t offset = 0;
 	while(offset < length) {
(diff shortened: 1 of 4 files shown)


kernel/thor/generic/hel.cpp
```diff
@@ -117,15 +117,15 @@ HelError translateError(Error error) {
 HelError helLog(const char *string, size_t length) {
 	size_t offset = 0;
 	while(offset < length) {
-		auto chunk = frg::min(length - offset, size_t{100});
+		LogMessage log;
+		auto chunk = frg::min(length - offset, sizeof(log.text));
```
qookei, line 121: Why not just use `logLineLength` here?

avdgrinten:
The only thing that we should consider here is that we don't have infinite stack space in the kernel, so we shouldn't push the limit too far. 256 still seems affordable though.

PR #461, opened 2022-06-06
protocols/fs: Refactor the NODE_TRAVERSE_LINKS request to bragi

Dennisbonke:
As paths can become quite large, we should use the new bragi style messages with a variable sized tail to communicate those paths in `NODE_TRAVERSE_LINKS`. This fixes crashes on long paths, as I managed to trigger with Chromium.

Part of the Chromium on Managarm project.

### DIFF
--- a/protocols/fs/src/server.cpp
+++ b/protocols/fs/src/server.cpp
@@ -8,6 +8,7 @@
 #include <helix/ipc.hpp>
 
 #include <protocols/fs/server.hpp>
+#include <bragi/helpers-std.hpp>
 #include "fs.bragi.hpp"
 
 namespace protocols {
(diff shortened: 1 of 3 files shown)


protocols/fs/src/server.cpp
```diff
@@ -1122,8 +1123,20 @@ async::detached serveNode(helix::UniqueLane lane, std::shared_ptr<void> node,
 		auto conversation = accept.descriptor();
 
 		managarm::fs::CntRequest req;
-		req.ParseFromArray(recv_req.data(), recv_req.length());
+		auto preamble = bragi::read_preamble(recv_req);
+		assert(!preamble.error());
```
avdgrinten, line 1127: Should we print an error and return gracefully?

protocols/fs/src/server.cpp
```diff
@@ -1122,8 +1123,20 @@ async::detached serveNode(helix::UniqueLane lane, std::shared_ptr<void> node,
...
+
+		// managarm::posix::CntRequest req;
+		if (preamble.id() == managarm::fs::CntRequest::message_id) {
+			auto o = bragi::parse_head_only<managarm::fs::CntRequest>(recv_req);
+			if (!o) {
+				std::cout << "posix: Rejecting request due to decoding failure" << std::endl;
+				break;
+			}
+
+			req = *o;
```
avdgrinten, line 1138: Why is this variable called `o`?

protocols/fs/src/server.cpp
Dennisbonke: Could be possible. I’ll refactor it.

protocols/fs/src/server.cpp
Dennisbonke: No reason other than the fact that it was copied from the handling code in posix main.cpp

qookei (approved):

PR #471, opened 2022-08-08
posix: implement FIONREAD for fifo.read

no92:


posix/subsystem/src/fifo.cpp
```diff
@@ -8,8 +8,10 @@
 #include <async/recurring-event.hpp>
 #include <helix/ipc.hpp>
 #include "fifo.hpp"
+#include "fs.bragi.hpp"
 
 #include <experimental/coroutine>
+#include <sys/ioctl.h>
```
Dennisbonke, line 14: Missing newline after this line.

posix/subsystem/src/fifo.cpp
```diff
@@ -161,6 +163,38 @@ struct ReaderFile : File {
 		co_return 0;
 	}
 
+	async::result<void>
```
Dennisbonke, line 166: Missing newline.

Dennisbonke (changes requested):
LGTM apart from 2 styling issues. Where did you observe this being a problem?

posix/subsystem/src/fifo.cpp
```diff
@@ -161,6 +163,38 @@ struct ReaderFile : File {
...
+
+		switch(req.command()) {
+			case FIONREAD: {
+				resp.set_error(managarm::fs::Errors::SUCCESS);
+
+				if(_channel->packetQueue.empty()) {
+					resp.set_fionread_count(0);
+				} else {
+					auto packet = &_channel->packetQueue.front();
+					resp.set_fionread_count(packet->buffer.size() - packet->offset);
```
ArsenArsen, line 178: is there ever more than one element in the queue? iterating the whole thing would probably also avoid the size check

no92:

posix/subsystem/src/fifo.cpp
```diff
@@ -161,6 +164,41 @@ struct ReaderFile : File {
...
 
+	async::result<void>
+	ioctl(Process *process, managarm::fs::CntRequest req, helix::UniqueLane conversation) override {
+		managarm::fs::SvrResponse resp;
+
+		switch(req.command()) {
+			case FIONREAD: {
+				resp.set_error(managarm::fs::Errors::SUCCESS);
+
+				if(_channel->packetQueue.empty()) {
```
ArsenArsen, line 175: this check is now redundant as a result of using std::acumulate

posix/subsystem/src/fifo.cpp
```diff
@@ -161,6 +164,41 @@ struct ReaderFile : File {
...
+					resp.set_fionread_count(0);
+				} else {
+					size_t count = std::accumulate(_channel->packetQueue.cbegin(), _channel->packetQueue.cend(), 0, [](size_t sum, const Packet &p) {
```
ArsenArsen, line 178: this could also be `reduce` (since this operation does not rely on iteration order), but in practice, here it doesn't make a difference

ArsenArsen (changes requested):

ArsenArsen (approved):

PR #475, opened 2022-08-17
posix: Change signal/sigaction code to use SIG_DFL and SIG_IGN instead of hard coded constants

ElectrodeYT:


Dennisbonke (approved):
LGTM

PR #477, opened 2022-09-10
thor: handle pages marked uncached

no92:


qookei (approved):
This was originally omitted due to the fact that the MTRRs override WB with UC for MMIO on most (all?) machines, but it doesn't hurt to have for other use cases.

no92:
I managed to trip the assert below, so something with QEMU tripped that

qookei:
Ah wait, this is in the new cursor code, not the old mapping code. Yeah this is good to merge then.

PR #479, opened 2022-09-24
Split large compilation units

no92:


avdgrinten:
:+1: This is a good idea (didn't review everything carefully but overall it looks good).

Dennisbonke (approved):
LGTM

posix/subsystem/src/supercalls.cpp
```diff
@@ -0,0 +1,610 @@
+#include "gdbserver.hpp"
+#include "supercalls.hpp"
```
qookei, line 2: This file also handles other observations apart from supercalls (faults and such), perhaps we could pick a better name.

qookei (commented):
Looks good for the most part

no92:

qookei (approved):

PR #482, opened 2022-10-08
posix: Handle POLLWRNORM in the poll request

Dennisbonke:
This fixes `curl` crashing with unknown poll events and together with two mlibc PRs allows `curl` to work.

### DIFF
--- a/posix/subsystem/src/requests.cpp
+++ b/posix/subsystem/src/requests.cpp
@@ -2397,7 +2397,7 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 
 				// Translate POLL events to EPOLL events.
 				if(req.events(i) & ~(POLLIN | POLLPRI | POLLOUT | POLLRDHUP | POLLERR | POLLHUP
-						| POLLNVAL)) {
+						| POLLNVAL | POLLWRNORM)) {
 					std::cout << "\e[31mposix: Unexpected events for poll()\e[39m" << std::endl;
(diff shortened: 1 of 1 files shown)


posix/subsystem/src/requests.cpp
```diff
@@ -2407,6 +2407,8 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 				unsigned int mask = 0;
 				if(req.events(i) & POLLIN) mask |= EPOLLIN;
 				if(req.events(i) & POLLOUT) mask |= EPOLLOUT;
+				// TODO Dennis: Verify that this is correct, curl seems to think that it is.
```
ArsenArsen, line 2410: it should be

ArsenArsen (approved):

no92 (approved):

posix/subsystem/src/requests.cpp
Dennisbonke: I'll remove the comment then.

ArsenArsen:
this looks fine, but it might also be worth looking for other places in order to see whether these flags are synced up there (I have to wonder why they aren't one flag...)

Dennisbonke:
> this looks fine, but it might also be worth looking for other places in order to see whether these flags are synced up there (I have to wonder why they aren't one flag...)

Honestly we'll get there when we get there but inside posix probably nowhere as I turn it into `EPOLLOUT` at the entrypoint.

ArsenArsen:
yeah, the changes would just be the poll entries/exits in mlibc, no need to pass WRNORM to POSIX

PR #483, opened 2022-10-08
Some bragi recvInline fixes

Dennisbonke:
This PR addresses some missing reset calls after the use of a buffer that was received with `recvInline`. This should catch all uses in the Managarm codebase. Note that some of the users already did this, so not all users have been changed.


no92 (approved):
What's the use case of this `.reset()` call? Also, this might cause the `helix_ng` conversion PRs to need a rebase.

PR #488, opened 2022-10-11
Rework hel headers

no92:


ArsenArsen (approved):
seems okay, unsure if those types were truly unused (but I can't find any reference to them), so someone else will have to fill in their purpose

hel/include/hel-types.h
```diff
@@ -0,0 +1,107 @@
...
+
+//! Integer handle that represents a kernel resource.
+typedef int64_t HelHandle;
+
+typedef int64_t HelNanotime;
+
+
+enum {
+	// largest system call number plus 1
+	kHelNumCalls = 103,
```
avdgrinten, line 22: Syscall numbers are not really a "type" (so this should not be in `hel-types.h`).

hel/include/hel.h
```diff
@@ -225,15 +130,6 @@ struct HelAction {
...
 };
 
-enum {
-	kHelDescMemory = 1,
-	kHelDescAddressSpace = 2,
-	kHelDescThread = 3,
-	kHelDescEndpoint = 5,
-	kHelDescIrq = 9,
-	kHelDescIo = 10,
-};
```
avdgrinten, line 235: I think these were returned by `helDescriptorInfo` in a previous iteration.

no92:

Dennisbonke (approved):
LGTM

PR #489, opened 2022-10-12
posix: Guard against nullptrs in FSTATAT request

Dennisbonke:
Calling `associatedLink` on files that don't support it can lead to trying to call nullptr's. As this is bad, I figured that returning ENOENT is a good fallback (while keeping the log about which kind of file is being called intact).


no92 (approved):
Given that the default for `associatedLink` is to warn about null links already, LGTM.

PR #491, opened 2022-10-29
posix: Add /proc/[pid]/{status,cwd}

Dennisbonke:
This PR adds two new nodes in `/proc`, namely `/proc/[pid]/status` and `/proc/[pid]/cwd`. While the `status` node is lacking in comparison to the Linux counterpart, it does expose some information we didn't export before, and the layout matches what I have in my man page (see the comments in the file for more explanation). The `cwd` node is behaving the same as the Linux version, minus perhaps some permission stuff.

### DIFF
--- a/posix/subsystem/src/procfs.cpp
+++ b/posix/subsystem/src/procfs.cpp
@@ -265,6 +265,7 @@ std::shared_ptr<Link> DirectoryNode::createProcDirectory(std::string name,
 	proc_dir->directMkregular("comm", std::make_shared<CommNode>(process));
 	proc_dir->directMkregular("stat", std::make_shared<StatNode>(process));
 	proc_dir->directMkregular("statm", std::make_shared<StatmNode>(process));
+	proc_dir->directMkregular("status", std::make_shared<StatusNode>(process));
 
 	return link;
 }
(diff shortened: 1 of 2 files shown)


no92 (approved):

posix/subsystem/src/procfs.cpp
```diff
@@ -415,6 +417,108 @@ async::result<frg::expected<Error, FileStats>> RootLink::getStats() {
...
+	// See man 5 proc for more details.
+	// Based on the man page from Linux man-pages 6.01, updated on 2022-10-09.
+	std::stringstream stream;
+	auto path = _process->path();
+	size_t pos = path.rfind('/');
+	assert(pos != std::string::npos);
+	stream << "Name: " << path.substr(pos + 1) << "\n"; // Name is hardcoded to be the last part of the path
+	stream << "Umask: 0022\n"; // Hardcoded to 0022, which is what we hardcode in the mlibc sysdeps.
+	stream << "State: R\n"; // Hardcoded to R, running.
+	stream << "Tgid: " << _process->gid() << "\n"; // Thread group id, same as gid for now
```
qookei, line 431: Tgid should be same as Pid, not Gid:
```
              Tgid   Thread group ID (i.e., Process ID).

              Pid    Thread ID (see gettid(2)).
```

posix/subsystem/src/procfs.cpp
```diff
@@ -415,6 +417,108 @@ async::result<frg::expected<Error, FileStats>> RootLink::getStats() {
...
+	stream << "NGid: 0\n"; // NUMA Group ID, 0 if none.
+	stream << "Pid: " << _process->pid() << "\n";
+	// This avoids a crash when asking for the parent of init.
+	if(_process->pid() != 1) {
```
qookei, line 435: Perhaps here it'd be better to check whether `_process->getParent()` isn't null? In case if any other process doesn't ever have a parent

posix/subsystem/src/procfs.cpp
```diff
@@ -415,6 +417,108 @@ async::result<frg::expected<Error, FileStats>> RootLink::getStats() {
...
 
+async::result<std::string> StatusNode::show() {
+	// Everything that has a value of N/A is not implemented yet.
+	// See man 5 proc for more details.
+	// Based on the man page from Linux man-pages 6.01, updated on 2022-10-09.
+	std::stringstream stream;
+	auto path = _process->path();
+	size_t pos = path.rfind('/');
+	assert(pos != std::string::npos);
+	stream << "Name: " << path.substr(pos + 1) << "\n"; // Name is hardcoded to be the last part of the path
```
qookei, line 428: We now have a member for this

Dennisbonke:
Addressed the comments above.

qookei (approved):

PR #493, opened 2022-10-31
protocols/fs: fix error reply for PT_SOCKNAME

no92:
Previously, if sockname was unsupported, this resulted in `End of lane` errors. This prevented curl from working, which is fixed by this.


Dennisbonke:
LGTM

PR #494, opened 2022-11-04
Add i8254x NIC support

no92:


servers/netserver/src/main.cpp
```diff
@@ -23,11 +23,24 @@
...
+const std::string VENDOR_REDHAT = "1af4";
+
+std::set<std::string> nic_vendor_ids = {
+	VENDOR_REDHAT, /* virtio */
+	VENDOR_INTEL, /* Intel */
+};
+
+std::set<std::string> i8254x_device_ids = {
+	"100e", /* QEMU's e1000 device */
+};
```
ArsenArsen, line 38: these should not be std::strings, they're constant, they can be string views, and the set can be unordered

servers/netserver/src/main.cpp
```diff
@@ -49,6 +62,27 @@ async::result<void> doBind(mbus::Entity base_entity, virtio_core::DiscoverMode d
 	nic::runDevice(device);
 }
 
+async::result<void> doBindi8254x(mbus::Entity base_entity) {
```
ArsenArsen, line 65: this code should not be duplicated, only two lines of it depend on nic type

servers/netserver/include/netserver/nic.hpp
```diff
@@ -62,3 +63,10 @@ struct Link {
 
 async::detached runDevice(std::shared_ptr<Link> dev);
 } // namespace nic
+
+inline std::ostream &operator<<(std::ostream &os, const nic::MacAddress &mac) {
+	char buf[19];
+	sprintf(buf, "%02x:%02x:%02x:%02x:%02x:%02x", mac[0], mac[1], mac[2], mac[3], mac[4], mac[5]);
```
ArsenArsen, line 69: use snprintf or some other bound-checking operation, even if presumably safe

drivers/nic/i8254x/src/tx.cpp
```diff
@@ -0,0 +1,90 @@
...
+	co_await postDescriptor(payload, nic, &ev_req, raiseEvent);
+	co_await ev_req.event.wait();
+}
+
+async::result<void> TxQueue::postDescriptor(arch::dma_buffer_view payload, Intel8254xNic &nic, Request *req, async::detached (*complete)(Request *)) {
+	auto head = this->head();
+	auto tail = this->tail();
+	auto next = tail + 1;
+
+	if(logDebug) printf("i8254x/TxQueue: tx head=%zu tail=%zu next=%zu\n", head(), tail(), next());
```
ArsenArsen, line 32: these should probably be if constexpr

ArsenArsen: isn't that ub also (since it passes a non-integer type to zu, even if it happens to match bitwise)

drivers/nic/i8254x/src/i8254x.cpp
```diff
@@ -0,0 +1,217 @@
...
+			_rxQueue->ackAll();
+		}
+
+		HEL_CHECK(helAcknowledgeIrq(_irq.getHandle(), kHelAckAcknowledge, sequence));
+
+		uint32_t unhandled = uint32_t(status) & ~(handled);
+
+		if(unhandled) {
+			char *buffer;
+			asprintf(&buffer, "i8254x: unhandled IRQ with status %#x", unhandled);
```
ArsenArsen, line 204: btw frigg has a format_to container now

drivers/nic/i8254x/src/rx.cpp
```diff
@@ -0,0 +1,97 @@
...
+			memcpy(request->frame.data(), virt_addr, desc->length);
+
+			/* run the callback for packet reception */
+			request->complete(request);
+
+			/* remove the completed request from the queue */
+			_requests.pop();
+			/* reset the status field for reusing the descriptor */
+			desc->status = flags::rx::status::done(false);
+			/* advance the tail pointer to indicate successful handling of the packet to hardware */
```
ArsenArsen, line 75: does this assume unspecified order? (I don't know much about this nic)

drivers/nic/i8254x/src/i8254x.cpp
```diff
@@ -0,0 +1,217 @@
...
+	auto bar0 = co_await _device.accessBar(0);
+
+	_mmio_mapping = {bar0, barInfo.offset, barInfo.length};
+	_mmio = _mmio_mapping.get();
+
+	_mmio.store(regs::ctrl, flags::ctrl::reset(true));
+	/* p. 228, Table 13-3: To ensure that global device reset has fully completed and that the Ethernet controller
+	 * responds to subsequent access, wait approximately 1 µs after setting and before attempting to check to see
+	 * if the bit has cleared or to access any other device register. */
+	usleep(1);
```
ArsenArsen, line 37: is a cooperative sleep more appropriate here?

drivers/nic/i8254x/include/nic/i8254x/rx.hpp
```diff
@@ -0,0 +1,105 @@
...
+
+		memset(descriptor_space, 0, 2048 * _descriptor_count);
+
+		uintptr_t physical;
+		HEL_CHECK(helPointerPhysical(descriptor_space, &physical));
+
+		for(size_t i = 0; i < _descriptor_count; i++) {
+			_descriptors[i].address = physical + (i * 2048);
+		}
+	};
```
ArsenArsen, line 61: see Koronas comment about dma_buffer

ArsenArsen (commented):
not a thorough review (partially because I don't remember how to use the managarm apis ;), but this should be a start

drivers/nic/i8254x/src/tx.cpp
no92: `head()` et al. return a QueueIndex, which can be cast to a `size_t`. According to LSP, this is what happens implicitly (and which is why `-Wformat` does not complain)

Dennisbonke:
Is this still relevant after the FreeBSD e1000 driver?

PR #496, opened 2022-11-05
posix: Add /proc/[pid]/comm and /proc/[pid]/root

Dennisbonke:
This PR adds the `/proc/[pid]/comm` and `/proc/[pid]/root` nodes, exposing more information that might be of use to the userspace. They should both match Linux from what I can tell.

### DIFF
--- a/posix/subsystem/src/procfs.cpp
+++ b/posix/subsystem/src/procfs.cpp
@@ -261,6 +261,7 @@ std::shared_ptr<Link> DirectoryNode::createProcDirectory(std::string name,
 
 	proc_dir->directMknode("exe", std::make_shared<ExeLink>(process));
 	proc_dir->directMkregular("maps", std::make_shared<MapNode>(process));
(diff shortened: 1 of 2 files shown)


qookei (commented):
Seems good. `CommNode::show` may be broken if the path doesn't have a `/` at all (can that even happen?) and if the value of `std::string::npos` ever changes to anything but -1, so perhaps it'd be good to add an assert there that the `pos` is not `string::npos`.

Dennisbonke:
> Seems good. `CommNode::show` may be broken if the path doesn't have a `/` at all (can that even happen?) and if the value of `std::string::npos` ever changes to anything but -1, so perhaps it'd be good to add an assert there that the `pos` is not `string::npos`.

Added the assert as requested.

qookei (approved):

no92 (approved):

Geertiebear (commented):
I had a comment still :sweat_smile:

PR #498, opened 2022-11-07
Enhancements to /proc/[pid]/comm

Dennisbonke:
Addresses concerns raised in #496

### DIFF
--- a/posix/subsystem/src/process.cpp
+++ b/posix/subsystem/src/process.cpp
@@ -842,7 +842,10 @@ bool Process::checkOrRequestSignalRaise() {
 async::result<std::shared_ptr<Process>> Process::init(std::string path) {
 	auto hull = std::make_shared<PidHull>(1);
 	auto process = std::make_shared<Process>(std::move(hull), nullptr);
+	size_t pos = path.rfind('/');
+	assert(pos != std::string::npos);
 	process->_path = path;
+	process->_name = path.substr(pos + 1);
 	process->_vmContext = VmContext::create();
 	process->_fsContext = FsContext::create();
 	process->_fileContext = FileContext::create();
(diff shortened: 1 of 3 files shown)


posix/subsystem/src/process.cpp
```diff
@@ -842,7 +842,11 @@ bool Process::checkOrRequestSignalRaise() {
 async::result<std::shared_ptr<Process>> Process::init(std::string path) {
 	auto hull = std::make_shared<PidHull>(1);
 	auto process = std::make_shared<Process>(std::move(hull), nullptr);
+	size_t pos = path.rfind('/');
+	assert(pos != std::string::npos);
+	auto name = path.substr(pos + 1);
 	process->_path = path;
+	process->_name = name;
```
qookei, line 849: why not `process->_name = path.substr(pos + 1);` instead of the extra variable?

posix/subsystem/src/process.cpp
```diff
@@ -1080,7 +1086,11 @@ async::result<Error> Process::exec(std::shared_ptr<Process> process,
 	process->_fileContext->closeOnExec();
 
 	// "Commit" the exec() operation.
+	size_t pos = path.rfind('/');
+	assert(pos != std::string::npos);
+	auto name = path.substr(pos + 1);
 	process->_path = std::move(path);
+	process->_name = std::move(name);
```
qookei, line 1093: same as above (although here you'd need to switch the assignment to path and name as the path is moved away)

qookei (commented):
Looks fine apart from the nitpick comments

posix/subsystem/src/process.cpp
Dennisbonke: Because we need to be able to change the name from userspace (just write into /proc/[pid]/comm).

qookei: that does not prevent that? the local `name` variable is unnecessary

Dennisbonke: I'm sorry I misunderstood. You're correct. I'll fix it at the earliest possibility.

qookei (approved):

PR #500, opened 2022-11-17
drivers/kbd: allow for skipping scancode set selection

no92:
When booting on my ThinkPad Yoga 370, booting failed due to multiple failures that this commit addressed. Firstly, the identifier for the keyboard was 0xAB84, which has been whitelisted. Secondly, if setting the scancode fails (which happens on that machine), we just skip it as we can assume that scancode 2 is the default.

For good measure, we ensure that the keyboard is not disabled and that scancode set translation from set 2 to set 1 is not enabled in the PS/2 controller.


qookei (approved):

no92:
Merging despite CI failure as it is unrelated to this PR.

PR #502, opened 2022-12-14
Port of FreeBSD's e1000 driver

no92:


servers/netserver/src/main.cpp
```diff
@@ -59,23 +67,56 @@ async::result<protocols::svrctl::Error> bindDevice(int64_t base_id) {
...
+		if(auto device_str = std::get_if<mbus::StringItem>(&properties["pci-device"]); device_str) {
+			if(device_str->value == "1000")
+				discover_mode = virtio_core::DiscoverMode::transitional;
+			else if(device_str->value == "1041")
+				discover_mode = virtio_core::DiscoverMode::modernOnly;
+			else
+				co_return protocols::svrctl::Error::deviceNotSupported;
+		}
+
+		device = co_await setupVirtioDevice(std::move(hwDevice), discover_mode);
```
ArsenArsen, line 92: Why not dump this code into setupVirtioDevice?  We could reuse the same interface for both the Virtio and Intel paths.

drivers/nic/freebsd-e1000/src/managarm.cpp
```diff
@@ -0,0 +1,280 @@
...
+#include <net/ethernet.h>
+#include <protocols/hw/client.hpp>
+#include <unistd.h>
+
+#include <nic/freebsd-e1000/common.hpp>
+
+E1000Nic::E1000Nic(protocols::hw::Device device)
+	: nic::Link(1500, &_dmaPool), _device{std::move(device)},
+	_rxIndex(0, RX_QUEUE_SIZE), _txIndex(0, TX_QUEUE_SIZE) {
+	async::run(this->init(), helix::currentDispatcher);
```
ArsenArsen, line 50: why not just make this detached? ditto in other similar cases

ArsenArsen (commented):
not much to say, it's the driver that was there in fbsd, I'm a bit concerned about the multiple seconds of uncooperative sleeps this driver has in a few places, presumably because it assumes it works on it's own thread, but it should be fine seeing as it's in it's own server anyway

no92:
Rebased this onto current master, ready for review.

drivers/nic/freebsd-e1000/src/managarm.cpp
```diff
@@ -0,0 +1,282 @@
...
+	} else if (_hw.mac.type >= e1000_pch_spt) {
+		/**
+		 * In the new SPT device flash is not a separate BAR, rather it is also in BAR0,
+		 * so use the same tag and an offset handle for the FLASH read/write macros in the shared code.
+		 */
+
+		hw2flashbase(&_hw) = ((struct e1000_osdep *) _hw.back)->membase + E1000_FLASH_BASE_ADDR;
+	}
+
+	assert(e1000_setup_init_funcs(&_hw, true) == E1000_SUCCESS);
```
qookei, line 125: This pattern of `assert(function_with_side_effects());` is broken if you change the build type from `debugoptimized` to `release` or so. I noticed it in at least one other place in this file.

no92: Removed the function calls in asserts with the rebase.

PR #504, opened 2022-12-27
RTL8139/RTL8168 driver

no92:


drivers/nic/rtl8168/src/rtl8168.cpp
```diff
@@ -0,0 +1,218 @@
...
+	if(_model == Model::RTL8139) {
+		_mmio.store(regs::config1, 0);
+	} else if(_model == Model::RTL8168) {
+		_mmio.store(regs::cp_cmd, flags::cp_cmd::vlan_detag(true));
+	}
+
+	_mmio.store(regs::cmd, flags::cmd::reset(true));
+
+	while(_mmio.load(regs::cmd) & flags::cmd::reset)
+		usleep(1);
```
ArsenArsen, line 120: uhhhhh

drivers/nic/rtl8168/src/rtl8168.cpp
```diff
@@ -0,0 +1,218 @@
...
+	uint64_t sequence = 0;
+	while(true) {
+		auto await = co_await helix_ng::awaitEvent(_irq, sequence);
+		HEL_CHECK(await.error());
+		sequence = await.sequence();
+
+		auto status = _mmio.load(regs::interrupt_status);
+		_mmio.store(regs::interrupt_status, status);
+
+		printf("drivers/rtl8168: IRQ received status 0x%04x\n", uint16_t(status));
```
ArsenArsen, line 186: frg::format can print directly to std::cout now, better to do that than fall back to a possibly narrowing cast and a possibly wrong conversion specifier in C printfs

drivers/nic/rtl8168/src/rx.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+	co_return;
+}
+
+void RxQueue::handleRxOk() {
+	while(!_requests.empty()) {
+		auto req = _requests.front();
+		auto i = req->index;
+
+		if(_descriptors[i].flags & (1 << 31)) // Descriptor was not transmitted?
```
ArsenArsen, line 61: if this is an edge case, it might be good to log something about it?

drivers/nic/rtl8168/src/tx.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+	++tx_index;
+	nic._mmio.store(regs::tppoll, flags::tppoll::poll_normal_prio(true));
+
+	co_return;
+}
+
+void TxQueue8168::handleTxOk() {
+	for(size_t i = 0; i < _descriptor_count; i++) {
+		if(_descriptors[i].flags & (1 << 31)) // Descriptor was not transmitted?
```
ArsenArsen, line 62: these magic flags could use a name (same for above cases)

servers/netserver/include/netserver/nic.hpp
```diff
@@ -62,3 +63,10 @@ struct Link {
 
 async::detached runDevice(std::shared_ptr<Link> dev);
 } // namespace nic
+
+inline std::ostream &operator<<(std::ostream &os, const nic::MacAddress &mac) {
+	char buf[19];
+	sprintf(buf, "%02x:%02x:%02x:%02x:%02x:%02x", mac[0], mac[1], mac[2], mac[3], mac[4], mac[5]);
```
ArsenArsen, line 69: frg::fmt again (I believe this was written before that, right?)

ArsenArsen (commented):
cursory look, between lectures; I don't know whether the driver itself is correct yet though.

drivers/nic/rtl8168/src/rtl8168.cpp
```diff
@@ -0,0 +1,218 @@
...
+			_mmio.store(regs::tnpds_high, (_txQueue->getBase() >> 32) & 0xFFFFFFFF);
+			break;
+		}
+	}
+}
+
+void RealtekNic::configureQueues() {
+	switch(_model) {
+		case Model::RTL8139: {
+			_mmio.store(regs::transmit_config, flags::transmit_config::mxdma(flags::transmit_config::mxdma_2048) | flags::transmit_config::ifg(flags::transmit_config::ifg_normal));
```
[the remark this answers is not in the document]
Dennisbonke: We do indent cases in a switch statement, see the handbook.

drivers/nic/rtl8168/src/rtl8168.cpp
qookei: `co_await helix::sleepFor(1'000'000'000);`

drivers/nic/rtl8168/src/rtl8168.cpp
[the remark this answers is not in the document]
ArsenArsen: wtf?  I thought only broken editors did that

well whatever, that line is still too long

drivers/nic/rtl8168/include/nic/rtl8168/queue.hpp
```diff
@@ -0,0 +1,65 @@
...
+#include <stdint.h>
+
+struct [[gnu::packed]] Descriptor {
+	uint32_t flags;
+	uint32_t vlan;
+	uint32_t base_low;
+	uint32_t base_high;
+};
+
+struct QueueIndex {
```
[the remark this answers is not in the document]
no92: This was part of some other driver PRs, but has not been merged yet.

drivers/nic/rtl8168/src/rx.cpp
no92: That's not really an edge case; we get an IRQ once at least one RX descriptor is ready for read-out. This condition triggers once we reach the first descriptor that is in-flight (i.e. queued but not filled yet), which is expected to happen.

drivers/nic/rtl8168/include/nic/rtl8168/queue.hpp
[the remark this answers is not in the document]
ArsenArsen: let's not have two of them around, if you break that into it's own commit and own file, it should get skipped during a rebase when the other PR gets applied (or vice versa)

Dennisbonke (approved):
LGTM with my somewhat limited knowledge.

core/include/core/queue.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+#include <cstddef>
+
+/**
+ * Represents an index in a queue.
+ *
+ * This class serves to encapsulate the peculiarities of queue indices, especially their
+ * wrap-around modular arithmetic.
+ */
+struct QueueIndex {
```
Geertiebear, line 11: This seems like something that is better suited to maybe put in frigg?

drivers/nic/rtl8168/include/nic/rtl8168/queue.hpp
```diff
@@ -0,0 +1,24 @@
+#pragma once
```
Geertiebear, line 1: Why is this file named ``queue``?

drivers/nic/rtl8168/include/nic/rtl8168/tx.hpp
```diff
@@ -0,0 +1,43 @@
...
+#include <nic/rtl8168/common.hpp>
+#include <nic/rtl8168/queue.hpp>
+#include <queue>
+
+struct RealtekNic;
+
+struct TxQueue {
+	TxQueue(size_t descriptors, RealtekNic &nic);
+
+	virtual ~TxQueue() = default;
```
Geertiebear, line 13: What's the need for this virtual destructor?

drivers/nic/rtl8168/include/nic/rtl8168/tx.hpp
```diff
@@ -0,0 +1,43 @@
...
+	virtual async::result<void> postDescriptor(arch::dma_buffer_view frame, RealtekNic &nic, std::shared_ptr<Request> req) = 0;
+protected:
+	size_t _descriptor_count;
+	std::vector<arch::dma_buffer> _descriptor_buffers;
+	std::queue<std::shared_ptr<Request>> _requests;
+	arch::dma_array<Descriptor> _descriptors;
+	QueueIndex tx_index;
+};
+
+struct TxQueue8168 final : public TxQueue {
```
Geertiebear, line 31: It might be cleaner to use define these as template specializations instead of using inheritance, or is the polymorphism needed somewhere?

core/include/core/queue.hpp
ArsenArsen: this seems like it's part of a more general ring buffer style container

drivers/nic/rtl8168/include/nic/rtl8168/queue.hpp
```diff
@@ -0,0 +1,24 @@
...
+#include <core/queue.hpp>
+#include <stddef.h>
+#include <stdint.h>
+
+struct [[gnu::packed]] Descriptor {
+	arch::bit_value<uint32_t> flags{0};
+	uint32_t vlan;
+	uint32_t base_low;
+	uint32_t base_high;
+};
```
ArsenArsen, line 16: i'd add an assert on the size here too, just in case

drivers/nic/rtl8168/include/nic/rtl8168/rx.hpp
```diff
@@ -0,0 +1,28 @@
...
+#include <nic/rtl8168/common.hpp>
+#include <nic/rtl8168/queue.hpp>
+#include <queue>
+
+struct RealtekNic;
+
+struct RxQueue {
+	RxQueue(size_t descriptors, RealtekNic &nic);
+
+	uintptr_t getBase() {
```
ArsenArsen, line 13: could be const

drivers/nic/rtl8168/include/nic/rtl8168/tx.hpp
ArsenArsen: hmm, why do these even need to be keyed on the NIC if they behave identically?

drivers/nic/rtl8168/src/tx.cpp
```diff
@@ -0,0 +1,99 @@
...
+}
+
+async::result<void> TxQueue8168::postDescriptor(arch::dma_buffer_view payload, RealtekNic &nic, std::shared_ptr<Request> req) {
+	_requests.push(req);
+
+	auto desc = &_descriptors[tx_index()];
+
+	memcpy(_descriptor_buffers[tx_index()].data(), payload.data(), payload.size());
+
+	desc->flags |= flags::tx::ownership(flags::tx::owner_nic) | flags::tx::first_segment(true) |
```
ArsenArsen, line 42: hmm, should we break stuff like this into one line per flag rather than packing them?  there's a lot of shared text

ArsenArsen (commented):
i'll trust your judgement on the driver code itself since i have damn near zero experience with the RTLs this is a driver for.

seems good overall, a few comments left.

p.s. sorry about the delay.

drivers/nic/rtl8168/include/nic/rtl8168/tx.hpp
qookei: They don't behave identically, see `src/tx.cpp`

qookei: Also re: templates, eventually you will need virtual somewhere (here or higher up) and IMO putting it on the TxQueue is cleaner

Geertiebear: Ah alright if it's needed than that's fine.

core/include/core/queue.hpp
Geertiebear: Yeah it also seems like it's used this way in the code. Might be cleaner to just use a straight up ring-buffer container.

drivers/nic/rtl8168/include/nic/rtl8168/tx.hpp
no92: gcc complains otherwise: `warning: delete called on 'TxQueue' that is abstract but has non-virtual destructor`

drivers/nic/rtl8168/include/nic/rtl8168/rx.hpp
no92: If I make `getBase` constant, gcc complains about the `operator[]` method not being const.

no92:
Addressed the comments above and rebased on master.

drivers/nic/rtl8168/include/nic/rtl8168/common.hpp
```diff
@@ -9,23 +9,107 @@
...
 };
 
+
 struct RealtekNic : nic::Link {
 	friend TxQueue8168;
-	friend TxQueue8139;
 public:
 	RealtekNic(protocols::hw::Device device);
 
+	enum MacRevision : uint16_t {
```
ArsenArsen, line 36: we haven't added it yet but surround this code with clang-format tags if you're intentionally indenting unconventionally

ElectrodeYT:
The driver is now capable of sending and receiving packets somewhat reliably on my revision of the RTL8168 card (MacVer46, rtl8168h).
Support for most other cards is not yet implemented, and the code is still extremely ugly and hacked together. I will try and improve the code a bit and also add support for a few other RTL network cards i can test with.

drivers/nic/rtl8168/include/nic/rtl8168/common.hpp
ElectrodeYT: i honestly cant remember why i did this, but i removed the enums in question since there is no way they could ever be needed

Dennisbonke (approved):
LGTM

Dennisbonke (approved):
LGTM

PR #506, opened 2023-01-03
kernel: Fix broken handling of of libgcc

ArsenArsen:
With the GCC13 release, this hack will become obsolete, because I'll fix up the -kernel toolchain to not require libc, startfiles or libm at all, which would permit us to simply not pass -nostdlib instead.  This hinges on a few outstanding patches that I've already sent upstream, and maybe an extra ten or so lines of code that would be added locally.

For now, this is good enough.  Hell of a lot better than redistributing a libgcc.a in git anyway.


Dennisbonke (approved):
LGTM

Issue #463, opened 2022-06-07
XHCI MSIs are broken on Intel Panther Point chipset


qookei:
No IRQs (for command completion for example) arrive unless the driver is forced to use PCI pin IRQs.

avdgrinten:
Any clue why this happens?

qookei:
Presumably related to the discussion in #451.

qookei:
Fixed by #507.

PR #507, opened 2023-01-09
thor: Fix PCI MSI handling around EnableMsi and InstallMsi requests

qookei:
Closes #451.

Also fixes panic messages being truncated after first 256 bytes.


qookei:
Just tested and it also fixes #463.

PR #509, opened 2023-01-18
docs: Clarify braces usage in single line if, while, for loops

Dennisbonke:
Personally, it improves readability. I left it out in the initial version because the discussion about it was going nowhere. Let's reopen it!

### DIFF
--- a/docs/src/contributing/coding-style.md
+++ b/docs/src/contributing/coding-style.md
@@ -13,9 +13,11 @@ This section lists the most important coding style guidelines that Managarm foll
 - Braces are placed similarly to existing code.
 - Single line bodies of `if`, `while`, `for` all go to a seperate line, like
 ```c
-if(expression)
+if(expression) {
 	one-line statement;
+}
 ```
+- Single line bodies of `if`, `while`, `for` should use braces for readability.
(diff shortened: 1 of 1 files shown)


avdgrinten:
My opinion on this has not changed; I still dislike this.

What about adding `-Werror=misleading-indentation` instead of changing the style?

Dennisbonke:
Personally, I'm in favor of both. Makes it really clear what is guarded behind the statement and avoids weird indentation that misleads people.

avdgrinten:
Can you give an example where the indentation is confusing?

This arguably not confusing in general. Python doesn't use any braces and people can read Python just fine.

All else being equal, shorter code should be preferred, and additional braces increase the vertical space required to render the code.

Dennisbonke:
This PR is obsolete based upon this line in the coding style.

```
Braces shall follow "K&R Bracing Style variant One True Bracing". The K&R Bracing Style was first introduced in The C Programming Language by Brian Kernighan and Dennis Ritchie.
```

I do agree with adding the `-Werror=misleading-indentation` flag to the compilation flags, I'll get to work on doing that in a separate PR.

PR #513, opened 2023-02-05
Posix wait

ikbenlike:
This PR aims to implement the `pid = 0` and `pid <= -1` cases in `Process::wait`. It also adds checks in `init-stage2` to ensure that processes launched with `execl` actually get executed.


posix/init/src/stage2.cpp
```diff
@@ -31,8 +31,7 @@ int main() {
 	std::cout << "init: Starting udevd" << std::endl;
 	auto udev = fork();
 	if(!udev) {
-//		execl("/usr/sbin/udevd", "udevd", nullptr);
-		execl("/usr/sbin/udevd", "udevd", "--debug", nullptr);
+		assert(-1 != execl("/usr/sbin/udevd", "udevd", "--debug", nullptr));
```
ArsenArsen, line 34: Building with NDEBUG should not remove udev.

This definitely classifies as assert abuse.

Instead, we can write a wrapper macro that calls `error()`

posix/subsystem/src/process.cpp
```diff
@@ -1157,15 +1157,30 @@ async::result<void> Process::terminate(TerminationState state) {
...
 	while(true) {
 		for(auto it = _notifyQueue.begin(); it != _notifyQueue.end(); ++it) {
-			if(pid > 0 && pid != it->pid())
+			if(pid > 0 && pid != it->pid()){
+				continue;
+			}
+			else if(pid == 0 && groupid != it->gid()){
+				continue;
+			}
+			else if(pid < -1 && groupid != it->gid()){
```
ArsenArsen, line 1179: Why are these cases distinct?

posix/subsystem/src/requests.cpp
```diff
@@ -300,24 +300,40 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+			if(!pid_error) {
+				auto error_code = pid_error.error();
+				if(error_code == WaitError::noChild) {
+					resp.set_error(managarm::posix::Errors::NO_CHILD);
+				}
+				else if(error_code == WaitError::interrupted) {
+					resp.set_error(managarm::posix::Errors::WAIT_INTERRUPTED);
+				}
+				else {
+
```
ArsenArsen, line 318: We really need a better way to map managarm errors to posix (not something for you to address, just saying)

ArsenArsen (changes requested):

posix/subsystem/src/process.cpp
ikbenlike: I did it like this because I was just going down the list of cases in the man-page; they do mean different things but because they all just `continue;` it could also be one singular if-statement if so desired

posix/init/src/stage2.cpp
ikbenlike: Good point. I'll write a wrapper that calls `error()` and then terminates by either calling `exit()` or `abort()` (whichever is deemed more appropriate).

posix/subsystem/src/requests.cpp
ikbenlike: Agreed, doing it this way feels clunky and the amount of code overhead isn't really optimal. I was thinking about just returning a POSIX error from `Process::wait`, but that didn't seem appropriate either

posix/init/src/stage2.cpp
ArsenArsen: error(3) terminates after printing the error if passed a nonzero value as the exit code

ikbenlike: This has been implemented; however due to the absense of `error()` in mlibc this is stuck on [mlibc#793](https://github.com/managarm/mlibc/pull/793)

qookei (changes requested):
Indentation is wrong in some places (4 spaces instead of a tab). Also `else` goes on the same line as the closing `}` of the `if` body (as in: `} else {`)

no92:
Superseded by #760.

PR #516, opened 2023-03-01
core/virtio: print PCI capability names for debugging

no92:


core/virtio/src/core.cpp
```diff
@@ -587,6 +588,23 @@ void StandardPciQueue::notifyTransport() {
 	_transport->_notifySpace().store(_notifyRegister, queueIndex());
 }
 
+std::map<uint8_t, std::string> cap_names = {
```
ArsenArsen, line 591: best to use an unordered_map

core/virtio/src/core.cpp
```diff
@@ -617,8 +635,9 @@ discover(protocols::hw::Device hw_device, DiscoverMode mode) {
 			auto bir = co_await hw_device.loadPciCapability(i, 4, 1);
 			auto offset = co_await hw_device.loadPciCapability(i, 8, 4);
 			auto length = co_await hw_device.loadPciCapability(i, 12, 4);
-			std::cout << "virtio: Subtype: " << subtype << ", BAR index: " << bir
-					<< ", offset: " << offset << ", length: " << length << std::endl;
+			std::cout << "virtio: Subtype: " << capName(subtype).value_or("<invalid>") << " ("
```
ArsenArsen, line 638: this can be formatted nicer but otherwise lgtm

PR #518, opened 2023-03-12
DRM fixes for virtio-gpu

no92:


ElectrodeYT (approved):

core/drm/include/core/drm/mode-object.hpp
```diff
@@ -153,12 +153,12 @@ struct Encoder : ModeObject {
 };
 
 struct ConnectorState {
-	ConnectorState(std::weak_ptr<Connector> connector) : connector(connector) {};
+	ConnectorState(std::weak_ptr<Connector> connector) : connector{connector}, crtc{}, encoder{} {};
```
avdgrinten, line 156: No action needed by that doesn't do anything (?).

core/drm/include/core/drm/mode-object.hpp
```diff
@@ -253,21 +253,21 @@ struct Plane : ModeObject {
 };
 
 struct PlaneState {
-	PlaneState(std::weak_ptr<Plane> plane) : plane(plane) {};
+	PlaneState(std::weak_ptr<Plane> plane) : plane{plane}, crtc{}, fb{} { };
```
avdgrinten, line 256: Same here.

avdgrinten (approved):

core/drm/include/core/drm/mode-object.hpp
qookei: Indeed it doesn't, `shared_ptr` has a non-trivial constructor that accepts no arguments, so it gets called automatically.

PR #510, opened 2023-01-23
Additions for `iproute2`, mostly centered on rtnetlink

no92:

PR #521, opened 2023-03-31
netserver: implement operators for `CidrAddress` and `Route`

no92:
These were previously part of #510, but got dropped mistakenly.


servers/netserver/src/ip/ip4.cpp
```diff
@@ -48,10 +48,28 @@ bool operator<(const CidrAddress &lhs, const CidrAddress &) {
 	return std::tie(lhs.prefix, lhs.ip) < std::tie(lhs.prefix, lhs.ip);
 }
 
-bool operator<(const Route &lhs, const Route &rhs) {
+bool operator==(const CidrAddress &lhs, const CidrAddress &rhs) {
+	return (lhs.ip == rhs.ip) && (lhs.prefix == rhs.prefix);
+}
```
ArsenArsen, line 53: why is this new?

PR #523, opened 2023-04-02
posix: Rewrite two bragi messages to use the new format

Dennisbonke:
`WebKitGTK` did some long paths on `ACCESSAT` and `MKDIRAT`, so I rewrote them to account for this.

Blocked on managarm/mlibc#822


no92 (approved):
LGTM

PR #524, opened 2023-04-08
drivers/nic/virtio: Put the sending and sent frame logs behind a guard

Dennisbonke:
Reduces e9 spam by a lot when working with `WebKitGTK`.


no92 (approved):
LGTM

PR #525, opened 2023-04-09
posix: Remove several netlink asserts

Dennisbonke:
WebKitGTK ran into these asserts, but commenting them out broke nothing. Seems to me that some if not all of these asserts are thus either wrong or not needed.

I'm not sure if this is correct, so someone with knowledge about this should review this.

### DIFF
--- a/posix/subsystem/src/nl-socket.cpp
+++ b/posix/subsystem/src/nl-socket.cpp
@@ -110,9 +110,9 @@ public:
 		using namespace protocols::fs;
 		if(logSockets)
 			std::cout << "posix: Recv from socket \e[1;34m" << structName() << "\e[0m" << std::endl;
-		assert(!flags);
-		// assert(!(flags & ~(MSG_DONTWAIT | MSG_CMSG_CLOEXEC)));
-		assert(max_addr_length >= sizeof(struct sockaddr_nl));
+		if(!(flags & ~(MSG_DONTWAIT | MSG_CMSG_CLOEXEC))) {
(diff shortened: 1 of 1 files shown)


ikbenlike:
it might be better to turn some of these into error returns to propagate the error up the call chain - but then it would probably be good to test what impact this would have on WebKitGTK

ArsenArsen (changes requested):
as he said..  each we can't just randomly remove these

Dennisbonke:
I'm aware that we can't just remove them (unless they are wrong obviously). I was hoping to get some information on what to do with them (turn into some error return? Change the assert? Log unsupported flags for the flags assert?).

posix/subsystem/src/nl-socket.cpp
```diff
@@ -110,9 +110,7 @@ struct OpenFile : File {
 		using namespace protocols::fs;
 		if(logSockets)
 			std::cout << "posix: Recv from socket \e[1;34m" << structName() << "\e[0m" << std::endl;
-		assert(!flags);
```
no92, line 113: We should probably warn about flags we don't support, e.g. by printing the flags `if(!(flags & ~(MSG_DONTWAIT | MSG_CMSG_CLOEXEC)))`, as is done in similar fashion elsewhere, too. Unsupported flags should be silently ignored AFAICT (looking at the man page and POSIX).

no92 (changes requested):
We can mostly drop these asserts, but some minor modifications might be needed to make it correct.

no92 (approved):
LGTM

PR #526, opened 2023-04-09
netserver: Stub setOption for SO_PASSCRED on netlink sockets

Dennisbonke:
Required for `WebKitGTK` to work.

### DIFF
--- a/servers/netserver/src/netlink/netlink.hpp
+++ b/servers/netserver/src/netlink/netlink.hpp
@@ -41,9 +41,16 @@ public:
 		co_return protocols::fs::Error::none;
 	}
 
+	static async::result<void> setOption(void *, int option, int) {
+		assert(option == SO_PASSCRED);
+		std::cout << "netserver: netlink sockets do not support SO_PASSCRED yet!" << std::endl;
+		co_return;
+	}
+
 	static async::result<size_t> sockname(void *, void *, size_t);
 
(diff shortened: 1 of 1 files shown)


no92 (changes requested):
Not supplying `SCM_CREDENTIALS` here seems like a bug, according to the (unclearly written) man page. I'd suggest simply copying the implementation from `nl-socket.cpp`:

https://github.com/managarm/managarm/blob/9124706b79480ea49ca892c2abd2c79e067f82d4/posix/subsystem/src/nl-socket.cpp#L140-L150

Dennisbonke:
> Not supplying `SCM_CREDENTIALS` here seems like a bug, according to the (unclearly written) man page. I'd suggest simply copying the implementation from `nl-socket.cpp`:
> 
> https://github.com/managarm/managarm/blob/9124706b79480ea49ca892c2abd2c79e067f82d4/posix/subsystem/src/nl-socket.cpp#L140-L150

Seems simple enough, but the `Packet` struct for netlink in netserver lacks the ever important `senderPid` member, and it'd be a massive refactor to get that properly set (`recvMsg` and friends in posix have a `Process *process` as their first argument, which is lacking in `recvMsg` in netserver). Is it refactor time or can we get away without refactoring this?

Dennisbonke:
Closed in favor of #535

PR #528, opened 2023-04-28
meta: Add .clang-format file

Dennisbonke:
BLOCKED ON managarm/bootstrap-managarm#233

Initial groundwork for this PR was laid by @ArsenArsen, I just improved on it and did some testing. I believe this matches our style guide pretty closely. If accepted, I'll do a styling run and PR this same .clang-format into mlibc and supporting repositories where this makes sense.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,222 @@
+﻿---
+# commented out settings are not available in current host-llvm-toolchain but
+# could still prove useful
+Language: Cpp
+
+# I don't see any stable default for this, lets just hope LLVM is sane enough
+BasedOnStyle: LLVM
+
+# private: should line up with struct
+AccessModifierOffset: -8
+
+# alignment is mostly problematic since it requires modifying multiple lines on
+# a change to keep things aligned...
+AlignArrayOfStructures: None
+AlignConsecutiveAssignments: false
+AlignConsecutiveBitFields: false
+AlignConsecutiveMacros: false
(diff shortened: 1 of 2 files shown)


Dennisbonke:
I request reviews by @avdgrinten and @qookei

Dennisbonke:
@avdgrinten @qookei @ArsenArsen I once again request reviews now that llvm 16 is building on xbbs.

ElectrodeYT (approved):
LGTM

moodyhunter:
a gentle ping on this :)

qookei:
The current configuration produces some unoptimal results when splitting long binary/ternary expressions, like here:
image

Personally I think the indentation should always be at least one tab wide, and in this case 2 tabs are beneficial since IMO otherwise the condition blends in with the body.

AFAICT this is controlled by [AlignOperands](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#alignoperands).

ArsenArsen:
IMO this is alignment, not indentation, and

```
if (EXPR1
    OP expr2)
```

is more readable than

```
if (EXPR1
        OP EXPR2)
```

qookei:
I suppose this is just preference (I don't like the alignment being narrower than tab width). IMO it should be fine either way, and shouldn't be that problematic since it's enforced by clang-format.

no92:
Updated the `.clang-format` to align more with current managarm style. As of note, the include grouping has been defined properly.

PR #532, opened 2023-05-02
thor: Remove a not needed assert in remapPresentPagesByCursor

Dennisbonke:
As discussed on Discord, the assert is not correct, and should be changed to an if statement.


no92 (commented):
Merging the two ifs into one would be nice. Do we need to handle the non-present case?

Other than that, LGTM.

no92 (approved):
LGTM

PR #535, opened 2023-07-09
netserver: Stub netlink SO_PASSCRED

Dennisbonke:
I'm pretty sure 'WebKitGTK' needed it, but we don't see it called. The `SO_PASSCRED` implementation is lacking things, but it works for 'WebKitGTK', so I want it in.

Supercedes #526


no92 (approved):
LGTM

PR #537, opened 2023-07-13
netserver: send correct destination MAC address in the ARP body

ElectrodeYT:


ArsenArsen:
what makes this correct?

no92 (approved):
LGTM

ArsenArsen:
that's not incorrect, though.  that field is ignored for arp requests (which are necessarily the type of request sent with all-zero targets in the netserver).

if you observe routers breaking on this for some reason, though, then I don't mind the change (but reword the commit to explain that).

there shouldn't be a need to ask for a followup like this when reviewing git history.

ArsenArsen:
I don't think it's ignored, the RFC provides an example where it sets that field to `don't care`.

again, since this field is ignored, broadcast is fine here too.

note that this isn't done for *multicast* requests, this is done for all requests of any kind.  in addition, if a field is meant to be ignored there's no single correct way to fill it, so saying that this 'fills it correctly' is not accurate

ArsenArsen:
it isn't being filled with correct information, though, just information that appeases some implementation for some reason.

if you prefer, I can reword and merge, skipping a CI run

ElectrodeYT:
Sorry if I was being unclear, but by "correct" I don't mean "the spec requires it to be correct", but with the correct information we have: we have a target field, and we have a target; it should be filled correctly.

Feel free to reword it if you really want to, but just like this entire discussion I don't personally think it is necessary.

Dennisbonke:
LGTM

Dennisbonke (approved):

PR #538, opened 2023-07-16
Two exec fixes required for chromium

Dennisbonke:
The first fix is allowing read only mappings in libraries, ICU loves them.
The second fix is ignoring `PT_NOTE`, as we should.


qookei (approved):

ElectrodeYT (approved):
LGTM

no92 (approved):
LGTM

PR #542, opened 2023-07-22
thor: disable I/O and memory decode before reading BAR length

no92:
This prevents devices from spuriously decoding addresses leading to invalid reads. This tends to happen mostly with PCI passthrough devices in my experience.


ElectrodeYT (approved):
LGTM

PR #547, opened 2023-07-28
Request cancellation

Geertiebear:
Various changes needed to support canceling in progress requests.

- Kernel changes to allow a blocked thread to be interrupted by ``helInterruptThread()``
- Passing cancellation tokens through the async call chains (not implemented for quite some requests)
- Receiving a ``OneshotEvent`` from the user process to signal cancellation.

Depends on managarm/mlibc#884.


hel/include/helix/ipc-structs.hpp
```diff
@@ -120,6 +120,10 @@ struct DismissResult {
 		_valid = true;
 	}
 
+	void set_error(HelError error) {
+		_error = error;
```
no92, line 124: This `set_error` method is the only one not to set `_valid = true;`; is this correct?

posix/subsystem/src/requests.cpp
```diff
@@ -10,6 +10,7 @@
 
 #include <helix/timer.hpp>
 
+#include "helix/ipc-structs.hpp"
```
no92, line 13: Is this header required or erroneously included by LSP?

kernel/thor/generic/thread.cpp
```diff
@@ -61,28 +61,28 @@ void Thread::migrateCurrent() {
 	}, &this_thread->_executor);
 }
 
-void Thread::blockCurrent() {
+bool Thread::blockCurrent(bool interruptible) {
```
no92, line 64: Would be nice to document what the return value means.

no92 (commented):
Minor nitpicks, but looks good so far.

posix/subsystem/src/requests.cpp
qookei: I'd guess it's clangd's doing considering the quotes

posix/subsystem/src/inotify.cpp
```diff
@@ -8,6 +9,7 @@
 #include "fs.hpp"
 #include "inotify.hpp"
 #include "process.hpp"
+#include "protocols/fs/common.hpp"
```
ElectrodeYT, line 12: Is this LSP or is it actually required?

posix/subsystem/src/nl-socket.cpp
```diff
@@ -64,27 +65,32 @@ struct OpenFile : File {
...
 				std::cout << "posix: netlink socket would block" << std::endl;
-			co_return Error::wouldBlock;
+			co_return std::make_pair(protocols::fs::Error::wouldBlock, 0);
 		}
 
-		while(_recvQueue.empty())
-			co_await _statusBell.async_wait();
+		while(_recvQueue.empty()) {
+			co_await _statusBell.async_wait(ce);
+			if (logSockets)
```
ElectrodeYT, line 81: Should this print be under the cancellation request or is it OK here?

ElectrodeYT (approved):
Aside from a couple tiny comments, LGTM.

posix/subsystem/src/inotify.cpp
Geertiebear: probably LSP but either way I think it should be there.

posix/subsystem/src/nl-socket.cpp
Geertiebear: I'm actually not entirely sure _what_ this print is meant to inform the user about so maybe i'll just remove it.

posix/subsystem/src/inotify.cpp
qookei: Even then it should use `<>` not `""`

Geertiebear: thats true

kernel/thor/generic/thor-internal/thread.hpp
```diff
@@ -117,15 +133,33 @@ struct Thread final : smarter::crtp_counter<Thread, ActiveHandle>, ScheduleEntit
...
+
+		return true;
 	}
 
-	template<typename Sender>
-	requires (!std::is_same_v<typename Sender::value_type, void>)
-	static typename Sender::value_type asyncBlockCurrent(Sender s, WorkQueue *wq) {
+	template<typename Sender, typename Tag = asyncBlockCurrentNormal>
+	requires ( (std::is_invocable_v<Sender, async::cancellation_token> && !std::is_same_v<typename std::invoke_result_t<Sender, async::cancellation_token>::value_type, void>)
+			 || !std::is_same_v<typename Sender::value_type, void>)
```
avdgrinten, line 147: The `void` version of this function (i.e., `std::is_same_v<typename Sender::value_type, void>` case) uses a bool return to indicate that the operation has been cancelled. This version (= the non-void one) doesn't use any indication AFAICT. This should be unified. One option is returning a `frg::expected`, another other option is adding an out parameter.

kernel/thor/generic/thor-internal/thread.hpp
```diff
@@ -72,16 +72,30 @@ struct Thread final : smarter::crtp_counter<Thread, ActiveHandle>, ScheduleEntit
...
 	}
 
-	template<typename Sender>
-	requires std::is_same_v<typename Sender::value_type, void>
-	static void asyncBlockCurrent(Sender s, WorkQueue *wq) {
+	template<typename Sender, typename Tag = asyncBlockCurrentNormal>
+	requires
+	( std::is_invocable_v<Sender, async::cancellation_token> && std::is_same_v<typename std::invoke_result_t<Sender, async::cancellation_token>::value_type, void>)
+	  || std::is_same_v<typename Sender::value_type, void>
+	static bool asyncBlockCurrent(Sender s, WorkQueue *wq, Tag tag = asyncBlockCurrentNormal{}) {
```
avdgrinten, line 87: This function prototype is quite hard to read (especially since there is no documentation).

avdgrinten: It's a bit confusing that `Sender` is either a sender or a sender factory function. Maybe it would be better to always take a factory in the main overload. Then we could add a two argument overload (without the tag) that takes a sender and delegates to the main overload.

kernel/thor/generic/thread.cpp
```diff
@@ -352,7 +357,15 @@ void Thread::interruptOther(smarter::borrowed_ptr<Thread> thread) {
 	// TODO: Perform the interrupt immediately if possible.
 
 //	assert(thread->_pendingSignal == kSigNone);
+
 	thread->_pendingSignal = kSigInterrupt;
+	lock.unlock();
```
avdgrinten, line 362: Instead of using explicit `unlock()` calls, we should put the lock into a scope.

kernel/thor/generic/thread.cpp
```diff
@@ -352,7 +357,15 @@ void Thread::interruptOther(smarter::borrowed_ptr<Thread> thread) {
...
+
+	// If the thread is blocked and can be interrupted,
+	// then unblock it to notify.
+	if (thread->_runState == kRunInterruptableBlocked)
```
avdgrinten, line 366: Looking at `_runState` without the lock taken is not safe. We can determine whether we want to unblock while we still have the lock.

kernel/thor/generic/hel.cpp
```diff
@@ -2991,21 +2991,25 @@ HelError helFutexWait(int *pointer, int expected, int64_t deadline) {
...
 				[&] (async::cancellation_token cancellation) {
 					return generalTimerEngine()->sleep(deadline, cancellation);
-				}
-			)
-		);
+				});
+			}, Thread::asyncBlockCurrentInterruptible{}
+			)) {
+			return kHelErrCancelled;
+		}
```
avdgrinten, line 3051: In this case (the one with the timeout), we do nothing with `ct`. We probably need an overload of `race_and_cancel` that takes a cancellation token to make this work. We can probably leave the timeout case as a TODO (since it's not needed in mlibc).

no92:
Unfortunately, I don't know how to fix the remaining comments. Help would be appreciated!

kernel/thor/generic/hel.cpp
```diff
@@ -247,41 +247,49 @@ HelError helCreateUniverse(HelHandle *handle) {
 	return kHelErrNone;
 }
 
-HelError helTransferDescriptor(HelHandle handle, HelHandle universe_handle,
+HelError helTransferDescriptor(HelHandle handle, HelHandle universe_handle, int in,
```
ElectrodeYT: Might be a good idea to document what `in` here is; I'm not well versed in this code, and I can not understand what that argument means right now

kernel/thor/generic/thor-internal/thread.hpp
avdgrinten: Actually, _not returning whether cancellation happened_ is more consistent with other high level APIs that we have. So the return of the `void` overload can be switched back from `bool` to `void`. Call sites that need to know whether cancellation happened or not can use either the return value or out of band mechanisms to communicate that.

posix/subsystem/src/requests.cpp
```diff
@@ -459,7 +464,14 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 
 			logRequest(logRequests, "WAIT", "pid={}", req.pid());
 
-			auto wait_result = co_await self->wait(req.pid(), flags);
+			async::cancellation_event ce;
+			([] (helix::UniqueDescriptor event,
+			     async::cancellation_event &ce) -> async::detached {
+				co_await helix_ng::awaitEvent(event, 1);
+				ce.cancel();
+			})(cancel_event.descriptor(), ce);
```
avdgrinten: This pattern is problematic since `ce` may go out of scope.

kernel/thor/generic/thread.cpp
```diff
@@ -470,7 +470,9 @@ Thread::~Thread() {
 void Thread::dispose(ActiveHandle) {
 	if(logCleanup)
 		urgentLogger() << "thor: Killing thread due to destruction" << frg::endlog;
+
 	_kill();
+	workOnExecutor(&_executor);
```
avdgrinten: This doesn't make sense. Might be from an earlier iteration of the code (at least I don't see what this does -- it causes the thread to enter the work queue the next time it is scheduled but the thread is killed anyway?).

kernel/thor/generic/thor-internal/work-queue.hpp
```diff
@@ -22,6 +23,7 @@ struct Worklet {
 private:
 	smarter::shared_ptr<WorkQueue> _workQueue;
 	void (*_run)(Worklet *);
+	bool _cancelled;
```
avdgrinten: As far as I can tell, this variable is never used.

kernel/thor/generic/thor-internal/work-queue.hpp
```diff
@@ -94,12 +96,17 @@ struct WorkQueue {
...
 			worklet_.setup([] (Worklet *base) {
 				auto self = frg::container_of(base, &EnterOperation::worklet_);
-				async::execution::set_value_noinline(self->r_);
+				async::execution::set_value_noinline(self->r_, base->_cancelled);
 			}, wq_);
 			if(enter(&worklet_)) {
-				async::execution::set_value_inline(r_);
+				async::execution::set_value_inline(r_, false);
 				return true;
 			}
```
avdgrinten: This returns `true` if the submission failed which is counterintuitive. Instead, we should return `true` if entering the work queue was successful. Also, it should be documented what the return value of `.enter()` means.

kernel/thor/generic/thor-internal/thread.hpp
```diff
@@ -367,8 +367,9 @@ struct Thread final : smarter::crtp_counter<Thread, ActiveHandle>, ScheduleEntit
 	AssociatedWorkQueue _pagingWorkQueue;
 
 	Mutex _mutex;
-
+public:
 	RunState _runState;
+private:
```
avdgrinten: This should not be necessary and it's problematic since `_runState` is protected by the thread's mutex which should not be taken outside of the `Thread` class.

kernel/thor/generic/thread.cpp
```diff
@@ -416,9 +421,15 @@ void Thread::interruptOther(smarter::borrowed_ptr<Thread> thread) {
 	auto lock = frg::guard(&thread->_mutex);
 
 	// TODO: Perform the interrupt immediately if possible.
+	// assert(thread->_pendingSignal == kSigNone);
 
-//	assert(thread->_pendingSignal == kSigNone);
 	thread->_pendingSignal = kSigInterrupt;
```
avdgrinten: This assertion should not be commented out (otherwise we may e.g., overwrite a kSigKill by a kSigInterrupt).

kernel/thor/generic/thread.cpp
avdgrinten: The last iteration of the code changed this to call `unblockOther()` with the thread lock still taken. That is problematic since `unblockOther()` takes the same lock and deadlocks.

The easiest fix is to check the run state while the lock is taken and to call into `unblockOther()` after the lock is released:
```cpp
bool unblock;
{
    // Take lock, update _pendingSignal etc.
    // [...]
    unblock = (thread->_runState == kRunInterruptableBlocked);
}
if (unblock)
    unblockOther(thread);
```

kernel/thor/generic/hel.cpp
```diff
@@ -247,41 +247,49 @@ HelError helCreateUniverse(HelHandle *handle) {
...
 		HelHandle *out_handle) {
 	auto this_thread = getCurrentThread();
 	auto this_universe = this_thread->getUniverse();
 
 	AnyDescriptor descriptor;
-	smarter::shared_ptr<Universe> universe;
+	smarter::shared_ptr<Universe> dst_universe;
```
avdgrinten: Local variables use `camelCase` (it probably makes sense to convert all local vars in this function to that style).

kernel/thor/generic/hel.cpp
```diff
@@ -247,41 +247,49 @@ HelError helCreateUniverse(HelHandle *handle) {
...
+			dst_universe = universe_it->get<UniverseDescriptor>().universe;
+		}
+
+		auto src_universe = this_universe;
+		if (in) {
+			src_universe = dst_universe;
+			lock = Universe::Guard(src_universe->lock);
+
+			dst_universe = this_universe.lock();
 		}
```
avdgrinten: This chunk is a bit confusing because it swaps `src_universe` and `dst_universe`. I'd be cleaner to first get the `universe` (= a variable with a neutral name) from the `universe_handle`. Afterwards, the code should decide whether it's the source or destination (instead of swapping these vars).

kernel/thor/generic/hel.cpp
avdgrinten: Maybe this should be an enum with `kHelTransferIn` and `kHelTransferOut` variants.

kernel/thor/generic/thor-internal/work-queue.hpp
no92: It's never written to, but read in line 106.

hel/include/helix/ipc.hpp
```diff
@@ -365,6 +365,19 @@ struct AwaitClock : Operation {
...
+		return result_.error;
+	}
+
+private:
+	void parse(const void *ptr) override {
+		memcpy(&result_, ptr, sizeof(result_));
+	}
+
+	HelEventResult result_;
+};
```
avdgrinten: There is already an `AwaitEventSender` that should be used instead. We can adjust it to take an optional `cancellation_token`. `Submission` / `Operation` is an obsolete interface.

protocols/fs/src/server.cpp
```diff
@@ -198,7 +198,18 @@ async::detached handlePassthrough(smarter::shared_ptr<void> file,
...
+			{
+				async::cancellation_callback cb{{ce}, [&] {
+					HEL_CHECK(helCancelAsync(helix::Dispatcher::global().acquire(),
+							awaitEvent.asyncId()));
+				}};
+				co_await submit.async_wait();
+			}
+
 			ce.cancel();
 		})(cancel_event.descriptor(), ce);
```
avdgrinten: This block should use `race_and_cancel` instead to race until (1) `file_ops->read()` is completed or (2) the event is raised.

kernel/thor/generic/hel.cpp
```diff
@@ -2984,21 +2984,38 @@ HelError helFutexWait(int *pointer, int expected, int64_t deadline) {
...
-				},
-				[&] (async::cancellation_token cancellation) {
-					return generalTimerEngine()->sleep(deadline, cancellation);
-				}
-			)
-		);
+		if (!Thread::asyncBlockCurrentInterruptible(
+		        [&](async::cancellation_token ct) {
+			        return async::race_and_cancel(
+			            [&](async::cancellation_token cancellation) {
```
avdgrinten: Sender factories with captures in `race_and_cancel` should be avoided (either by wrapping them in `async::lambda` or by using something like `std::bind_front` / `std::bind_back`).

kernel/thor/generic/hel.cpp
```diff
@@ -2984,21 +2984,38 @@ HelError helFutexWait(int *pointer, int expected, int64_t deadline) {
...
+				            return generalTimerEngine()->sleep(deadline, cancellation);
+			            },
+			            [&](async::cancellation_token cancellation) -> coroutine<void> {
+				            async::cancellation_event ce;
+				            async::cancellation_callback cb1{ct, [&ce]() { ce.cancel(); }};
+				            async::cancellation_callback cb2{cancellation, [&ce]() {
+					                                             ce.cancel();
+				                                             }};
+				            co_await async::suspend_indefinitely({ce});
```
avdgrinten: While this works, it would be preferable to make `race_and_cancel` itself take a `cancellation_token`.

kernel/thor/generic/thor-internal/thread.hpp
```diff
@@ -122,30 +149,44 @@ struct Thread final : smarter::crtp_counter<Thread, ActiveHandle>, ScheduleEntit
 				// might have consumed the unblock latch.
 				continue;
 			}
-			Thread::blockCurrent();
+			if (!Thread::blockCurrent(interruptible)) {
+				ce.cancel();
+				return false;
+			}
```
avdgrinten: We can't just return false here. When we return, `operation` is destructed but the async operation is still ongoing which can cause a UAF.

We can should do `ce.cancel()` (only once per call to this function) but continue afterwards (this time with `interruptible = false`).

kernel/thor/generic/thor-internal/thread.hpp
```diff
@@ -168,13 +211,17 @@ struct Thread final : smarter::crtp_counter<Thread, ActiveHandle>, ScheduleEntit
 				// might have consumed the unblock latch.
 				continue;
 			}
-			Thread::blockCurrent();
+			if (!Thread::blockCurrent(interruptible)) {
+				ce.cancel();
+			}
```
avdgrinten: Likewise.

kernel/thor/generic/hel.cpp
```diff
@@ -3157,78 +3157,101 @@ HelError helAcknowledgeIrq(HelHandle handle, uint32_t flags, uint64_t sequence)
...
+	void complete() override {
+		frg::destruct(*kernelAlloc, this);
+	}
+
+private:
+	Worklet worklet;
+	async::cancellation_event cancelEvent;
+	smarter::shared_ptr<IpcQueue> _queue;
+	QueueSource source;
+	HelEventResult result;
```
avdgrinten: This should be zero initialized as not all members will be overwritten in the cancelled case.

kernel/thor/generic/hel.cpp
```diff
@@ -3157,78 +3157,101 @@ HelError helAcknowledgeIrq(HelHandle handle, uint32_t flags, uint64_t sequence)
...
-		static void awaited(Worklet *worklet) {
-			auto closure = frg::container_of(worklet, &EventClosure::worklet);
-			closure->result.error = translateError(closure->eventNode.error());
-			closure->result.sequence = closure->eventNode.sequence();
-			closure->result.bitset = closure->eventNode.bitset();
-			closure->_queue->submit(closure);
+		void handleCancellation() override {
+			_cancelled = true;
+			cancelEvent.cancel();
 		}
```
avdgrinten: The handling of `_cancelled` is asymmetric in EventClosure vs. IrqClosure. I'd suggest to either make it symmetric, or to remove it from `IrqClosure`.

drivers/libblockfs/src/libblockfs.cpp
```diff
@@ -85,18 +86,18 @@ async::result<protocols::fs::Error> flock(void *object, int flags) {
...
 	protocols::ostrace::Timer timer;
 	auto self = static_cast<ext2fs::OpenFile *>(object);
-
 	if(self->inode->fileType == FileType::kTypeDirectory) {
 		co_return protocols::fs::Error::isDirectory;
 	}
 
-	co_await self->inode->readyJump.wait();
+	if(!co_await self->inode->readyJump.wait(ce))
+		co_return protocols::fs::Error::interrupted;
```
avdgrinten: Block file systems do not support `EINTR` on UNIX.

drivers/nic/usb_net/src/usb-mbim.cpp
```diff
@@ -85,17 +85,19 @@ write(void *object, helix_ng::CredentialsView credentials, const void *buffer, s
...
 	(void) credentials;
 
 	auto self = static_cast<CdcWdmDevice *>(object);
 
-	auto p = co_await self->nic->queue_.async_get();
-	assert(p);
+	auto p = co_await self->nic->queue_.async_get(ct);
+	if (!p && ct.is_cancellation_requested())
+		co_return protocols::fs::Error::interrupted;
+	assert(p && !ct.is_cancellation_requested());
```
avdgrinten: This assert can trigger if cancellation is triggered from a separate thread.

drivers/uart/src/main.cpp
```diff
@@ -234,7 +234,9 @@ read(void *, helix_ng::CredentialsView , void *buffer, size_t length) {
 	if(!recvBuffer.empty())
 		completeRecvs();
 
-	co_await req.event.wait();
+	co_await req.event.wait(ce);
+	if (ce.is_cancellation_requested())
+		co_return protocols::fs::Error::interrupted;
 	co_return req.progress;
```
avdgrinten: There are two issues here:
1. We must only return EINTR if there is no progress.
2. This will not remove the request from the request queue.

hel/include/helix/ipc-structs.hpp
```diff
@@ -122,6 +122,11 @@ struct DismissResult {
...
+		_valid = true;
+	}
```
avdgrinten: Where are the new set_error functions actually used? I don't see any code changes in `hel/` that make use of them.

posix/subsystem/src/file.hpp
```diff
@@ -385,16 +389,17 @@ struct File : private smarter::crtp_counter<File, DisposeFileHandle> {
 	virtual async::result<frg::expected<Error, off_t>>
 	seek(off_t offset, VfsSeek whence);
 
-	virtual async::result<frg::expected<Error, size_t>>
-	readSome(Process *process, void *data, size_t max_length);
+	virtual async::result<protocols::fs::ReadResult>
+	readSome(Process *process, void *data, size_t max_length,
```
avdgrinten: `protocols::fs::ReadResult` is `std::variant<Error, size_t>` which seems strictly worse than `frg::expected<Error, size_t>`. Either do a conversion in `ptRead()` (as was done before this PR) or change `protocols::fs::ReadResult` to `frg::expected`. If we don't want to blow up this PR we can keep the conversion and refactor to `frg::expected` in a separate PR.

posix/subsystem/src/exec.cpp
```diff
@@ -214,12 +214,12 @@ execute(ViewPath root, ViewPath workdir,
 			}
 
 			char buffer[128];
-			auto readResult = co_await execFile->readSome(nullptr, buffer, 128);
-			if(!readResult) {
+			auto readResult = co_await execFile->readSome(nullptr, buffer, 128, {});
+			if (std::get_if<protocols::fs::Error>(&readResult)) {
```
avdgrinten: This demonstrates why `std::variant` is more ugly than `frg::expected`.

posix/subsystem/src/file.hpp
avdgrinten: Also, we should keep the POSIX-internal error type for POSIX-internal functions.

posix/subsystem/src/inotify.cpp
```diff
@@ -103,15 +105,19 @@ struct OpenFile : File {
...
-	async::result<frg::expected<Error, size_t>>
-	readSome(Process *, void *data, size_t maxLength) override {
-		if(_queue.empty() && nonBlock_) {
-			co_return Error::wouldBlock;
+	async::result<protocols::fs::ReadResult>
+	readSome(Process *, void *data, size_t maxLength, async::cancellation_token ce) override {
+		if (_queue.empty() && nonBlock_)
+			co_return protocols::fs::Error::wouldBlock;
+		else if (_queue.empty() && ce.is_cancellation_requested())
+			co_return protocols::fs::Error::interrupted;
```
avdgrinten: This else is unnecessary since the loop below will exit in that case anyway.

posix/subsystem/src/process.hpp
```diff
@@ -289,15 +289,25 @@ struct SignalContext {
 
 	CheckSignalResult checkSignal();
 
-	// TODO: If we ever need to cancel this operation, it would be better to
-	//       take a cancellation token instead of nonBlock.
-	async::result<SignalItem *> fetchSignal(uint64_t mask, bool nonBlock, async::cancellation_token ct = {});
+	async::result<SignalItem *> fetchSignal(uint64_t mask, async::cancellation_token ct);
```
avdgrinten: This should be reverted to the previous state (and the comment should be deleted). Passing in an already cancelled token seems hacky to me.

posix/subsystem/src/observations.cpp
```diff
@@ -101,6 +101,9 @@ async::result<void> observeThread(std::shared_ptr<Process> self,
 		std::shared_ptr<Generation> generation) {
 	auto thread = self->threadDescriptor();
 
+	SignalItem *delayedSignal = nullptr;
+	std::optional<SignalContext::SignalHandling> delayedSignalHandling = std::nullopt;
```
avdgrinten, line 106: Nit: the delayed signal stuff is in the same commit as all the read changes. A separate commit would be cleaner.

posix/subsystem/src/observations.cpp
```diff
@@ -353,9 +370,24 @@ async::result<void> observeThread(std::shared_ptr<Process> self,
...
+				async::cancellation_event ev;
+				ev.cancel();
+				auto active = co_await self->signalContext()->fetchSignal(~self->signalMask(), {ev});
+				if(active) {
+					auto handling = self->signalContext()->determineHandling(active, self.get());
+					killed = handling.killed;
+					self->cancelEvent();
+					co_await self->signalContext()->raiseContext(active, self.get(), handling);
+				}
+			}
```
avdgrinten: Is the `else` case needed here? There should always be a delayed signal if `superSigRaise` is entered (except for userspace misbehavior, i.e., if it calls `superSigRaise` randomly in which case we can just log a warning and continue).

posix/subsystem/src/observations.cpp
```diff
@@ -557,8 +607,22 @@ async::result<void> observeThread(std::shared_ptr<Process> self,
...
+				auto raceDeath = [&generation](async::cancellation_token c) -> async::result<void> {
+					async::cancellation_event ce;
+					async::cancellation_callback cb1{c, [&ce]{
+						ce.cancel();
+					}};
+					async::cancellation_callback cb2{generation->cancelServe, [&ce]{
+						ce.cancel();
+					}};
+					co_await async::suspend_indefinitely({ce});
 				};
```
avdgrinten: This can be replaced by a cancelable `race_and_cancel`.

avdgrinten: Also, this is part of the read commit.

posix/subsystem/src/process.cpp
```diff
@@ -653,30 +676,31 @@ async::result<void> SignalContext::raiseContext(SignalItem *item, Process *proce
 					launchGdbServer(process);
 					co_await async::suspend_indefinitely({});
 				}
-				[[fallthrough]];
+				break;
 			default:
-				std::cout << "posix: Thread killed as the result of signal "
-						<< item->signalNumber << std::endl;
```
avdgrinten, line 670: Do we want to keep this log for debugging purposes? For example by moving it to the `process->terminate()` call below.

posix/subsystem/src/signalfd.cpp
```diff
@@ -23,14 +24,21 @@ void OpenFile::serve(smarter::shared_ptr<OpenFile> file) {
...
+	async::cancellation_event ev;
+	if (_nonBlock) {
+		ev.cancel();
+		ce = async::cancellation_token{ev};
+	}
+	auto active = co_await process->signalContext()->fetchSignal(_mask, ce);
+	if (!active && _nonBlock)
+		co_return protocols::fs::Error::wouldBlock;
+	if (!active && !_nonBlock)
+		co_return protocols::fs::Error::interrupted;
```
avdgrinten: Needs to be revisited when reverting the `fetchSignal()` signature.

posix/subsystem/src/sysfs.cpp
```diff
@@ -99,33 +100,39 @@ async::result<frg::expected<Error, off_t>> AttributeFile::seek(off_t offset, Vfs
 	co_return _offset;
 }
 
-async::result<frg::expected<Error, size_t>>
-AttributeFile::readSome(Process *process, void *data, size_t max_length) {
+async::result<protocols::fs::ReadResult>
+AttributeFile::readSome(Process *process, void *data, size_t max_length, async::cancellation_token ce) {
+	// TODO(geert): do something with this.
+	(void)ce;
```
avdgrinten: Reading `AttributeFile` finishes in finite time, so this comment can be removed.

posix/subsystem/src/sysfs.cpp
```diff
@@ -99,33 +100,39 @@ async::result<frg::expected<Error, off_t>> AttributeFile::seek(off_t offset, Vfs
...
+		// TODO(geert): Don't assume this doesn't block.
 		if(auto res = co_await node->_attr->show(node->_object); res) {
 			_buffer = res.value();
 			_cached = true;
-		} else
-			co_return res.error();
+		} else {
+			// TODO(geert): Change show() to return a protocols::fs::Error as well.
+			co_return protocols::fs::Error::brokenPipe;
+		}
```
avdgrinten: I don't understand why the error handling was changed here.

posix/subsystem/src/tmp_fs.cpp
```diff
@@ -663,12 +664,12 @@ MemoryFile::writeAll(Process *, const void *buffer, size_t length) {
...
 }
 
-async::result<frg::expected<Error, size_t>>
+async::result<protocols::fs::ReadResult>
 MemoryFile::pread(Process *, int64_t offset, void *buffer, size_t length) {
 	auto node = static_cast<MemoryNode *>(associatedLink()->getTarget().get());
 
 	if(static_cast<size_t>(offset) >= node->_fileSize)
-		co_return 0;
+		co_return protocols::fs::Error::endOfFile;
```
avdgrinten: This commit contains a lot of `0` -> `endOfFile` changes that are unrelated to the cancellation implementation.

posix/subsystem/src/un-socket.cpp
```diff
@@ -153,33 +154,49 @@ struct OpenFile : File {
...
-			}
-		);
+			if(ct.is_cancellation_requested())
+				co_return protocols::fs::Error::interrupted;
 
-		if(_recvQueue.empty())
-			co_return Error::wouldBlock;
+			co_await async::race_and_cancel(
+				[&](async::cancellation_token c) { return raceReceiveTimeout(c); },
+				[&](async::cancellation_token c) -> async::result<void> {
```
avdgrinten: `race_and_cancel` sender factory with captures.

posix/subsystem/src/un-socket.cpp
```diff
@@ -153,33 +154,49 @@ struct OpenFile : File {
...
+						if (c.is_cancellation_requested())
+							co_return;
+					}
+				},
+				[&](async::cancellation_token c) -> async::result<void> {
+					async::cancellation_event ce;
+					async::cancellation_callback cb1{ct, [&ce]() { ce.cancel(); }};
+					async::cancellation_callback cb2{c, [&ce]() { ce.cancel(); }};
+					co_await async::suspend_indefinitely({ce});
+				}
```
avdgrinten: This should use a cancelable `race_and_cancel`.

posix/subsystem/src/requests.cpp
```diff
@@ -459,11 +463,27 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 
 			logRequest(logRequests, "WAIT", "pid={}", req.pid());
 
-			auto wait_result = co_await self->wait(req.pid(), flags);
+			auto event = cancel_event.descriptor();
+			std::string data;
+			data.resize(req.size());
```
avdgrinten: AFAICT this string does nothing.

hel/include/hel.h
```diff
@@ -646,8 +651,9 @@ HEL_C_LINKAGE HelError helCreateUniverse(HelHandle *handle);
 //!    	Handle to the destination universe.
 //! @param[out] outHandle
 //!    	Handle to the copied descriptor (valid in the universe specified by @p universeHandle).
-HEL_C_LINKAGE HelError helTransferDescriptor(HelHandle handle, HelHandle universeHandle,
-		HelHandle *outHandle);
+HEL_C_LINKAGE HelError
+helTransferDescriptor(HelHandle handle, HelHandle universeHandle,
+		enum HelTransferDescriptorFlags direction, HelHandle *outHandle);
```
avdgrinten, line 656: `hel` changes need to be tested with Rust.

posix/subsystem/src/process.hpp
```diff
@@ -722,10 +723,14 @@ struct Process : std::enable_shared_from_this<Process> {
 	helix::UniqueDescriptor _threadPageMemory;
 	helix::Mapping _threadPageMapping;
 
+	helix::UniqueDescriptor _cancelEventMemory;
+	helix::Mapping _cancelEventMapping;
```
avdgrinten, line 730: `_cancelEventMapping` is used in the read commit but only added in the last commit.

kernel/thor/generic/hel.cpp
```diff
@@ -247,41 +247,57 @@ HelError helCreateUniverse(HelHandle *handle) {
...
 				return kHelErrBadDescriptor;
-			universe = universe_it->get<UniverseDescriptor>().universe;
+			universe = universeIt->get<UniverseDescriptor>().universe;
 		}
+
+		if(direction == kHelTransferDescriptorOut) {
+			srcUniverse = thisUniverse;
+			dstUniverse = universe;
+		} else {
+			assert(direction == kHelTransferDescriptorIn);
```
avdgrinten, line 282: You can only assert this if you validate `direction` before. Otherwise, this assert can be triggered from user space.

kernel/thor/generic/hel.cpp
```diff
@@ -247,41 +247,57 @@ HelError helCreateUniverse(HelHandle *handle) {
...
+	auto thisThread = getCurrentThread();
+	auto thisUniverse = thisThread->getUniverse();
 
 	AnyDescriptor descriptor;
-	smarter::shared_ptr<Universe> universe;
-	{
-		auto irq_lock = frg::guard(&irqMutex());
-		Universe::Guard lock(this_universe->lock);
+	smarter::borrowed_ptr<Universe> srcUniverse;
+	smarter::shared_ptr<Universe> dstUniverse;
```
avdgrinten: Both of these need to be `shared_ptr` as they can both potentially refer to other universes.

PR #548, opened 2023-08-01
virtio-gpu driver rework

no92:


drivers/gfx/virtio/src/spec.hpp
```diff
@@ -11,18 +11,54 @@ namespace format {
...
+	inline constexpr uint32_t ctxDetachResource = 0x203;
+	inline constexpr uint32_t create3d = 0x204;
+	inline constexpr uint32_t transferToHost3d = 0x205;
+	inline constexpr uint32_t transferFromHost3d = 0x206;
+	inline constexpr uint32_t submit3d = 0x207;
+	inline constexpr uint32_t resourceMapBlob = 0x208;
+	inline constexpr uint32_t resourceUnmapBlob = 0x209;
+	/* cursor commands */
+	inline constexpr uint32_t updateCursor = 0x300;
+	inline constexpr uint32_t moveCursor = 0x301;
```
ikbenlike, line 42: would it not make more sense to instead turn this into an enum class or something like that?

no92: Given that this is how it used to work before, and that changing this in this PR would require editing most commits, I'd prefer to keep it this way for now and change it later.

no92:
First half of virtio-gpu changes, split up due to size. The other half will deal with 3D acceleration.

drivers/gfx/virtio/src/spec.hpp
ikbenlike: fair enough

drivers/gfx/virtio/src/commands.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+private:
+	virtio_core::Queue *_queue;
+	virtio_core::Handle _descriptor;
+	std::coroutine_handle<> _handle;
+};
+
+async::result<void> Cmd::transferToHost2d(uint32_t width, uint32_t height, uint32_t resourceId, GfxDevice *device) {
+	spec::XferToHost2d xfer;
+	memset(&xfer, 0, sizeof(spec::XferToHost2d));
```
ElectrodeYT, line 38: Do we need this memset here? If we are setting all / most of the members, we could wasting CPU time zeroing it (applies to the other functions in commands.cpp too)

ElectrodeYT (approved):
LGTM

drivers/gfx/virtio/src/commands.cpp
ArsenArsen: on top of that, this is default constructed

Dennisbonke (approved):
LGTM after the memsets are removed.

drivers/gfx/virtio/src/commands.cpp
qookei: > on top of that, this is default constructed

AFAICT this is an aggregate of primitive types (and aggregates), so default construction means uninitialized.

qookei (approved):

PR #553, opened 2023-08-14
drivers/libblockfs: Implement ext2 double block allocation

Qwinci:
Implement double indirect block allocation. This should be checked by someone because I am not completely sure that its correct and it could probably be improved in some ways.


ElectrodeYT (approved):
LGTM

qookei (approved):
Seems fine to me. Perhaps we should refactor the handling of indirect blocks (currently the logic is pretty much copy pasted), but that's potentially out of scope for this PR.

Dennisbonke (approved):
LGTM

PR #554, opened 2023-08-14
posix/un-socket: Add a SpecialLink link

Qwinci:
Add a SpecialLink link to unix sockets, because it allows stat to work on them.


no92 (approved):
LGTM

PR #556, opened 2023-08-18
posix: Only launch gdbserver once

Dennisbonke:
This avoids launching 3 gdbservers if one process crashes 3 threads (firefox and chromium I'm looking at you).


no92 (approved):
LGTM

PR #557, opened 2023-08-18
posix: Launch gdbserver if thread is killed by signal

Dennisbonke:


no92 (approved):
LGTM

posix/subsystem/src/process.cpp
```diff
@@ -5,12 +5,15 @@
...
+#include "gdbserver.hpp"
 #include "process.hpp"
 
 #include <protocols/posix/data.hpp>
 
+#include "debug-options.hpp"
+
 static bool logFileAttach = false;
-static bool logCleanup = false;
+static bool logCleanup_ = false;
```
Geertiebear, line 16: Why is this change necessary?

ElectrodeYT: Conflicts with `debug-options.hpp`

ikbenlike: I feel it would be clearer to instead rename it `processLogCleanup` or something along those lines. Other than that, looks fine to me.

PR #566, opened 2023-10-04
fs: Implement O_APPEND for ext2

Dennisbonke:
This PR implements handling for O_APPEND.

### DIFF
--- a/drivers/libblockfs/src/libblockfs.cpp
+++ b/drivers/libblockfs/src/libblockfs.cpp
@@ -164,6 +164,9 @@ async::result<frg::expected<protocols::fs::Error, size_t>> write(void *object, c
 	}
 
 	auto self = static_cast<ext2fs::OpenFile *>(object);
+	if(self->append) {
+		self->offset = self->inode->fileSize();
+	}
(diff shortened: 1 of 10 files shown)


drivers/libblockfs/src/libblockfs.cpp
```diff
@@ -330,10 +333,11 @@ getStats(std::shared_ptr<void> object) {
 }
 
 async::result<protocols::fs::OpenResult>
-open(std::shared_ptr<void> object) {
+open(std::shared_ptr<void> object, bool append) {
```
no92, line 336: Maybe having this be a `flags`-type parameter would have been nicer, but looking through the other flags in POSIX I don't see any others that would need to be passed down to libblockfs, so this is fine.

protocols/fs/fs.bragi
```diff
@@ -266,6 +266,8 @@ head(128):
 		tag(69) int64 pgid;
 
 		tag(84) int32 seals;
+
+		tag(85) int32 append;
```
no92, line 270: Minor, but could also be of type `byte`.

no92 (commented):
Minor comments, but LGTM overall.

PR #567, opened 2023-10-04
thor: Use the mmap constants from mlibc

Dennisbonke:
This will help us eventually moving over to the linux abi, if we desire to do so. It also improves readability.

### DIFF
--- a/kernel/thor/generic/service.cpp
+++ b/kernel/thor/generic/service.cpp
@@ -12,6 +12,10 @@
 #include <protocols/posix/data.hpp>
 #include <protocols/posix/supercalls.hpp>
 
+#define __MLIBC_ABI_ONLY
+#include <sys/mman.h>
+#undef __MLIBC_ABI_ONLY
+
 namespace thor {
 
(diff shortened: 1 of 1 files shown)


Dennisbonke:
Apparently I can't use mlibc headers on ci? @avdgrinten @ArsenArsen might know more here?

kernel/thor/generic/service.cpp
```diff
@@ -584,8 +588,8 @@ namespace posix {
 				}
 
 				smarter::shared_ptr<MemoryView> fileMemory;
-				if(req->flags() & 8) { // MAP_ANONYMOUS.
-					if(req->flags() & 1) { // MAP_PRIVATE.
+				if(req->flags() & MAP_ANONYMOUS) { // MAP_ANONYMOUS.
```
no92, line 591: The comment can be removed then, same goes for the others below.

no92 (commented):
This would need a bootstrap change too, as `managarm-kernel` currently does not depend on mlibc.

PR #570, opened 2023-10-07
posix, thor: handle divide by zero

ElectrodeYT:


posix/subsystem/src/observations.cpp
```diff
@@ -591,6 +591,27 @@ async::result<void> observeThread(std::shared_ptr<Process> self,
 			if(killed)
 				break;
 			HEL_CHECK(helResume(thread.getHandle()));
+		}else if(observe.observation() == kHelObserveDivByZero) {
+			printf("\e[31mposix: Divide by zero in process %s\n", self->path().c_str());
+			dumpRegisters(self);
+			printf("\e[39m");
+			fflush(stdout);
+
+			if(debugFaults) {
```
no92, line 600: I'm not sure if this is the best way to place this; maybe gdb should only be launched when the process got killed?

no92 (commented):
LGTM, one question though.

posix/subsystem/src/observations.cpp
Dennisbonke: Agreed.

ElectrodeYT: All other faults use the same code; while that would be (imo) a good change, it should be done to all faults at the same time.

I would personally prefer to get this merged in first as there may be reasons not known to me that might cause that to not be a good idea, and blocking this on it would probably not be the best cause of action

Dennisbonke: As stated on discord, I'm 99% sure that v8 / jsc catch SIGFPE to do some cursed stuff. While I agree with the fact that it should be changed everywhere, I do propose we do that now. I'm fine with you adding that to this PR if you want,

Dennisbonke (approved):
LGTM

no92 (approved):
LGTM

PR #575, opened 2023-10-11
[WIP] thor: improve scheduler heuristics gathering

ElectrodeYT:


hel/include/hel.h
```diff
@@ -1078,6 +1087,27 @@ HEL_C_LINKAGE HelError helEnableFullIo();
...
+//! @param[out] global
+//!     Summarised information from all schedulers.
+//!     This is optional; if this information is not required, pass a nullptr.
+HEL_C_LINKAGE HelError helGetSchedulerInformation(size_t bufferSize, HelSchedulerStats *cpuBuffer, HelSchedulerStats *global);
+
+
+//! Get general system information.
+//! @param[out] nanosSinceBoot
+//!     Number of nanoseconds since time keeping began.
+HEL_C_LINKAGE HelError helGetSystemInformation(uint64_t *nanosSinceBoot);
```
qookei, line 1109: This is a duplicate of `helGetClock`

kernel/thor/generic/schedule.cpp
```diff
@@ -201,6 +203,39 @@ void Scheduler::update() {
 	if(n)
 		_systemProgress += deltaTime * fixedInverse(n);
 
+	switch(_current->type()) {
```
qookei, line 206: Maybe instead of the 2 bools store the previous ScheduleType itself?

qookei (commented):
Haven't yet had a proper look at the migration logic yet. Also style issues (left-sided pointers, snake_case names)

hel/include/hel.h
ElectrodeYT: Didn't notice, thanks!
That being said, i plan on exposing other things here related to system (debug and performance) info, so I will keep it for now.

qookei: Do drop the nanos since boot from it though, no need to expose it twice

ArsenArsen: are you sure it should all go over one syscall?  and on top of that, via out parameters?

ElectrodeYT: no, but i hacked all of these things together between classes and while IT class, so it was basically a case of "how quickly can i get this to work"

ArsenArsen: it really shouldn't be called 'helGetSystemInformation' if it can't be arbitrarily expanded to accommodate for new system information

ElectrodeYT: Call has been removed

kernel/thor/generic/schedule.cpp
ElectrodeYT: Done

ElectrodeYT:
Scope of the PR has been reduced; I found that the CPU migration that I implemented was not fully reliable. The /proc/stat code, however, is functional, and ready for review.

I may revisit the idea of CPU migration in the future, however this PR is not it.

kernel/thor/generic/hel.cpp
```diff
@@ -3403,3 +3403,35 @@ HelError helQueryRegisterInfo(int set, HelRegisterInfo *info) {
 	return kHelErrNone;
 }
 
+HelError helGetSchedulerInformation(size_t bufferSize, HelSchedulerStats* cpuBuffer, HelSchedulerStats* global) {
+	if(cpuBuffer != nullptr && bufferSize < (size_t)getCpuCount()) {
+		infoLogger() << "thor: helGetSchedulerInformation: buffer too small, core count is " << getCpuCount() << ", vector size is " << bufferSize << frg::endlog;
+		return kHelErrBufferTooSmall;
+	}
+
+	if (cpuBuffer == nullptr && global == nullptr)
```
no92, line 3412: I wonder whether it would be useful to also specify the `nonnull` attribute for this function in `hel.h`.

no92 (approved):
LGTM

PR #578, opened 2023-10-17
thor: Don't launch gdbserver more than once if servers die

Dennisbonke:
We've already done this for posix, let's do it for the kernel too.


ElectrodeYT (approved):
LGTM

qookei (approved):

qookei (approved):

PR #581, opened 2023-10-20
thor: Add helGetCurrentCpu syscall

Qwinci:


no92 (approved):
LGTM

hel/include/hel.h
```diff
@@ -929,6 +930,8 @@ HEL_C_LINKAGE HelError helWriteGsBase(void *pointer);
 
 HEL_C_LINKAGE HelError helReadGsBase(void **pointer);
 
+HEL_C_LINKAGE HelError helGetCurrentCpu(int *index);
```
qookei, line 933: Could you document this?

kernel/thor/generic/main.cpp
```diff
@@ -876,6 +876,9 @@ void handleSyscall(SyscallImageAccessor image) {
 	case kHelCallSetAffinity: {
 		*image.error() = helSetAffinity((HelHandle)arg0, (uint8_t *)arg1, (size_t)arg2);
 	} break;
+	case kHelCallGetCurrentCpu: {
+		*image.error() = helGetCurrentCpu((int*)arg0);
```
ElectrodeYT, line 880: This should not use output paramters; this should use `*image.out{}`

ElectrodeYT (changes requested):
One thing

kernel/thor/generic/main.cpp
Qwinci: I changed it to use out0. And I assume helSyscall0_1 is the correct one to use in hel-syscalls.h for that?

ElectrodeYT: Yep, it should be. However, you should still be returning a HelError in `*image.error()`. Returning `kHelErrNone` unconditionally should suffice.

Dennisbonke (approved):
LGTM

PR #582, opened 2023-10-20
Add GetMemoryInformation request to kerncfg/posix

Qwinci:


avdgrinten:
Getting informative data from the kernel is better done by adding a `kerncfg` request. That way, we don't need to add a lot of additional syscalls, and it's easier to do access control and versioning.

Qwinci:
I moved it to kerncfg like you suggested.

qookei (commented):
Seems fine to me, although I'm unsure if this is the right time to start using separate response messages instead of adding fields to the `SvrResponse` message.

no92 (approved):
LGTM

ElectrodeYT (approved):
LGTM

Dennisbonke (approved):
LGTM

PR #584, opened 2023-10-22
[EXTREME wip] rework netserver <-> nic comms, split out NICs from netserver

ElectrodeYT:
This (in theory) has the following advantages:

- A error in a NIC driver cannot bring the rest of netserver down with it
- netserver can be started regardless of whether a NIC is present
  - Needed for loopback devices
- Userspace could easily create virtual NICs
  - Needed for eg. Docker, would be useful for virtual machines
  - Does bring up potential security implications though
- Would allow NIC drivers to be built out-of-tree
  - Not really a benefit, but still possible
- If we get more NIC drivers, would reduce the size of netserver
  - Also means that its technically possible to not have to install all network drivers
- Communication between the NIC and userspace directly might be possible?
  - I do not really know if this would make sense, since if routes change during communication they would not apply to any currently open ports
- Also more microkernel-like :meme:

This PR also aims to make transition for the two NIC PRs simple, by implementing `nic_core::Nic`, which deals with all of the communication between the NIC and netserver. 

The way this PR currently implements this rework has a few disadvantages, which (for simplicity of review and to keep the amount of changes at once lower) will be addressed in a future:

- Adds a couple more `memcpy`s in the loop
  - Considering that we likely have other performance problems and that the data is copied several times anyway this likely does not have that big of a performance hit, but I have not tested this yet
  - Also its networking, throughput is likely limited by the link speed anyway
  - Should be pretty simple to avoid, by simply using shared memory for creating the package and sending it to the NIC driver
  - This also means that dma memory is not used for sending packets
    - Should also be somewhat simple to workaround with shared memory, however
 - The packet IDs for the net protocol are offset by 50, since the same `serve()` loop that handles the FS requests to netserver also handles the net protocol requests needed to register a NIC
   - Im honestly not sure if this is needed, however I also could not be bothered to check, so I did it like this to make sure
   - Could be solved by publishing another lane onto mbus (e.g. `netserver-nics` or `netserver-net`?) with a different `server()` function


servers/netserver/src/main.cpp
```diff
@@ -19,83 +19,64 @@
...
 		// 10.0.2.0/24
-		ip4Router().addRoute({ { 0x0a000200, 24 }, device });
+		ip4Router().addRoute({ { 0x0a000200, 24 }, nic });
 		// inet 10.10.2.15/24
-		ip4().setLink({ 0x0a0a020f, 24 }, device);
+		ip4().setLink({ 0x0a0a020f, 24 }, nic);
 	}
-	baseDeviceMap.insert({base_entity.getId(), device});
-	nic::runDevice(device);
+	registeredNics.insert(registeredNics.end(), nic);
```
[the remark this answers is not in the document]
ElectrodeYT: Yeah, don't know what I was thinking here; changed to `push_back`.

core/nic/include/core/nic/buffer.hpp
```diff
@@ -0,0 +1,108 @@
...
+		assert(_data <= ((char*)(_buffer->data) + _buffer->size));
+	}
+
+	void *_data;
+	size_t _size;
+
+	// This keeps the memory mapped until the last buffer_view is destroyed.
+	// We store the offset so that we can quickly copy the handle from the buffer_owner.
+	size_t _offset;
+	std::shared_ptr<buffer_owner> _buffer;
```
avdgrinten, line 105: Why are these semantics needed? This class _almost_ re-implements `arch::dma_buffer_view` except that this is not really a view but rather a buffer with shared ownership semantics.

core/nic/include/core/nic/buffer.hpp
```diff
@@ -0,0 +1,108 @@
...
+		assert(offset <= _size);
+		return buffer_view{_offset + offset, _size - offset, _buffer};
+	}
+
+	HelHandle copyHandle(size_t &allocatedSize) const {
+		HelHandle new_handle;
+		allocatedSize = (_size + 0x1000 - 1) & ~(0x1000 - 1);
+		HEL_CHECK(helCopyOnWrite(_buffer->getHandle(), _offset, _size, &new_handle));
+		return new_handle;
+	}
```
avdgrinten, line 90: Why is this useful?

avdgrinten (commented):
As mentioned on Discord, it's preferable to continue to represent NICs as libraries, with a generic server stub to extract them into their own processes.

That way, NICs can both be compiled into netserver _and_ into standalone processes.

core/nic/include/core/nic/buffer.hpp
ElectrodeYT: There were other reasons that are no longer valid (I was going to pass around descriptors instead of buffers in IPC, however this became quite complicated so I decided to leave it for a later PR). This has been undone locally on my end.

core/nic/include/core/nic/buffer.hpp
ElectrodeYT: See above comment

PR #585, opened 2023-10-22
Implement USB sysfs support

no92:
Conveniently, this also enables USB support for udev, no recompile needed.

Aside from that, device numbering (for e.g. `/dev/ttyS0`, `/dev/ttyS1` etc.) is now handled by posix. Previously, drivers hardcoded a number, with obvious implications.

### DIFF
--- a/drivers/usb/hcds/uhci/src/main.cpp
+++ b/drivers/usb/hcds/uhci/src/main.cpp
@@ -146,11 +146,11 @@ async::result<frg::expected<proto::UsbError, size_t>> EndpointState::transfer(pr
 // Controller.
 // ----------------------------------------------------------------------------
 
-Controller::Controller(protocols::hw::Device hw_device, uintptr_t base,
+Controller::Controller(protocols::hw::Device hw_device, mbus::Entity entity, uintptr_t base,
 		arch::io_space space, helix::UniqueIrq irq)
 : _hwDevice{std::move(hw_device)}, _ioBase{base}, _ioSpace{space}, _irq{std::move(irq)},
 		_lastFrame{0}, _frameCounter{0},
-		_enumerator{this} {
+		_entity{std::move(entity)}, _enumerator{this} {
 	for(int i = 1; i < 128; i++) {
 		_addressStack.push(i);
 	}
@@ -485,6 +485,11 @@ async::result<void> Controller::enumerateDevice(std::shared_ptr<proto::Hub> pare
 		_enumerator.observeHub(std::move(hub));
 	}
 
+	char name[3];
+	snprintf(name, 3, "%.2x", address);
+
+	std::string mbps = protocols::usb::getSpeedMbps(speed);
+
 	mbus::Properties mbus_desc {
 		{ "usb.type", mbus::StringItem{"device"}},
 		{ "usb.vendor", mbus::StringItem{vendor}},
@@ -492,13 +497,14 @@ async::result<void> Controller::enumerateDevice(std::shared_ptr<proto::Hub> pare
 		{ "usb.class", mbus::StringItem{class_code}},
 		{ "usb.subclass", mbus::StringItem{sub_class}},
 		{ "usb.protocol", mbus::StringItem{protocol}},
-		{ "usb.release", mbus::StringItem{release}}
+		{ "usb.release", mbus::StringItem{release}},
+		{ "usb.hub_port", mbus::StringItem{name}},
+		{ "usb.bus", mbus::StringItem{std::to_string(_entity.getId())}},
+		{ "usb.speed", mbus::StringItem{mbps}},
+		{ "unix.subsystem", mbus::StringItem{"usb"}},
 	};
 
 	auto root = co_await mbus::Instance::global().getRoot();
-
-	char name[3];
-	sprintf(name, "%.2x", address);
 
 	auto handler = mbus::ObjectHandler{}
 	.withBind([=] () -> async::result<helix::UniqueDescriptor> {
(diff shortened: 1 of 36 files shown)


protocols/usb/src/api.cpp
```diff
@@ -1,30 +1,40 @@
 
 #include "protocols/usb/api.hpp"
 
+namespace protocols::usb {
```
qookei, line 4: No need to prefix all the names below with `protocols::usb::` if this is here :^)

protocols/usb/src/api.cpp
```diff
@@ -1,30 +1,40 @@
...
+	return _state->deviceDescriptor();
+}
+
+async::result<frg::expected<UsbError, std::string>> protocols::usb::Device::configurationDescriptor() const {
 	return _state->configurationDescriptor();
 }
 
-async::result<frg::expected<UsbError, Configuration>> Device::useConfiguration(int number) const {
+async::result<frg::expected<UsbError, uint8_t>> protocols::usb::Device::currentConfigurationValue() const {
+	return _state->currentConfigurationValue();
```
qookei, line 30: Why not implement this by performing the transfer here (via `Device::transfer(ControlTransfer)`). Then you wouldn't need to add a new request for this.

Same applies to `deviceDescriptor`, although that's an analogue for `configurationDescriptor` so that's probably fine?

posix/subsystem/src/subsystem/usb/root-hub.hpp
```diff
@@ -0,0 +1,167 @@
+#pragma once
+
+#include <array>
+#include <cstdint>
+
+namespace usb_subsystem::root_hub {
+
+constexpr std::array<uint8_t, 18> desc_usb_3_0 = {
```
qookei, line 8: camelCase names for these?

drivers/usb/hcds/ehci/src/spec.hpp
```diff
@@ -8,127 +8,127 @@
...
 // QueueHead
 //-------------------------------------------------
 
 namespace qh_horizontal {
-	arch::field<uint32_t, bool> terminate(0, 1);
-	arch::field<uint32_t, uint8_t> typeSelect(1, 2);
-	arch::field<uint32_t, uint32_t> horizontalPtr(0, 32);
+	constexpr arch::field<uint32_t, bool> terminate(0, 1);
+	constexpr arch::field<uint32_t, uint8_t> typeSelect(1, 2);
+	const arch::field<uint32_t, uint32_t> horizontalPtr(0, 32);
```
qookei, line 77: Why is this (and some other fields below) `const` and not `constexpr`?

drivers/usb/hcds/uhci/src/main.cpp
```diff
@@ -416,89 +429,117 @@ async::result<void> Controller::enumerateDevice(std::shared_ptr<Hub> parentHub,
...
-		auto hub = (co_await createHubFromDevice(parentHub, Device{std::move(state)}, port)).unwrap();
+		auto hub = (co_await createHubFromDevice(parentHub, protocols::usb::Device{std::move(state)}, port)).unwrap();
 		_enumerator.observeHub(std::move(hub));
 	}
 
+	char name[3];
+	snprintf(name, 3, "%.2x", address);
+
+	std::string mbps;
+	switch(speed) {
```
qookei, line 498: Put this in `protocols/usb/usb.hpp` or thereabout?

drivers/usb/hcds/xhci/src/xhci.hpp
```diff
@@ -138,48 +140,55 @@ struct Controller final : BaseController {
...
-		async::result<PortState> pollState(int port) override;
-		async::result<frg::expected<UsbError, DeviceSpeed>> issueReset(int port) override;
+		async::result<protocols::usb::PortState> pollState(int port) override;
+		async::result<frg::expected<protocols::usb::UsbError, protocols::usb::DeviceSpeed>> issueReset(int port) override;
 
 		SupportedProtocol *protocol() {
 			return _proto;
 		}
 
+		mbus::_detail::EntityId entityId() {
```
qookei, line 157: Accessing the detail namespace here? If it's not available otherwise maybe use `auto`.

drivers/usb/hcds/xhci/src/main.cpp
```diff
@@ -388,47 +403,78 @@ async::result<void> Controller::enumerateDevice(std::shared_ptr<Hub> parentHub,
...
 
 		_enumerator.observeHub(std::move(hub));
 	}
 
+	char name[3];
+	snprintf(name, 3, "%.2lx", device->slot());
+
+	std::string mbps;
+
+	switch(speed) {
```
qookei, line 432: Ditto.

qookei (commented):
I'm wondering whether maybe we should use `using namespace protocols::usb;` in the USB drivers to reduce the amount of changes, or at least add a namespace alias to make it shorter.

drivers/usb/hcds/ehci/src/spec.hpp
no92: The compiler started throwing errors and that made it work, I don't remember the details though.

no92:
Applied the suggested fixes, however xbstrap should use a new void linux repo URL which is why GHA fails.

drivers/usb/hcds/ehci/src/spec.hpp
no92: From discord, this should be changed to use `scalar_variable`.

PR #587, opened 2023-10-22
core/drm: general improvements

no92:
These miscellaneous improvements came up while integrating lil.

### DIFF
--- a/core/drm/src/core.cpp
+++ b/core/drm/src/core.cpp
@@ -200,11 +200,7 @@ drm_core::File::pollStatus(void *object) {
 	co_return protocols::fs::PollStatusResult{self->_eventSequence, s};
 }
 
-async::detached
-drm_core::File::_retirePageFlip(std::unique_ptr<drm_core::Configuration> config,
-			uint64_t cookie, uint32_t crtc_id) {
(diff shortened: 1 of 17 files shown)


core/drm/src/core.cpp
```diff
@@ -200,15 +200,13 @@ drm_core::File::pollStatus(void *object) {
 	co_return protocols::fs::PollStatusResult{self->_eventSequence, s};
 }
 
-async::detached
-drm_core::File::_retirePageFlip(std::unique_ptr<drm_core::Configuration> config,
-			uint64_t cookie, uint32_t crtc_id) {
-	co_await config->waitForCompletion();
-
+async::detached drm_core::File::_retirePageFlip(uint64_t cookie, uint32_t crtc_id) {
```
qookei, line 203: This can probably now be a regular function and not a coroutine?

qookei (commented):
Just the one suggestion, otherwise looks good to me, although I'm not too knowledgeable about DRM.

core/drm/src/core.cpp
no92: Good catch, changed to return void.

qookei (approved):

PR #589, opened 2023-10-23
[WIP] netserver: Implement some net ioctls

Qwinci:
Implement some net ioctls.

### DIFF
--- a/posix/subsystem/src/requests.cpp
+++ b/posix/subsystem/src/requests.cpp
@@ -2266,6 +2266,42 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 			auto [send_resp] = co_await helix_ng::exchangeMsgs(conversation,
 					helix_ng::sendBuffer(ser.data(), ser.size()));
 			HEL_CHECK(send_resp.error());
+		}else if(preamble.id() == managarm::posix::NetserverIoctlRequest::message_id) {
+			auto req = bragi::parse_head_only<managarm::posix::NetserverIoctlRequest>(recv_head);
+
+			if(!req) {
+				std::cout << "posix: Rejecting request due to decoding failure" << std::endl;
+				break;
+			}
+
+			if(logRequests)
+				std::cout << "posix: Netserver ioctl" << std::endl;
+
+			auto req_data = req->SerializeAsString();
+			char buffer[128];
+
+			auto [offer, send_req, recv_resp] = co_await helix_ng::exchangeMsgs(
+				co_await net::getNetLane(),
+				helix_ng::offer(
+					helix_ng::sendBuffer(req_data.data(), req_data.size()),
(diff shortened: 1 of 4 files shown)


Dennisbonke:
Holding on #584 per request.

servers/netserver/src/main.cpp
```diff
@@ -173,6 +177,84 @@ async::detached serve(helix::UniqueLane lane) {
...
 			}
+		}else if(preamble.id() == managarm::posix::NetserverIoctlRequest::message_id) {
+			managarm::posix::NetserverIoctlRequest req;
+			req.ParseFromArray(recv_req.data(), recv_req.length());
+			recv_req.reset();
+
+			managarm::posix::NetserverIoctlReply resp;
+			resp.set_error(managarm::posix::Errors::ILLEGAL_ARGUMENTS);
+			if(req.command() == SIOCGIFCONF) {
+				auto links = nic::Link::getLinks();
```
ElectrodeYT, line 188: This looks like it gets used in basically every ioctl call; maybe it should be put outside of the if statement?

ElectrodeYT (approved):
LGTM

Looking at this, and comparing it to my WIP netserver changes, they look pretty easy to rebase in either direction.
Since I may or may not have time to finish the netserver changes next week (I really need to get some _Diplomarbeit_ stuff done lol), i am OK with this getting merged in before my PR.

If this PR gets changes which impact the _Link_ struct in netserver, I would like to request a ping, since that will get replaced by a very similar but still slightly different structure in a new core library (specifically `core_nic::Nic`)

servers/netserver/src/main.cpp
Qwinci: Yeah, I moved it outside the ifs.

posix/subsystem/src/requests.cpp
```diff
@@ -2266,6 +2266,42 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 			auto [send_resp] = co_await helix_ng::exchangeMsgs(conversation,
 					helix_ng::sendBuffer(ser.data(), ser.size()));
 			HEL_CHECK(send_resp.error());
+		}else if(preamble.id() == managarm::posix::NetserverIoctlRequest::message_id) {
```
ArsenArsen, line 2269: hmm, why is this forward needed?  doesn't ioctl operate on the FD lane?

Qwinci: > hmm, why is this forward needed? doesn't ioctl operate on the FD lane?

Maybe it could indeed work using the ioctl file op, assuming that its possible to have ioctl thats independent of the higher level protocol (tdp, udp etc) on ipv4 socket.

Qwinci: Because all these ioctls need is a socket with a specific family (in this case AF_INET)

PR #590, opened 2023-10-23
Sysfs seeking improvements

Dennisbonke:
`lsblk` was crashing posix on seeking on a sysfs directory. Return EINVAL for now and add some error handling while we're at it.

mlibc counterpart to be provided shortly.

### DIFF
--- a/posix/subsystem/src/file.cpp
+++ b/posix/subsystem/src/file.cpp
@@ -24,8 +24,11 @@ File::ptSeekAbs(void *object, int64_t offset) {
 	auto self = static_cast<File *>(object);
 	auto result = co_await self->seek(offset, VfsSeek::absolute);
 	if(!result) {
-		assert(result.error() == Error::seekOnPipe);
-		co_return protocols::fs::Error::seekOnPipe;
+		if(result.error() == Error::seekOnPipe)
+			co_return protocols::fs::Error::seekOnPipe;
+		else if(result.error() == Error::illegalArguments)
+			co_return protocols::fs::Error::illegalArguments;
+		assert(!"Unexpected error!");
 	}else{
 		co_return result.value();
 	}
@@ -36,8 +39,11 @@ File::ptSeekRel(void *object, int64_t offset) {
 	auto self = static_cast<File *>(object);
 	auto result = co_await self->seek(offset, VfsSeek::relative);
 	if(!result) {
-		assert(result.error() == Error::seekOnPipe);
-		co_return protocols::fs::Error::seekOnPipe;
+		if(result.error() == Error::seekOnPipe)
+			co_return protocols::fs::Error::seekOnPipe;
+		else if(result.error() == Error::illegalArguments)
+			co_return protocols::fs::Error::illegalArguments;
+		assert(!"Unexpected error!");
 	}else{
 		co_return result.value();
 	}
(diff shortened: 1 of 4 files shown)


posix/subsystem/src/sysfs.cpp
```diff
@@ -119,6 +119,11 @@ void DirectoryFile::handleClose() {
 	_cancelServe.cancel();
 }
 
+// TODO: Verify that this is correct
+async::result<frg::expected<Error, off_t>> DirectoryFile::seek(off_t, VfsSeek) {
+	co_return Error::illegalArguments;
```
qookei, line 124: Not sure what to make of this (or how it affects `readdir` results):
```
Python 3.11.5 (main, Aug 29 2023, 02:02:00) [GCC 13.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.open('/sys/class', os.O_DIRECTORY)
3
>>> os.lseek(3, 10, os.SEEK_CUR)
10
>>> 
```

posix/subsystem/src/file.cpp
```diff
@@ -24,35 +24,47 @@ File::ptSeekAbs(void *object, int64_t offset) {
...
-		co_return protocols::fs::Error::seekOnPipe;
+		if(result.error() == Error::seekOnPipe)
+			co_return protocols::fs::Error::seekOnPipe;
+		else if(result.error() == Error::illegalArguments)
+			co_return protocols::fs::Error::illegalArguments;
+		assert(result.error() == Error::seekOnPipe || result.error() == Error::illegalArguments);
 	}else{
 		co_return result.value();
 	}
+	__builtin_unreachable();
```
qookei, line 35: `__builtin_unreachable` would've been better placed after the assert itself. Also why do we need it in the first place? I'd rewrite the assert as just `assert(!"Unexpected error");` since it's condition will always be false due to the preceding ifs.

posix/subsystem/src/file.cpp
```diff
@@ -24,35 +24,47 @@ File::ptSeekAbs(void *object, int64_t offset) {
...
```
qookei, line 51: Ditto.

posix/subsystem/src/file.cpp
```diff
@@ -24,35 +24,47 @@ File::ptSeekAbs(void *object, int64_t offset) {
...
```
qookei, line 67: Ditto.

posix/subsystem/src/file.cpp
Dennisbonke: Clang was complaining about a potential path not having a return value. I'll reword the asserts.

qookei: With the assert's condition being always false (since it's a negation of a pointer to a string literal) it should see that it can't reach below since it always enters the `noreturn` assert failure function.

posix/subsystem/src/sysfs.cpp
qookei: Understandable. I have no idea what seeking on a directory is supposed to do in relation to `readdir` et al. anyway.

posix/subsystem/src/file.cpp
Dennisbonke: clang disagrees I'm afraid.

qookei (approved):

PR #591, opened 2023-10-23
General sysfs block improvements

Dennisbonke:
This PR exposes some more attributes over sysfs for blockdevices. Most are hardcoded to a sane value (R/O is always zero), some are just a placeholder (size being 3.3G) and some are correct (major:minor number). This makes `lsblk` show `sda0`!

### DIFF
--- a/drivers/libblockfs/src/libblockfs.cpp
+++ b/drivers/libblockfs/src/libblockfs.cpp
@@ -5,6 +5,7 @@
 #include <string>
 #include <sys/epoll.h>
 #include <linux/cdrom.h>
+#include <linux/fs.h>
 
 #include <helix/ipc.hpp>
 #include <protocols/fs/server.hpp>
@@ -753,6 +754,33 @@ async::detached servePartition(helix::UniqueLane lane) {
 			);
 			HEL_CHECK(send_resp.error());
 			HEL_CHECK(push_node.error());
+		} else if(preamble.id() == managarm::fs::GenericIoctlRequest::message_id) {
+			auto req = bragi::parse_head_only<managarm::fs::GenericIoctlRequest>(recv_head);
+
+			if(!req) {
+				std::cout << "libblockfs: Rejecting request due to decoding failure" << std::endl;
+				break;
+			}
+
+			if(req->command() == BLKGETSIZE64) {
+				managarm::fs::GenericIoctlReply rsp;
+				rsp.set_error(managarm::fs::Errors::SUCCESS);
+				rsp.set_size(co_await fs->device->getSize());
+
+				auto ser = rsp.SerializeAsString();
+				auto [send_resp] = co_await helix_ng::exchangeMsgs(
+						conversation,
+						helix_ng::sendBuffer(ser.data(), ser.size())
+				);
+				HEL_CHECK(send_resp.error());
+			} else {
+				std::cout << "\e[31m" "libblockfs: Unknown ioctl() message with ID "
+						<< req->command() << "\e[39m" << std::endl;
+
+				auto [dismiss] = co_await helix_ng::exchangeMsgs(
+					conversation, helix_ng::dismiss());
+				HEL_CHECK(dismiss.error());
+			}
 		}else{
 			throw std::runtime_error("Unexpected request type " + std::to_string((int)req.req_type()));
 		}
--- a/posix/subsystem/src/subsystem/block.cpp
+++ b/posix/subsystem/src/subsystem/block.cpp
@@ -1,6 +1,7 @@
 
 #include <string.h>
 #include <iostream>
+#include <linux/fs.h>
 
 #include <protocols/mbus/client.hpp>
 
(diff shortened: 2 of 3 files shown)


posix/subsystem/src/subsystem/block.cpp
```diff
@@ -51,16 +51,71 @@ struct Device final : UnixDevice, drvcore::BlockDevice {
...
+		char majorBuf[2];
+		char minorBuf[4];
+		snprintf(majorBuf, 2, "%d", major);
+		snprintf(minorBuf, 4, "%d", minor);
 		ue.set("SUBSYSTEM", "block");
+		ue.set("MAJOR", majorBuf);
+		ue.set("MINOR", minorBuf);
 	}
 
+	int major;
```
qookei, line 63: No need to add new members, this is stored by `UnixDevice`, and can be accessed with `UnixDevice::getId()` (this returns the `DeviceId`, which is a `std::pair<int, int>` so the major is `.first`, minor is `.second`).

posix/subsystem/src/subsystem/block.cpp
```diff
@@ -51,16 +51,71 @@ struct Device final : UnixDevice, drvcore::BlockDevice {
 	}
 
 	void composeUevent(drvcore::UeventProperties &ue) override {
+		char majorBuf[2];
+		char minorBuf[4];
```
qookei, line 55: Theoretically (at least on Linux) the major can be anywhere in the range [0, 4096), and the minor anywhere in the range [0, 1048576), so in this file at least the minor buffer should maybe be larger? (considering the major is always assigned to 8 for these devices).

Perhaps we could just use `std::to_string` instead of `snprintf` since we're not doing any fancy formatting?

posix/subsystem/src/subsystem/block.cpp
```diff
@@ -51,16 +51,71 @@ struct Device final : UnixDevice, drvcore::BlockDevice {
...
+	char buffer[5]; // The format is 0:0\n\0.
+	auto device = static_cast<Device *>(object);
+	sprintf(buffer, "%d:%d\n", device->major, device->minor);
+	co_return std::string{buffer};
+}
+
+async::result<std::string> SizeAttribute::show(sysfs::Object *object) {
+	char buffer[9]; // The format is TBD.
+	// Hardcoded value, this turns out to be 3.3G according to lsblk.
+	sprintf(buffer, "6942069\n");
```
qookei, line 115: Again `std::to_string`, although this is hardcoded anyway.

drivers/libblockfs/src/libblockfs.cpp
```diff
@@ -753,6 +754,28 @@ async::detached servePartition(helix::UniqueLane lane) {
 			);
 			HEL_CHECK(send_resp.error());
 			HEL_CHECK(push_node.error());
+		} else if(preamble.id() == managarm::fs::GenericIoctlRequest::message_id) {
+			auto req = bragi::parse_head_only<managarm::fs::GenericIoctlRequest>(recv_head);
+			assert(req);
```
ElectrodeYT, line 759: This should be:
```cpp
			if (!req) {
				std::cout << "libblockfs: Rejecting request due to decoding failure" << std::endl;
				break;
			}
```

posix/subsystem/src/subsystem/block.cpp
```diff
@@ -51,16 +56,66 @@ struct Device final : UnixDevice, drvcore::BlockDevice {
...
+	// Hardcode to zero as we don't support ro mounts yet.
+	sprintf(buffer, "0\n");
+	co_return std::string{buffer};
+}
+
+async::result<std::string> DevAttribute::show(sysfs::Object *object) {
+	char buffer[5]; // The format is 0:0\n\0.
+	auto device = static_cast<Device *>(object);
+	auto dev = device->getId();
+	sprintf(buffer, "%d:%d\n", dev.first, dev.second);
```
ElectrodeYT, line 110: This doesn't seem safe, are we sure that using a static sized buffer (especially one of this size) is a good idea here?

ElectrodeYT (changes requested):

ElectrodeYT (approved):
LGTM

drivers/libblockfs/src/libblockfs.cpp
```diff
@@ -753,6 +754,33 @@ async::detached servePartition(helix::UniqueLane lane) {
...
+				rsp.set_size(co_await fs->device->getSize());
+
+				auto ser = rsp.SerializeAsString();
+				auto [send_resp] = co_await helix_ng::exchangeMsgs(
+						conversation,
+						helix_ng::sendBuffer(ser.data(), ser.size())
+				);
+				HEL_CHECK(send_resp.error());
+			} else {
+				std::cout << "\e[31m" "core/drm: Unknown ioctl() message with ID "
```
Dennisbonke, line 777: This isn't core/drm, this is libblockfs.

Dennisbonke (commented):
Found one issue.

drivers/libblockfs/src/libblockfs.cpp
no92: Fixed.

qookei (approved):

PR #593, opened 2023-10-28
posix: Create /proc/filesystems

Dennisbonke:


posix/subsystem/src/procfs.cpp
```diff
@@ -358,6 +359,25 @@ async::result<void> UptimeNode::store(std::string) {
 	co_return;
 }
 
+async::result<std::string> FilesystemsNode::show() {
+	// See man 5 proc for more details.
+	// Based on the man page from Linux man-pages 6.01, updated on 2022-10-09.
+	std::stringstream stream;
```
ElectrodeYT, line 365: I'm OK with this being done "manually" at the moment, but maybe we should log it? This is a quite big hack, and if there are reasons we can not do it properly at the moment, we would want to fix this probably the moment we can

ElectrodeYT (approved):
LGTM

posix/subsystem/src/procfs.cpp
Dennisbonke: Fixed

PR #594, opened 2023-10-28
posix: Create /proc/fs/ext2/sda0

Dennisbonke:


posix/subsystem/src/procfs.cpp
```diff
@@ -235,6 +235,12 @@ std::shared_ptr<Link> DirectoryNode::createRootDirectory() {
 
 	the_node->directMkregular("uptime", std::make_shared<UptimeNode>());
 
+	auto fsLink = the_node->directMkdir("fs");
+	auto fs = std::static_pointer_cast<DirectoryNode>(fsLink->getTarget());
+	auto ext2Link = fs->directMkdir("ext2");
+	auto ext2 = std::static_pointer_cast<DirectoryNode>(ext2Link->getTarget());
+	ext2->directMkdir("sda0");
```
ElectrodeYT, line 242: This is a massive hack, we should have a `// TODO` here (or a debug log, but this only gets called at the very beginning so probably not needed)

ElectrodeYT (approved):
LGTM

PR #595, opened 2023-10-28
Tcp additions

Qwinci:
Implement sockname/peername and FIONREAD ioctl.

### DIFF
--- a/servers/netserver/src/ip/tcp4.cpp
+++ b/servers/netserver/src/ip/tcp4.cpp
@@ -14,6 +14,8 @@
 #include <arpa/inet.h>
 #include <netinet/ip.h>
 
+#include <bragi/helpers-std.hpp>
+
 #include "checksum.hpp"
 #include "ip4.hpp"
 #include "tcp4.hpp"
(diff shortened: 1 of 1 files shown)


ElectrodeYT (approved):
LGTM

servers/netserver/src/ip/tcp4.cpp
```diff
@@ -214,8 +217,10 @@ struct Tcp4Socket {
 			co_return protocols::fs::Error::accessDenied;
 		}
 
-		if (!ip4().hasIp(bindEp.ipAddress))
+		if (bindEp.ipAddress != INADDR_ANY && !ip4().hasIp(bindEp.ipAddress)) {
+			std::cout << "netserver: IP address " << std::setw(8) << std::hex << bindEp.ipAddress << " is not available" << std::endl;
```
no92, line 221: This should probably revert back to decimal with `std::dec`.

no92 (approved):
LGTM with one minor nitpick.

servers/netserver/src/ip/tcp4.cpp
Qwinci: I added that.

Dennisbonke (approved):
LGTM

PR #596, opened 2023-10-29
More proc additions

Dennisbonke:
This time I decided to tackle some low hanging fruits in `/proc/sys/kernel`, so here are `ostype` (Managarm), `osrelease` (0.0.1) and `arch` (set to either `x86_64` or `AArch64` depending on architecture).

### DIFF
--- a/posix/subsystem/src/procfs.cpp
+++ b/posix/subsystem/src/procfs.cpp
@@ -240,6 +240,8 @@ std::shared_ptr<Link> DirectoryNode::createRootDirectory() {
 	auto kernelLink = sys->directMkdir("kernel");
 	auto kernel = std::static_pointer_cast<DirectoryNode>(kernelLink->getTarget());
 
+	kernel->directMkregular("ostype", std::make_shared<OstypeNode>());
+
 	return link;
 }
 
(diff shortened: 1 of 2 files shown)


posix/subsystem/src/procfs.cpp
```diff
@@ -358,6 +367,54 @@ async::result<void> UptimeNode::store(std::string) {
...
+async::result<std::string> ArchNode::show() {
+	// See man 5 proc for more details.
+	// Based on the man page from Linux man-pages 6.01, updated on 2022-10-09.
+	std::stringstream stream;
+#if defined(__x86_64__)
+	stream << "x86_64\n";
+#elif defined(__aarch64__)
+	stream << "AArch64\n";
+#else
+	#error "Unknown architecture"
```
ElectrodeYT, line 407: Maybe we could add a riscv entry here? We dont support it yet, but there is no harm in adding it now

`#if defined(__riscv) && __riscv_xlen == 64`

posix/subsystem/src/procfs.cpp
```diff
@@ -358,6 +367,54 @@ async::result<void> UptimeNode::store(std::string) {
...
+	// TODO: proper error reporting.
+	std::cout << "posix: Can't store to a /proc/sys/kernel/ostype file" << std::endl;
+	co_return;
+}
+
+async::result<std::string> OsreleaseNode::show() {
+	// See man 5 proc for more details.
+	// Based on the man page from Linux man-pages 6.01, updated on 2022-10-09.
+	std::stringstream stream;
+	stream << "0.0.1\n";
```
ElectrodeYT, line 388: I'm not well versed in this, is there a standard response for a git or rolling version, or do we have to report a "real" version here?

ElectrodeYT (approved):
LGTM, but a couple questions

posix/subsystem/src/procfs.cpp
Dennisbonke: We don't have a version really, so placeholder I guess

posix/subsystem/src/procfs.cpp
Dennisbonke: Noted, will add

Dennisbonke: Fixed

posix/subsystem/src/procfs.cpp
Dennisbonke: Fixed

no92 (approved):
LGTM

PR #600, opened 2023-11-12
CP2102 driver + termios fixes

no92:
Depends on managarm/mlibc#932.

### DIFF
--- a/core/include/core/tty.hpp
+++ b/core/include/core/tty.hpp
@@ -1,0 +1,50 @@
+#pragma once
+
+#include <stddef.h>
+#include <termios.h>
+
+inline size_t ttyConvertSpeed(speed_t baud) {
+	switch(baud) {
+		case B0: return 0;
+		case B50: return 50;
+		case B75: return 75;
+		case B110: return 110;
+		case B134: return 134;
+		case B150: return 150;
+		case B200: return 200;
+		case B300: return 300;
+		case B600: return 600;
+		case B1200: return 1200;
+		case B1800: return 1800;
+		case B2400: return 2400;
+		case B4800: return 4800;
+		case B9600: return 9600;
+		case B19200: return 19200;
+		case B38400: return 38400;
+		case B57600: return 57600;
+		case B115200: return 115200;
+		case B230400: return 230400;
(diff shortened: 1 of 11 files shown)


no92:
Undrafted as it is ready for review now.

core/include/core/tty.hpp
```diff
@@ -0,0 +1,50 @@
+#pragma once
+
+#include <stddef.h>
+#include <termios.h>
+
+inline size_t tty_convert_speed(speed_t baud) {
```
qookei, line 6: Use camelCase for these functions?

drivers/usb/devices/serial/src/controller.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+}
+
+struct Controller {
+	Controller(protocols::usb::Device hw);
+
+	void setupWeakPtr(smarter::weak_ptr<Controller> self);
+
+	smarter::shared_ptr<Controller> sharedModeObject() {
+		return self_.lock();
```
qookei, line 31: If you used regular `std::shared_ptr` (other drivers, like the XHCI one, use it just fine), this could be done with `std::enable_shared_from_this`

drivers/usb/devices/serial/src/main.cpp
```diff
@@ -0,0 +1,348 @@
...
+
+arch::contiguous_pool pool;
+
+WriteQueue sendRequests;
+
+}
+
+async::result<frg::expected<protocols::fs::Error, size_t>>
+write(void *object, const char *, const void *buffer, size_t length) {
+	auto self = static_cast<Controller *>(object)->sharedModeObject();
```
qookei, line 48: You shouldn't need to lock the object at this point? If you're concerned with it possibly being removed at a later point in time, why do the other functions not lock it in? (You also don't handle the case of `self_.lock()` returning `nullptr` in case it's too late)

qookei (commented):
Some lines are very long (mainly around transfers), please break them up.

Also, no read support?

no92:
Addressed the comments raised.

qookei (approved):

PR #605, opened 2023-12-16
posix: handle more errors in SOCKET request

ElectrodeYT:


hel/include/hel.h
```diff
@@ -1155,6 +1155,9 @@ extern inline __attribute__ (( always_inline )) void _helCheckFailed(HelError er
 #define HEL_SOFT_CHECK(expr) do { HelError __error = expr; if(__error != kHelErrNone) \
 		_helCheckFailed(__error, "HEL_SOFT_CHECK failed: " #expr "\n" \
 		"    In file " __FILE__ " on line " HEL_STRINGIFY(__LINE__) "\n", 0); } while(0)
+#define HEL_CHECK_OR_EIEIO(expr) do { HelError __error = expr; if(__error != kHelErrNone) \
```
no92, line 1158: While I'm not against this, it appears that this is unused.

no92 (approved):
Does this depend on managarm/mlibc#959?

hel/include/hel.h
ArsenArsen: likely a far more reasonable approach than HEL_CHECK..

PR #610, opened 2024-02-04
posix: Support USB 3.1

Qwinci:
Add support for USB 3.1. I am not completely sure if SuperSpeedPlus also has some kind of companion descriptor different from the SuperSpeed one so thats something to check (though by a quick look at the specification I didn't find anything else than SuperSpeedPlus isochronous companion descriptor which probably isn't something the hub should have).

### DIFF
--- a/posix/subsystem/src/subsystem/usb/root-hub.hpp
+++ b/posix/subsystem/src/subsystem/usb/root-hub.hpp
@@ -5,6 +5,26 @@
 
 namespace usb_subsystem::root_hub {
 
+constexpr std::array<uint8_t, 18> descUsb3_1 = {
+	0x12, /* bLength */
+	0x01, /* bDescriptorType */
+	0x10, 0x03, /* bcdUsb */
+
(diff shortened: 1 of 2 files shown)


Dennisbonke (approved):
LGTM but waiting on Leo, the sysfs master, to review.

posix/subsystem/src/subsystem/usb/root-hub.hpp
```diff
@@ -5,6 +5,26 @@
 
 namespace usb_subsystem::root_hub {
 
+constexpr std::array<uint8_t, 18> descUsb3_1 = {
+	0x12, /* bLength */
+	0x01, /* bDescriptorType */
+	0x01, 0x03, /* bcdUsb */
```
no92, line 11: This should be `0x10, 0x03`.

no92 (commented):
One minor issue found, other than than LGTM.

no92 (approved):
LGTM

PR #612, opened 2024-02-07
protocols/usb: Support large configuration descriptors

Qwinci:
Send the buffer size in the response to support larger descriptors than recvInline() can handle.


no92 (approved):
LGTM

PR #613, opened 2024-02-08
posix: Launch dbus if it is installed

Dennisbonke:


posix/init/src/stage2.cpp
```diff
@@ -156,6 +156,29 @@ int main() {
 		udev_device_unref(dev);
 	}
 
+	// Launch dbus if it's available.
+	if(!access("/usr/bin/dbus-daemon", X_OK)) {
+		mkdir("/run/dbus", 1777);
+		mkdir("/var/lib/dbus", 1777);
```
qookei, line 162: These are missing the leading 0 to be octal literals.

Dennisbonke: Fixed as requested.

no92 (approved):
LGTM but having a proper init would be great.

PR #615, opened 2024-02-09
thor: Inherit dtb cells from parent

Qwinci:
Inherit the initial address/size/irq cells from the parent if there is one.


no92 (approved):
LGTM

PR #619, opened 2024-02-15
A tiny improvement to the intel gpu driver

feature-engineer:
Turning division by 4 to >> 2 results in 3 operations instead of five: https://godbolt.org/z/6xrTejdEW

### DIFF
--- a/drivers/gfx/intel/src/main.cpp
+++ b/drivers/gfx/intel/src/main.cpp
@@ -248,7 +248,7 @@ void Controller::run() {
 	auto plane = reinterpret_cast<uint32_t *>(_memory);
 	for(size_t x = 0; x < fb.width; x++)
 		for(size_t y = 0; y < fb.height; y++)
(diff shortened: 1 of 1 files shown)


no92:
The commit message doesn't follow the [style guide](https://docs.managarm.org/handbook/contributing/commit-messages.html).

Unrelated tangent: is it worth keeping this driver around anyways? It doesn't tie into DRM, so there's little use in supporting it.

qookei:
> Unrelated tangent: is it worth keeping this driver around anyways? It doesn't tie into DRM, so there's little use in supporting it.

I'd say that it isn't, considering that AFAIK none of us have the hardware to test it with anymore, but maybe it'd be possible to eventually integrate code from this driver into lil? (once we use lil in Managarm)

no92:
EWONTFIX

ArsenArsen:
note that fixes like these go into GCC, not actual code..  please file missing optimization bugs if you notice stuff like this

PR #623, opened 2024-02-18
posix: Partially implement waitid

Dennisbonke:
Needed for elogind

### DIFF
--- a/protocols/posix/posix.bragi
+++ b/protocols/posix/posix.bragi
@@ -521,3 +521,19 @@ head(128):
 tail:
 	uint8[] mask;
 }
+
+message WaitIdRequest 91 {
+head(128):
+	uint16 idtype;
+	uint64 id;
+	int32 flags;
+}
+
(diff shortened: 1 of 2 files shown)


protocols/posix/posix.bragi
```diff
@@ -254,6 +254,10 @@ head(128):
 
 		// returned by GET_RESOURCE_USAGE
 		tag(29) uint64 ru_user_time;
+
+		// Used by waitid
```
ElectrodeYT, line 258: We should make a new response here.

protocols/posix/posix.bragi
```diff
@@ -521,3 +525,10 @@ head(128):
 tail:
 	uint8[] mask;
 }
+
+message WaitIdRequest 91 {
+head(128):
+        uint16 idtype;
```
ElectrodeYT, line 531: Indentation is broken.

PR #624, opened 2024-02-18
posix: Support partial reads from sockets

Dennisbonke:
Both for unix sockets and netlink sockets. Fixes dbus.

### DIFF
--- a/posix/subsystem/src/nl-socket.cpp
+++ b/posix/subsystem/src/nl-socket.cpp
@@ -34,6 +34,8 @@ struct Packet {
 
 	// The actual octet data that the packet consists of.
 	std::vector<char> buffer;
+
+	size_t offset = 0;
 };
 
 struct OpenFile : File {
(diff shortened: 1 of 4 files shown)


qookei:
After thinking about this some more, I don't think this is actually truncation... (but this is still something we should do)

qookei:
As in, this is correct, but leave the TODO comments, since truncation would mean that we do something else for datagram sockets.

qookei:
Can you also reword the commits? I think the proper term we're looking for is partial reads.

avdgrinten:
UNIX sockets support both stream semantics and packet semantics. We only implement stream semantics here (?). We should check which type of socket we're actually dealing with and maybe also implement packet semantics.

qookei:
> UNIX sockets support both stream semantics and packet semantics. We only implement stream semantics here (?). We should check which type of socket we're actually dealing with and maybe also implement packet semantics.

That's true, I'm just pointing out this is implementing partial reads with stream semantics, not truncation with packet semantics.

Dennisbonke:
Reworded as requested.

no92 (approved):
LGTM

no92 (approved):
LGTM

PR #632, opened 2024-03-05
kernel/eir: Merge common aarch64 init code and make virt eir pie

Qwinci:
Merge common duplicated aarch64 eir init code and make it a pie on virt platform. It would also likely be possible to do that on raspi4 but I don't really know how the bootloader handles images with base address zero (and if it would be non-zero it would need to be taken into account somehow in eirRelocate() or alternatively making sure there are no relocations at all but that requires some ugly hacks eg. around constexpr string arrays in fault.cpp).

### DIFF
--- a/kernel/eir/arch/arm/fault.S
+++ b/kernel/eir/arch/arm/fault.S
@@ -50,7 +50,8 @@ eirCommonHandler:
 	and x1, x1, #0xFFFFFFFFFFFFFFFE
 	msr sctlr_el1, x1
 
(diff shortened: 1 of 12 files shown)


qookei (approved):
Looks good. I think in the future we could merge the raspi4 and virt targets completely, having a single `eir-linux-boot-proto` PIE binary, and consulting the DTB to figure out how to get serial and a framebuffer (at least by checking the `compatible` property of the root node). This is probably better done after we refactor eir to use the init graph though.

PR #636, opened 2024-03-14
Implement shutting down the system from userspace

Dennisbonke:


no92 (approved):
LGTM

qookei (commented):
NAK, we already have a way to power off from userspace: <https://github.com/managarm/managarm/blob/a8ebe4f48172982cc4f48dcd042a3d7494c4d313/drivers/libevbackend/src/libevbackend.cpp#L36-L83>

qookei:
OFC this should still be done via posix-subsystem, so we can perform the appropriate permission checks, but adding a new system call is unnecessary.

Dennisbonke:
> NAK, we already have a way to power off from userspace:
> 
> https://github.com/managarm/managarm/blob/a8ebe4f48172982cc4f48dcd042a3d7494c4d313/drivers/libevbackend/src/libevbackend.cpp#L36-L83

That seems to do a ACPI reset unconditionally, not an ACPI move to S5? So how would I adapt that?

qookei:
IMO just add a sleep type enum and member to `PmResetRequest` to tell it what to do (suspend/shutdown/reboot) and update the libevbackend code to do a reboot.

Dennisbonke:
Closed in favor of #724

PR #640, opened 2024-03-17
posix: Support USB 3.2

Qwinci:


qookei (approved):

posix/subsystem/src/subsystem/usb/root-hub.hpp
```diff
@@ -5,6 +5,26 @@
 
 namespace usb_subsystem::root_hub {
 
+constexpr std::array<uint8_t, 18> descUsb3_2 = {
+	0x12, /* bLength */
+	0x01, /* bDescriptorType */
+	0x20, 0x03, /* bcdUsb */
```
no92, line 11: AFAICT Linux doesn't have a separate descriptor for USB 3.2; if we were to follow Linux, we would just use the 3.1 descriptor instead.

posix/subsystem/src/subsystem/usb/usb.cpp
```diff
@@ -118,6 +118,11 @@ async::detached bindController(mbus::Entity entity, mbus::Properties properties,
 					device->descriptors.insert(device->descriptors.end(), root_hub::descUsb3_1.begin(), root_hub::descUsb3_1.end());
 					device->descriptors.insert(device->descriptors.end(), root_hub::descSuperSpeed.begin(), root_hub::descSuperSpeed.end());
 					break;
+				case 0x20:
+					device->speed = "20000";
+					device->descriptors.insert(device->descriptors.end(), root_hub::descUsb3_2.begin(), root_hub::descUsb3_2.end());
```
no92, line 123: See above.

no92 (commented):
LGTM but the question is whether we follow Linux or report USB 3.2 here. I don't think that really matters for anything, but would be nice to have a policy for.

no92:
Implemented by #648.

PR #641, opened 2024-03-18
posix: Implement `/proc/[pid]/fd`.

Dennisbonke:
Massive thanks to @48cf for the help!
Also brings some drive by fixes in error handling.

### DIFF
--- a/posix/subsystem/src/epoll.cpp
+++ b/posix/subsystem/src/epoll.cpp
@@ -8,6 +8,7 @@
 #include <helix/ipc.hpp>
 #include "common.hpp"
 #include "epoll.hpp"
+#include "fs.hpp"
 
 namespace {
 
@@ -394,7 +395,7 @@ public:
 	}
 
 	OpenFile()
-	: File{StructName::get("epoll"), File::defaultPipeLikeSeek}, _currentSeq{0} { }
+	: File{StructName::get("epoll"), nullptr, SpecialLink::makeSpecialLink(VfsType::regular, 0777), File::defaultPipeLikeSeek}, _currentSeq{0} { }
 
 private:
 	helix::UniqueLane _passthrough;
--- a/posix/subsystem/src/procfs.cpp
+++ b/posix/subsystem/src/procfs.cpp
@@ -283,6 +283,7 @@ std::shared_ptr<Link> DirectoryNode::createProcDirectory(std::string name,
 	proc_dir->directMknode("exe", std::make_shared<ExeLink>(process));
 	proc_dir->directMknode("root", std::make_shared<RootLink>(process));
 	proc_dir->directMknode("cwd", std::make_shared<CwdLink>(process));
+	proc_dir->directMknode("fd", std::make_shared<FdDirectoryNode>(process));
 	proc_dir->directMkregular("maps", std::make_shared<MapNode>(process));
 	proc_dir->directMkregular("comm", std::make_shared<CommNode>(process));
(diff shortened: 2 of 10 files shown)


posix/subsystem/src/epoll.cpp
```diff
@@ -394,7 +395,7 @@ struct OpenFile : File {
 	}
 
 	OpenFile()
-	: File{StructName::get("epoll"), File::defaultPipeLikeSeek}, _currentSeq{0} { }
+	: File{StructName::get("epoll"), nullptr, SpecialLink::makeSpecialLink(VfsType::regular, 0777), File::defaultPipeLikeSeek}, _currentSeq{0} { }
```
qookei, line 398: Note: On Linux these files (epoll, inotify, eventfd, signalfd, etc) have file type 0, not regular. I don't think this really matters, as I assume nothing expects the type to be 0 anyway (coreutils' `stat` says "weird file", and that's only if the type isn't any of the ones it expects).

posix/subsystem/src/procfs.cpp
```diff
@@ -719,6 +720,71 @@ async::result<frg::expected<Error, FileStats>> CwdLink::getStats() {
...
+FdDirectoryFile::FdDirectoryFile(std::shared_ptr<MountView> mount, std::shared_ptr<FsLink> link, Process *process)
+: File{StructName::get("procfs.fddir"), std::move(mount), std::move(link)},
+		_process{process}, _iter{_process->fileContext()->fileTable().begin()} {}
+
+void FdDirectoryFile::handleClose() {
+	_cancelServe.cancel();
+}
+
+FutureMaybe<ReadEntriesResult> FdDirectoryFile::readEntries() {
+	if(_iter != _process->fileContext()->fileTable().end()) {
```
qookei, line 739: This introduces a race between the reader and target processes, which can result in a posix-subsystem crash.

The target process can change it's open files in some way that causes iterators to be invalidated between the construction of `FdDirectoryFile` and a  `readEntries` call, or between two `readEntries` calls.

Unsure of what a good solution to work around this would be. Taking a copy of the file table seems too expensive (maybe it's not too bad if it's just a map of int to pointer?), and locking the file table is not really possible either (you could block processes from opening any file descriptors by not letting go of your open `fd` directory).

posix/subsystem/src/procfs.cpp
```diff
@@ -719,6 +720,71 @@ async::result<frg::expected<Error, FileStats>> CwdLink::getStats() {
...
+	co_return File::constructHandle(std::move(file));
+}
+
+async::result<frg::expected<Error, std::shared_ptr<FsLink>>> FdDirectoryNode::getLink(std::string name) {
+	for(const auto &[fdnum, fd] : _process->fileContext()->fileTable()) {
+		if(name != std::to_string(fdnum))
+			continue;
+		co_return std::make_shared<Link>(shared_from_this(), std::move(name), fd.file->associatedLink()->getTarget());
+	}
+	co_return nullptr; // TODO: Return an error code.
```
qookei, line 785: Why not return `Error::noSuchFile` here?

qookei (changes requested):

posix/subsystem/src/procfs.cpp
Dennisbonke: Fixed

qookei (approved):

no92 (approved):
LGTM

PR #642, opened 2024-03-20
Replace LAI with uACPI and add an EC driver

d-tatianin:


kernel/common/libc.cpp
```diff
@@ -266,6 +266,23 @@ void *memset(void *dest, int byte, size_t count) {
 
 #endif // __LP64__ / !__LP64__
 
+void *memmove(void *dest, const void *src, size_t count) {
+	auto *byte_dest = reinterpret_cast<uint8_t *>(dest);
+	auto *byte_src = reinterpret_cast<const uint8_t *>(src);
+
+	if(byte_src < byte_dest) {
```
avdgrinten, line 273: This comparison is technically UB in C++ since `byte_src` and `byte_dest` are not part of the same array.

kernel/common/libc.cpp
```diff
@@ -266,6 +266,23 @@ void *memset(void *dest, int byte, size_t count) {
 
 #endif // __LP64__ / !__LP64__
 
+void *memmove(void *dest, const void *src, size_t count) {
+	auto *byte_dest = reinterpret_cast<uint8_t *>(dest);
+	auto *byte_src = reinterpret_cast<const uint8_t *>(src);
```
avdgrinten, line 271: We should use `unsigned char` here.

kernel/common/libc.cpp
```diff
@@ -266,6 +266,23 @@ void *memset(void *dest, int byte, size_t count) {
...
+
+	if(byte_src < byte_dest) {
+		byte_src += count;
+		byte_dest += count;
+
+		while (count--)
+			*--byte_dest = *--byte_src;
+	} else {
+		memcpy(dest, src, count);
+	}
```
avdgrinten, line 281: This is not correct. If the ranges overlap, `memcpy` is not guaranteed to do the right thing.

kernel/common/libc.cpp
```diff
@@ -2,6 +2,8 @@
 #include <stdint.h>
 #include <string.h>
 
+#define FRG_DONT_USE_LONG_DOUBLE
```
avdgrinten, line 5: Why is this done? Should it be added to the meson file?

kernel/thor/system/acpi/glue.cpp
```diff
@@ -72,115 +111,379 @@ void laihost_unmap(void *ptr, size_t length) {
...
+			break;
+	}
+
+	return UACPI_STATUS_OK;
 }
 
-void laihost_sleep(uint64_t) { }
+/*
+ * Events & mutexes are a stub for now, as we don't have much use for them
+ * yet nor do we have the primitives to easily implement them. FIXME later
```
avdgrinten, line 470: Why primitives are missing here?

kernel/common/libc.cpp
d-tatianin: > Why is this done? Should it be added to the meson file?

Otherwise this fails when compiling for eir
```
/var/lib/managarm-buildenv/build/system-root/usr/share/frigg/include/frg/printf.hpp:56:52: error: 'f' requires  'long double' type support, but target 'i386-unknown-managarm-kernel' does not support it
                        return reinterpret_cast<T*>(&vsp->arg_list[pos].f);
                                                                        ^
```

kernel/thor/system/acpi/glue.cpp
```diff
@@ -72,115 +113,447 @@ void laihost_unmap(void *ptr, size_t length) {
...
+		}
+		}, &getCpuData(0)->scheduler);
+
+	KernelFiber::run([]() -> coroutine<void> {
+		while (true) {
+			auto work = KernelFiber::asyncBlockCurrent(
+				acpiNotifyWorkQueue.async_get());
+			workExec(*work);
+		}
+		});
```
qookei, line 424: I don't think these need to be coroutines? (and they aren't due to no `co_await`/`co_return`, but the return type should be dropped)

kernel/thor/system/acpi/glue.cpp
```diff
@@ -72,115 +113,447 @@ void laihost_unmap(void *ptr, size_t length) {
...
+	}
+	default:
+		return UACPI_STATUS_INVALID_ARGUMENT;
+	}
+
+	return UACPI_STATUS_OK;
+}
+
+uacpi_u64 uacpi_kernel_get_ticks(void) {
+	return systemClockSource()->currentNanos() / 100;
```
qookei, line 337: Is the unit 1/10ths of a microsecond? The functions below use microseconds.

kernel/thor/system/acpi/glue.cpp
```diff
@@ -72,115 +113,447 @@ void laihost_unmap(void *ptr, size_t length) {
...
+}
+
+void uacpi_kernel_stall(uacpi_u8 usec)
+{
```
qookei, line 341: Wrong placement of `{` here and in the next function.

kernel/thor/system/acpi/glue.cpp
```diff
@@ -72,115 +113,447 @@ void laihost_unmap(void *ptr, size_t length) {
...
+) {
+	auto addr = reinterpret_cast<uacpi_io_addr>(handle);
+	return uacpi_kernel_raw_io_read(addr + offset, byte_width, value);
+}
+
+uacpi_status uacpi_kernel_io_write(
+	uacpi_handle handle, uacpi_size offset,
+	uacpi_u8 byte_width, uacpi_u64 value
+)
+{
```
qookei, line 265: Wrong placement of `{`.

qookei (commented):
Mainly some minor stylistic issues :^)

Re. mutexes and events, I think the 10ms wait in the contended case should be fine? If it turns out to be a problem we can always come back to this and refactor the code to not have it.

kernel/thor/system/acpi/glue.cpp
d-tatianin: Yeah, weirdly enough, the `Timer` operator in AML returns ticks in 100ns increments: https://uefi.org/specs/ACPI/6.5/19_ASL_Reference.html#timer-get-64-bit-timer-value

kernel/thor/system/acpi/glue.cpp
d-tatianin: Oops, sorry for these

kernel/thor/system/acpi/glue.cpp
d-tatianin: > I don't think these need to be coroutines? (and they aren't due to no `co_await`/`co_return`, but the return type should be dropped)

Are you simply suggesting I drop the trailing type, or should something else be changed as well?

kernel/thor/system/acpi/glue.cpp
qookei: Huh, okay.

kernel/thor/system/acpi/glue.cpp
qookei: Just dropping the return type is enough yeah.

qookei (approved):

PR #650, opened 2024-03-31
Implement EISDIR for read

no92:
Depends on managarm/mlibc#1047.

Previously, running `cat` on a directory returned garbage. With this change, it correctly outputs `cat: <directory>: Is a directory`.

### DIFF
--- a/protocols/fs/src/server.cpp
+++ b/protocols/fs/src/server.cpp
@@ -150,8 +150,17 @@ async::detached handlePassthrough(smarter::shared_ptr<void> file,
 
 		managarm::fs::SvrResponse resp;
 		auto error = std::get_if<Error>(&res);
-		if(error && *error == Error::wouldBlock) {
-			resp.set_error(managarm::fs::Errors::WOULD_BLOCK);
+		if(error) {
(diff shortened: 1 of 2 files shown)


protocols/fs/src/server.cpp
```diff
@@ -150,37 +150,38 @@ async::detached handlePassthrough(smarter::shared_ptr<void> file,
...
 
 			auto ser = resp.SerializeAsString();
 			auto [send_resp] = co_await helix_ng::exchangeMsgs(
 				conversation,
 				helix_ng::sendBuffer(ser.data(), ser.size())
 			);
 			HEL_CHECK(send_resp.error());
-		}else if(error && *error == Error::illegalArguments) {
-			resp.set_error(managarm::fs::Errors::ILLEGAL_ARGUMENT);
+			co_return;
```
qookei, line 171: Pretty sure this should be a `continue;`

no92: This is not in a loop though?

qookei: Oh duh my bad.

qookei (approved):

PR #652, opened 2024-04-07
Fix LLVM 18 warnings and remove protobuf

no92:
This PR fixes new warnings emitted by LLVM 18. As a bonus, all remaining protobuf protocols have been converted to bragi, as protobuf was producing massive amounts of warnings.


tools/bakesvr/src/main.cpp
```diff
@@ -0,0 +1,38 @@
...
+#include <yaml-cpp/yaml.h>
+#include <svrctl.bragi.hpp>
+
+int main(int argc, char **argv) {
+	std::string input{};
+	std::string output{};
+
+	CLI::App app{"bakesvr: generate info for runsvr"};
+	app.add_option("input", input, "Path to the input file")->required();
+	app.add_option("-o,--output", output, "Path to the input file")->required();
```
Dennisbonke, line 15: This should say path to the output file.

Dennisbonke (changes requested):
Quick glance showed me one error

hel/include/hel.h
```diff
@@ -389,7 +389,7 @@ enum HelSyscallArgs {
 #endif
 
 struct HelQueueParameters {
-	uint32_t flags;
+	uint32_t flags = 0;
```
qookei, line 392: I don't think these default member initializers are correct. This header is supposed to be usable from both C and C++, and C doesn't support them.

drivers/usb/hcds/xhci/src/xhci.hpp
```diff
@@ -136,7 +133,7 @@ struct Controller final : proto::BaseController {
 		uint8_t getSpeed();
 		int _id;
 		std::shared_ptr<Device> _device;
-		Controller *_controller;
```
qookei, line 139: I think we can just remove this member completely instead.

tools/bakesvr/meson.build
```diff
@@ -1,14 +1,10 @@
...
+	'src/main.cpp',
 )
 
-install_data('bakesvr', install_dir : get_option('bindir'))
+src += cxxbragi.process([protos/'svrctl/svrctl.bragi'])
+
+yaml_cpp_subproject = import('cmake').subproject('yaml-cpp')
+yaml_cpp_dep = yaml_cpp_subproject.dependency('yaml-cpp')
+
+executable('bakesvr', src, dependencies: [yaml_cpp_dep, cli11_dep, bragi_dep], install_dir : get_option('bindir'), native: true, install: true)
```
qookei, line 10: Could you reformat this? Spaces around `:` on both sides, and one option per line.

no92:
Addressed the comments raised.

qookei (approved):

PR #653, opened 2024-04-11
Redesign mbus object creation and enumeration

### LINKED ISSUE #158: mbus: Redesign object creation
mbus works reasonably well but after a few years of usage, some design limitations become apparent. This issue is part of a redesign of mbus.

Regarding **object creation**, mbus has a few issues:
* Objects are arranged in a hierarchical tree but this feature it not used by anywhere in Managarm. Instead, objects should reside in a single global flat (i.e., non-hierarchical) collection.
* On `BIND`, mbus queries the object's owner for a descriptor (such as a stream). This opens up a denial-of-service vector against the mbus server: object owners can just refuse to ever respond to that request. Instead, a single descriptor should be passed to mbus during object creation. More generally, mbus should be redesigned so that it *never* requests anything from a client.
* This change will also remove all callbacks that the current client-side library uses. (These callbacks are tedious to deal with in an asynchronous environment.)
* Objects should be associated with two streams: one stream held by the object owner and one stream passed to clients that enumerate the object. When the owner closes its stream, the object should be destroyed.

To implement this new design, we should add a `CREATE_OBJECT_NG` request, migrate existing code to `CREATE_OBJECT_NG` and then delete `CREATE_OBJECT` in favor of `CREATE_OBJECT_NG`.

### LINKED ISSUE #159: mbus: Redesign object enumeration
This is a follow up on #158 and blocked on that issue.

Currently, **object enumeration** in mbus uses a push-based model: the mbus server pushes updates to clients. This is problematic for multiple reasons: (i) it is tedious to integrate into asynchronous code, (ii) the client-side library requires some sort of callbacks or client-side buffering, (iii) it enables the possibility to DoS the mbus server.

Instead, object enumeration should use a pull-based model, i.e., we should let clients ask the mbus server for updates. This can use a mechanism based on sequence numbers: add a request to mbus that returns all objects that match a given filter and that were modified after the given sequence number.

qookei:
Access control and renaming `mbus_ng` to `mbus` will come in a separate pull request. The top 4 commits aren't really related to this PR (topmost is related to access control), but throwing them in here doesn't hurt.

Closes #158.
Closes #159.


Dennisbonke (approved):
LGTM

no92 (approved):
LGTM

PR #655, opened 2024-04-21
`dmesg` support

no92:


posix/subsystem/src/devices/kmsg.cpp
```diff
@@ -0,0 +1,154 @@
...
+#include "kmsg.hpp"
+
+namespace {
+
+struct KmsgFile final : File {
+private:
+	async::result<frg::expected<Error, size_t>>
+	readSome(Process *, void *data, size_t length) override {
+		arch::contiguous_pool dmaPool_;
+		arch::dma_buffer chunkBuffer{&dmaPool_, 2048};
```
qookei, line 18: I'm quite sure this can be a normal buffer and not a `dma_buffer` (so just `std::vector<char> chunkBuffer(2048);`).

no92:
Addressed the comment and made minor improvements with the fixups.

qookei (approved):

PR #662, opened 2024-05-09
Additions for dhcpcd

no92:


no92:
Marking as ready for review, as this PR provides the managarm changes needed to make `dhcpcd` work. In order to merge `dhcpcd`, some mlibc changes will be required; these are being cleaned up now, but might take a few days to complete.

servers/netserver/src/bpf/bpf.cpp
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	Op last = Op(prog_.back().code);
+	if(last != Op::RET_K && last != Op::RET_A)
+		return false;
+
+	return true;
+}
+
+#define bpf_log_op(args...) if(logBpfOps) { printf("\t[%.2zu/%.2zu] ", pc, prog_.size() - 1); printf(args); puts(""); } while(0)
```
ArsenArsen, line 33: why is this a macro?

no92: Originally, for quick implementation convenience; the fixup replaces it with a lambda.

servers/netserver/src/bpf/bpf.cpp
```diff
@@ -0,0 +1,125 @@
...
+			}
+			default:
+				printf("netserver: unhandled BPF instruction 0x%02x\n", inst.code);
+				bpf_log_op("{ 0x%02x, %.2u, %.2u, 0x%08x }\n", inst.code, inst.jt, inst.jf, inst.k);
+				assert(!"unhandled BPF instruction");
+		}
+
+	}
+
+	assert(!"invalid BPF filter with no return!");
```
ElectrodeYT, line 124: We should probably not be asserting on something like this

servers/netserver/src/bpf/bpf.cpp
```diff
@@ -0,0 +1,125 @@
...
+				break;
+			}
+			case Op::RET_K: {
+				bpf_log_op("RET k (0x%02x)", inst.k);
+				return inst.k;
+			}
+			default:
+				printf("netserver: unhandled BPF instruction 0x%02x\n", inst.code);
+				bpf_log_op("{ 0x%02x, %.2u, %.2u, 0x%08x }\n", inst.code, inst.jt, inst.jf, inst.k);
+				assert(!"unhandled BPF instruction");
```
ElectrodeYT, line 119: We should probably not be asserting on something like this

ElectrodeYT (changes requested):
Mostly LGTM, but we should really stop adding more asserts like these, if possible...

servers/netserver/src/bpf/bpf.cpp
no92: Any such BPF filter should be caught in the validation step, and as such this assert seems justified.

servers/netserver/src/bpf/bpf.cpp
no92: Given that the BPF support in this PR is only covering a part of the instruction set, failing catastrophically instead of silently failing seemed more appropriate.

servers/netserver/src/bpf/bpf.cpp
ElectrodeYT: Probably, yeah, but might be better to log/return an error instead anyway

servers/netserver/src/bpf/bpf.cpp
ElectrodeYT: We should not fail "silently", but returning an error or something is probably more appropriate imo

no92:
This PR now includes the ioctls implemented in #589, superseding it once merged.

servers/netserver/src/ip/udp4.cpp
```diff
@@ -343,12 +363,61 @@ struct Udp4Socket {
 		co_return len;
 	}
 
+	static async::result<frg::expected<protocols::fs::Error, protocols::fs::PollWaitResult>>
+	pollWait(void *obj, uint64_t past_seq, int mask, async::cancellation_token cancellation) {
```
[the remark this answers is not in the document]
no92: I think this is correct, as we don't indent a function name (as opposed to e.g. parameters).

servers/netserver/src/raw.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+		Bpf bpf{optbuf};
+		if(!bpf.validate())
+			co_return protocols::fs::Error::illegalArguments;
+
+		self->filter_ = optbuf;
+	} else if(layer == SOL_SOCKET && number == SO_DETACH_FILTER) {
+		self->filter_ = std::nullopt;
+	} else {
+		printf("netserver: unhandled setsockopt layer %d number %d\n", layer, number);
```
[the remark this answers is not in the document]
no92: Updated the PR to return `ENOPROTOOPT`, as linux does.

core/lib/bpf/bpf.cpp
```diff
@@ -0,0 +1,130 @@
...
+				break;
+			}
+			case Op::RET_K: {
+				bpf_log_op("RET k (0x%02x)", inst.k);
+				return inst.k;
+			}
+			default:
+				printf("netserver: unhandled BPF instruction 0x%02x\n", inst.code);
+				bpf_log_op("{ 0x%02x, %.2u, %.2u, 0x%08x }\n", inst.code, inst.jt, inst.jf, inst.k);
+				assert(!"unhandled BPF instruction");
```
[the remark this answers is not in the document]
no92: Added a TODO, as currently it is very likely to be an unimplemented opcode, not an invalid BPF filter. In the future, doing error handling (like returning EINVAL) would be nice, making BPF failures very clear is quite important to me.

ElectrodeYT (approved):
LGTM

servers/netserver/src/ip/udp4.cpp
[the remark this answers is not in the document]
qookei: It is how we format similarly long function decls in other places

qookei (approved):

Dennisbonke (approved):
LGTM

PR #667, opened 2024-05-27
Implement SIOCGIFBRDADDR/SIOCGIFHWADDR

Qwinci:
depends on https://github.com/managarm/managarm/pull/662

### DIFF
--- a/servers/netserver/src/main.cpp
+++ b/servers/netserver/src/main.cpp
@@ -276,6 +276,38 @@ async::detached serve(helix::UniqueLane lane) {
 				} else {
 					resp.set_error(managarm::fs::Errors::ILLEGAL_ARGUMENT);
 				}
+			}else if(req.command() == SIOCGIFBRDADDR) {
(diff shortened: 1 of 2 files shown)


servers/netserver/src/main.cpp
```diff
@@ -276,6 +276,37 @@ async::detached serve(helix::UniqueLane lane) {
...
+					}
+				} else {
+					resp.set_error(managarm::fs::Errors::ILLEGAL_ARGUMENT);
+				}
+			}else if(req.command() == SIOCGIFHWADDR) {
+				auto link = nic::Link::byName(req.name());
+
+				if(link) {
+					std::vector<uint8_t> mac{link->deviceMac().data(), link->deviceMac().data() + 6};
+					resp.set_mac(std::move(mac));
```
no92, line 305: With the recent bragi changes, this now takes a `std::array` instead of a `std::vector`.

Qwinci: Done.

PR #669, opened 2024-05-28
Don't log handled page faults by default

Qwinci:
Logging handled page faults mostly just adds unnecessary noise to the log when running eg. java/wine so disable it by default and add an option to log them in case its needed for debugging.


qookei (approved):

Dennisbonke (approved):
LGTM

PR #671, opened 2024-05-30
posix, ext2fs, fs, libblockfs: do not unlink non-empty directories

ElectrodeYT:


posix/subsystem/src/requests.cpp
```diff
@@ -2149,7 +2149,13 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 			auto owner = target_link->getOwner();
 			auto result = co_await owner->rmdir(target_link->getName());
 			if(!result) {
+				if(result.error() == Error::directoryNotEmpty) {
+					co_await sendErrorResponse(managarm::posix::Errors::DIRECTORY_NOT_EMPTY);
```
Dennisbonke, line 2153: Can you remove the print on line 2126? That is no longer true now that this code is in.

Dennisbonke (changes requested):
LGTM outside the one comment. Can you also add this error to sys_rmdir in the Managarm sysdeps (see [here](https://github.com/managarm/mlibc/blob/0fd8c51f50c3c5ed6ffdd81d2c5d620501fe520f/sysdeps/managarm/generic/file.cpp#L2114))? Right now, this will trip the ensure success call (oh, and while that's a thing, it's also missing ENOTDIR and EIEIO handling :)).

Dennisbonke (approved):
LGTM

PR #680, opened 2024-06-09
posix: Implement getFileFlags for fifo write and report read only status for fifo read

Dennisbonke:


posix/subsystem/src/fifo.cpp
```diff
@@ -164,8 +164,8 @@ struct ReaderFile : File {
 
 	async::result<int> getFileFlags() override {
 		if(nonBlock_)
-			co_return O_NONBLOCK;
```
no92, line 167: It would be preferable to have a flags variable where values are ORed in, which then is returned.

no92 (commented):
One complaint.

no92 (approved):
LGTM

PR #687, opened 2024-06-27
posix: Improve logging on unimplemented file operations

Dennisbonke:
This PR adds an auto launch of elogind if it's installed, and adds some required items (mostly workarounds) on the Managarm side. As a bonus, logging was improved for not implemented file operations.

### DIFF
--- a/posix/subsystem/src/file.cpp
+++ b/posix/subsystem/src/file.cpp
@@ -347,12 +347,16 @@ async::result<frg::expected<Error, size_t>> File::pwrite(Process *, int64_t, con
 }
 
 async::result<ReadEntriesResult> File::readEntries() {
+	std::cout << "posix \e[1;34m" << structName()
+			<< "\e[0m: Object does not implement readEntries()" << std::endl;
 	throw std::runtime_error("posix: Object has no File::readEntries()");
 }
 
 async::result<protocols::fs::RecvResult>
 File::recvMsg(Process *, uint32_t, void *, size_t,
 		void *, size_t, size_t) {
+	std::cout << "posix \e[1;34m" << structName()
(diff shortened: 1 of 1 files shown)


posix/init/src/stage2.cpp
```diff
@@ -178,6 +178,15 @@ int main() {
 		}else assert(dbus != -1);
 	}
 
+	// Launch elogind if it's available.
+	if(!access("/usr/lib/elogind/elogind", X_OK)) {
+		sleep(5); // Wait for dbus to be fully started, this is a hack.
```
no92, line 183: It would be nice to either get InitWare working soon™ or figure out what heuristic we can use for detecting dbus-daemon activation.

posix/init/src/stage2.cpp
```diff
@@ -178,6 +178,15 @@ int main() {
 		}else assert(dbus != -1);
 	}
 
+	// Launch elogind if it's available.
+	if(!access("/usr/lib/elogind/elogind", X_OK)) {
```
no92, line 182: There doesn't seem to be an elogind port upstream or PRed in yet?

Dennisbonke: That's correct, but there's no reason to not add it now with the port coming hopefully soon. It's just a matter of cleaning it up.

posix/init/src/stage2.cpp
Dennisbonke: Agreed. I should retry that soon

Dennisbonke:
Dropped all elogind changes as that won't be needed anymore or is already merged. This PR now only improves some logging for unimplemented file operations.

PR #690, opened 2024-07-07
`dmesg` improvements

no92:
This PR ensures that logging to dmesg is pretty (by using priorities and stripping CSI sequences) and works as expected. This includes eudev's logging to `/dev/kmsg`, which uses prefixes as defined in RFC 5424.

Needs to be merged along with managarm/mlibc#1090.

### DIFF
--- a/kernel/thor/generic/hel.cpp
+++ b/kernel/thor/generic/hel.cpp
@@ -4,6 +4,7 @@
 #include <async/algorithm.hpp>
 #include <async/cancellation.hpp>
 #include <frg/container_of.hpp>
+#include <frg/formatting.hpp>
 #include <frg/dyn_array.hpp>
 #include <frg/small_vector.hpp>
 #include <thor-internal/event.hpp>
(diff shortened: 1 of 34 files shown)


kernel/thor/generic/hel.cpp
```diff
@@ -125,10 +147,57 @@ HelError helLog(const char *string, size_t length) {
 		if(!readUserArray(string + offset, log.text, chunk))
 			return kHelErrFault;
 
-		auto p = infoLogger();
-		for(size_t i = 0; i < chunk; i++)
-			p << frg::char_fmt(log.text[i]);
-		p << frg::endlog;
+		switch(severity) {
+			case kHelLogSeverityEmergency: {
+				auto p = panicLogger();
```
qookei, line 152: I don't think letting the user trigger a kernel panic is a good idea :^)

no92: Fixed by rolling it into the case below.

kernel/thor/generic/ubsan.cpp
```diff
@@ -73,6 +73,10 @@ namespace {
 	}
 }
 
+extern "C" void abort() {
```
Dennisbonke, line 76: Why is this added? It seems unrelated.

Dennisbonke (commented):
One comment.

kernel/thor/generic/ubsan.cpp
no92: Fixed.

Dennisbonke (approved):
LGTM

PR #695, opened 2024-07-22
posix: Handle files that are less than the size of an elf header in exec

Dennisbonke:


posix/subsystem/src/file.cpp
```diff
@@ -304,7 +304,7 @@ async::result<frg::expected<Error>> File::readExactly(Process *process,
 		auto result = FRG_CO_TRY(co_await readSome(process,
 				(char *)data + offset, length - offset));
 		if(!result)
-			co_return Error::wouldBlock;
+			co_return Error::ioError;
```
qookei, line 307: I don't think this is makes sense? This just indicates that the end of file was reached (and we even have `Error::eof`?).

qookei (approved):

no92 (approved):
LGTM

PR #696, opened 2024-07-25
USB, mbus fixes and additions for MBIM (LTE) modem support

no92:


Dennisbonke (approved):
LGTM

ElectrodeYT (approved):
LGTM

mbus/src/main.cpp
```diff
@@ -476,6 +484,40 @@ async::detached serve(helix::UniqueLane lane) {
...
+				resp.set_error(managarm::mbus::Error::NO_SUCH_ENTITY);
+			} else {
+				for(auto p : req->properties()) {
+					entity->updateProperty(p.name(), p.string_item());
+				}
+
+				resp.set_error(managarm::mbus::Error::SUCCESS);
+			}
+
+			entitySeqTree.remove(entity.get());
```
qookei, line 517: this is broken if `!entity`

mbus/src/main.cpp
```diff
@@ -43,14 +44,22 @@ struct Entity {
 		return _seq;
 	}
 
+	void updateSeq(uint64_t val) {
```
qookei, line 47: Perhaps we should combine updateSeq and updateProperty such that they can only be updated at the same time?

mbus/src/main.cpp
no92: Fixed by moving the code that uses `entity` into the else branch above.

mbus/src/main.cpp
no92: I'm not sure how these necessarily relate?

qookei: I was thinking that the only time the seq is updated is when properties are being updated, so merging them would be clearer, but I suppose the only place we do that is when handling the request, so it doesn't really matter.

On another note, maybe we should add an `assert(val > seq_);` to the setter? Or just get rid of the setter and getter and make the field public if they have no special logic.

drivers/usb/hcds/xhci/src/main.cpp
```diff
@@ -1112,7 +1112,7 @@ Controller::EndpointState::transfer(proto::ControlTransfer info) {
 	}
 
 	// TODO(qookie): Report the residue to the user
-	(void)FRG_CO_TRY(maybeResidue);
+	co_return info.buffer.size() - FRG_CO_TRY(maybeResidue);
```
qookei, line 1115: Get rid of the TODO comment above and the extra return below?

posix/subsystem/src/common.hpp
```diff
@@ -25,3 +30,10 @@ struct StructName {
 	const char *_type;
 	uint64_t _id;
 };
+
+template <>
+struct std::formatter<StructName> : std::formatter<std::string> {
+    auto format(const StructName& obj, std::format_context& ctx) const {
+        return std::formatter<std::string>::format(obj.str(), ctx);
+    }
+};
```
qookei, line 39: Wrong indentation

mbus/src/main.cpp
no92: Added the `assert`.

Now that I think about it, keeping `updateSeq` separate might be useful for implementing object removal, for instance.

qookei (approved):

posix/subsystem/src/common.hpp
```diff
@@ -13,8 +14,12 @@ struct StructName {
 		return StructName{type, idCounter++};
 	}
 
+	std::string str() const {
```
avdgrinten, line 17: Surely there is a way to avoid a heap allocated string here? Can't we override the `std::formatter` w/o using an `std::string`? Then we can use `std::format` to implement `operator<<`.

mbus/src/main.cpp
```diff
@@ -161,7 +162,12 @@ static bool matchesFilter(const Entity *entity, const AnyFilter &filter) {
 		auto it = properties.find(real->getProperty());
 		if(it == properties.end())
 			return false;
-		return it->second == real->getValue();
+		if(std::holds_alternative<mbus_ng::StringItem>(real->getValue()) &&
+				std::holds_alternative<mbus_ng::StringItem>(it->second)) {
+			return std::get<mbus_ng::StringItem>(it->second).value == std::get<mbus_ng::StringItem>(real->getValue()).value;
+		} else {
+			assert(!"unhandled type in mbus item matching");
```
avdgrinten, line 169: This should not assert but return an error.

posix/subsystem/src/subsystem/usb/usb.cpp
```diff
@@ -194,17 +194,20 @@ async::result<void> bindDevice(mbus_ng::Entity entity, mbus_ng::Properties prope
...
 	device->speed = std::get<mbus_ng::StringItem>(properties["usb.speed"]).value;
 
+	auto config_val = (co_await device->device().currentConfigurationValue()).value();
+	std::string raw_desc;
+
 	/* obtain the tree of configuration descriptors and its subdescriptors */
 	auto devdesc = reinterpret_cast<protocols::usb::DeviceDescriptor *>(raw_dev_desc.data());
 	for(size_t i = 0; i < devdesc->numConfigs; i++) {
 		auto raw_descs = (co_await device->device().configurationDescriptor(i)).value();
+		auto devdesc = reinterpret_cast<protocols::usb::ConfigDescriptor *>(raw_descs.data());
```
avdgrinten, line 204: This should probably do a memcpy instead. Or we can add some helper function to libarch to do the memcpy.

posix/subsystem/src/subsystem/usb/usb.cpp
```diff
@@ -194,17 +194,20 @@ async::result<void> bindDevice(mbus_ng::Entity entity, mbus_ng::Properties prope
...
+		if(devdesc->configValue == config_val)
```
avdgrinten, line 205: What about the endianness here?

avdgrinten (changes requested):

posix/subsystem/src/subsystem/usb/usb.cpp
no92: USB is little-endian, so I don't understand where endian conversion would come into play here?

avdgrinten: It becomes relevant if the host is big endian.

no92: `configValue` is a uint8_t anyways.

mbus/src/main.cpp
no92: Encountering this in the mbus server sure seems like a bug; how would a error return look like in this case?

avdgrinten: Why would that be a bug? This would be triggered if the user tried to match a string against an array, right? Isn't it triggerable by clients?

no92:
Addressed the comments in the fixups.

avdgrinten (approved):

Dennisbonke (approved):
LGTM

PR #698, opened 2024-07-26
Some self-hosting fixes

Dennisbonke:
Found while trying to build `GNU Hello` and `zlib` inside Managarm


ElectrodeYT (approved):
LGTM

ElectrodeYT:
While this code should work, it can be optimized: putting only the resp.set_error into the if statements, and moving the stuff below it to after the if-else-if tree would be optimal.

Dennisbonke:
> While this code should work, it can be optimized: putting only the resp.set_error into the if statements, and moving the stuff below it to after the if-else-if tree would be optimal.

Actually, that would break the fileNotFound case, which we currently ignore. We don't want to send an empty response in that case.

PR #707, opened 2024-09-01
libblockfs and init-stage1 improvements

qookei:
Now all partitions are exposed in /dev and /sys, and init-stage1 finds the root partition by looking for one with the appropriate type in the partition table.


Dennisbonke (approved):
LGTM

no92 (approved):
LGTM

PR #711, opened 2024-09-10
USB fixes and XHCI code cleanup

qookei:


drivers/usb/hcds/xhci/src/xhci.hpp
```diff
@@ -261,6 +262,18 @@ struct Controller final : proto::BaseController {
 		_dcbaa[slot] = helix::ptrToPhysical(ctx.rawData());
 	}
 
+	std::string_view name() const {
+		return _name;
+	}
+
+	friend std::ostream &operator<<(std::ostream &os, Controller *controller) {
```
no92, line 269: This might be personal preference, but I do like to use std::format for printing logs; that would also allow us to use the `<print>` header for this when we upgrade to GCC 14 and C++23.

no92 (approved):
As far as my knowledge goes (that is, nothing about HCDs), LGTM.

Dennisbonke (approved):
LGTM

PR #717, opened 2024-09-23
Implement sysconf for CPU online count

Dennisbonke:


posix/subsystem/src/requests.cpp
```diff
@@ -3483,6 +3484,55 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+				resp.set_value(kernResp->num_cpu());
+
+				auto [send_resp] = co_await helix_ng::exchangeMsgs(
+					conversation,
+					helix_ng::sendBragiHeadOnly(resp, frg::stl_allocator{})
+				);
+
+				HEL_CHECK(send_resp.error());
+			} else {
+				// Not handled, bubble up EINVAL.
```
no92, line 3526: Maybe we should log the unhandled sysconf request num?

posix/subsystem/src/requests.cpp
```diff
@@ -3483,6 +3484,55 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+				resp.set_error(managarm::posix::Errors::ILLEGAL_ARGUMENTS);
+
+				auto [send_resp] = co_await helix_ng::exchangeMsgs(
```
no92, line 3529: The sending of the reply in both branches could be moved outside of the if-else-statement.

no92 (commented):
Two small notes, other than that LGTM.

posix/subsystem/src/requests.cpp
Dennisbonke: That is done in mlibc land, EINVAL will get handled by the main sysconf function, logging for default values and panicking for unimplemented ones.

posix/subsystem/src/requests.cpp
Dennisbonke: True, will fix

no92 (approved):
LGTM

PR #724, opened 2024-10-01
Implement shutting down the system from userspace

Dennisbonke:
Supersedes #636 (I did an oopsie with git).
I fixed the comments on that PR and use the Hw protocol to accomplish the same.
Managarm counterpart of managarm/mlibc#1034

### DIFF
--- a/kernel/thor/system/acpi/pm-interface.cpp
+++ b/kernel/thor/system/acpi/pm-interface.cpp
@@ -56,24 +56,7 @@ private:
 		if (preamble.error())
 			co_return Error::protocolViolation;
 
-		if(preamble.id() == bragi::message_id<managarm::hw::PmResetRequest>) {
-			auto req = bragi::parse_head_only<managarm::hw::PmResetRequest>(reqBuffer, *kernelAlloc);
-
-			if (!req) {
-				infoLogger() << "thor: Closing lane due to illegal HW request." << frg::endlog;
-				co_return Error::protocolViolation;
-			}
-
-			auto ret = uacpi_reboot();
-			if(uacpi_unlikely_error(ret))
-				infoLogger() << "thor: ACPI reset failed: " << uacpi_status_to_string(ret) << frg::endlog;
-
-#ifdef __x86_64__
-			issuePs2Reset();
-			infoLogger() << "thor: Reset using PS/2 controller failed" << frg::endlog;
-#endif
-			panicLogger() << "thor: We do not know how to reset" << frg::endlog;
-		} else if(preamble.id() == bragi::message_id<managarm::hw::RebootRequest>) {
+		if(preamble.id() == bragi::message_id<managarm::hw::RebootRequest>) {
 			auto req = bragi::parse_head_only<managarm::hw::RebootRequest>(reqBuffer, *kernelAlloc);
 
 			if(!req) {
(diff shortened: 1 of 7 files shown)


kernel/thor/system/acpi/pm-interface.cpp
```diff
@@ -71,6 +71,53 @@ struct PmInterfaceBusObject : private KernelBusObject {
...
+				co_return Error::protocolViolation;
+			}
+
+			uacpi_sleep_state state = UACPI_SLEEP_STATE_S0;
+			switch(req->cmd()) {
+				case 5:
+					state = UACPI_SLEEP_STATE_S5;
+					break;
+				case 6: {
+					auto ret = uacpi_reboot();
```
d-tatianin, line 88: This should still call `uacpi_prepare_for_sleep_state(S5)`, see https://github.com/torvalds/linux/commit/d60cd06331a3566d3305b3c7b566e79edf4e2095

kernel/thor/system/acpi/pm-interface.cpp
```diff
@@ -71,6 +71,53 @@ struct PmInterfaceBusObject : private KernelBusObject {
...
+			auto ret = uacpi_prepare_for_sleep_state(state);
+			if(uacpi_unlikely_error(ret))
+				infoLogger() << "thor: Preparing to enter sleep state S" << (int)state << " failed: " << uacpi_status_to_string(ret) << frg::endlog;
+
+			// uACPI Documentation states that you must call uacpi_enter_sleep_state with interrupts disabled.
+			disableInts();
+			
+			ret = uacpi_enter_sleep_state(state);
+			if(uacpi_unlikely_error(ret))
+				infoLogger() << "thor: Entering sleep state S" << (int)state << " failed: " << uacpi_status_to_string(ret) << frg::endlog;
```
no92, line 113: Using std::format seems cleaner.

kernel/thor/system/acpi/pm-interface.cpp
```diff
@@ -71,6 +71,53 @@ struct PmInterfaceBusObject : private KernelBusObject {
...
+
+			// Either reboot failed, poweroff failed, or someone is funny and thinks we can suspend.
+			// Either way, return succes.
+
+			// Oops, re-enable before returning.
+			enableInts();
+			co_return frg::success;
```
no92, line 120: Is returning success here the best play?

posix/subsystem/src/main.cpp
```diff
@@ -82,12 +82,17 @@ async::result<void> serve(std::shared_ptr<Process> self, std::shared_ptr<Generat
 
 namespace {
 	helix::UniqueLane kerncfgLane;
+	helix::UniqueLane hwLane;
```
no92, line 85: Naming this `hwLane` doesn't quite do justice to the fact that this lane is only connecting to our power management interface (using the hw protocol), maybe something like `pmLane` would be clearer?

no92 (commented):
Some nits.

posix/subsystem/src/main.cpp
Dennisbonke: Yeah sounds good. I'll refactor that.

kernel/thor/system/acpi/pm-interface.cpp
Dennisbonke: Well it should never happen anyway but I'm open for suggestions.

kernel/thor/system/acpi/pm-interface.cpp
Dennisbonke: I'll take a look at it

kernel/thor/system/acpi/pm-interface.cpp
qookei: IMO just panic here, we were asked to shut down and failed, so there's nothing better we can do.

posix/subsystem/src/requests.cpp
```diff
@@ -387,6 +388,43 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+			if (!req) {
+				std::cout << "posix: Rejecting request due to decoding failure" << std::endl;
+				break;
+			}
+
+			if(logRequests)
+				std::cout << "posix: REBOOT with sleep state: " << req->cmd() << std::endl;
+
+			if(self->uid() != 0)
+				sendErrorResponse(managarm::posix::Errors::INSUFFICIENT_PERMISSION);
```
qookei, line 408: Missing `continue;`?

protocols/hw/hw.bragi
```diff
@@ -121,3 +121,9 @@ tail:
 		tag(8) uint32 fb_type;
 	}
 }
+
+message RebootRequest 17 {
+head(128):
+	Errors error;
+	int32 cmd;
```
qookei, line 128: I think it'd be better if we had an enum of possible commands (shutdown, reboot) instead of the raw ACPI sleep state value (and a made up sleep state 6 for reboot). Also applies to the posix request (maybe there we could just pass through the RB_POWER_OFF etc values?).

d-tatianin: agreed because sleep states other than 5 require additional preparation, u cant just enter them willy nilly

protocols/hw/hw.bragi
```diff
@@ -98,8 +98,10 @@ message AccessFbMemoryRequest 11 {
 head(128):
 }
 
-message PmResetRequest 12 {
+message RebootRequest 12 {
 head(128):
+	Errors error;
```
qookei, line 103: Why does the request have an `error` member?

kernel/thor/system/acpi/pm-interface.cpp
```diff
@@ -54,23 +56,56 @@ struct PmInterfaceBusObject : private KernelBusObject {
...
+					panicLogger() << "thor: Poweroff failed" << frg::endlog;
+					break;
+				}
+				case RB_AUTOBOOT: {
+					state = UACPI_SLEEP_STATE_S5;
+					auto ret = uacpi_prepare_for_sleep_state(state);
+					if(uacpi_unlikely_error(ret))
+						infoLogger() << "thor: Preparing for reboot failed: " << uacpi_status_to_string(ret) << frg::endlog;
+
+					ret = UACPI_STATUS_OK;
```
qookei, line 92: What's this doing here? Also probably a good idea to turn off interrupts here too.

Dennisbonke: Yep, will remove

protocols/hw/hw.bragi
Dennisbonke: Copy paste, will remove

qookei (approved):

PR #726, opened 2024-10-03
meson: fix warning for boolean option

marv7000:
Removed the quotes around a boolean value, which causes a warning when building.


no92 (approved):
LGTM

PR #727, opened 2024-10-11
PS/2 driver rework for ACPI detection support

no92:
This PR enables detection of i8042 parameters (I/O ports, IRQ numbers) via ACPI.

### DIFF
--- a/kernel/thor/system/acpi/acpi.cpp
+++ b/kernel/thor/system/acpi/acpi.cpp
@@ -188,6 +188,71 @@ coroutine<frg::expected<Error>> AcpiObject::handleRequest(LaneHandle lane) {
 		auto ioError = co_await PushDescriptorSender{conversation, IoDescriptor{space}};
 		if(ioError != Error::success)
 			co_return ioError;
+	} else if(preamble.id() == bragi::message_id<managarm::hw::AccessIrqRequest>) {
+		auto req = bragi::parse_head_only<managarm::hw::AccessIrqRequest>(reqBuffer, *kernelAlloc);
+
+		if (!req) {
+			infoLogger() << "thor: Closing lane due to illegal HW request." << frg::endlog;
+			co_return Error::protocolViolation;
+		}
+
+		managarm::hw::SvrResponse<KernelAlloc> resp{*kernelAlloc};
+		resp.set_error(managarm::hw::Errors::SUCCESS);
+
+		struct InterruptInfo {
+			size_t requested_index;
+			size_t parsed_irqs;
+			std::optional<int> irq;
+		} interrupt_info = {req->index(), 0, std::nullopt};
+
+		// TODO(no92): we should cache this
+		auto ret = uacpi_for_each_device_resource(node, "_CRS",
+		[](void *ctx, uacpi_resource *res){
+			auto info = reinterpret_cast<struct InterruptInfo *>(ctx);
+
+			switch(res->type) {
+				case UACPI_RESOURCE_TYPE_END_TAG:
+					break;
+				case UACPI_RESOURCE_TYPE_IRQ:
+					for(size_t i = 0; i < res->irq.num_irqs; i++) {
+						if(info->parsed_irqs == info->requested_index)
+							info->irq = res->irq.irqs[i];
+						info->parsed_irqs++;
+					}
+					break;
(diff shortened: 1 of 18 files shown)


kernel/thor/system/acpi/ps2.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+		struct find_context {
+			frg::vector<smarter::shared_ptr<async::oneshot_event>, KernelAlloc> *events;
+			const uacpi_char *hid;
+		};
+
+		for(auto hid : ACPI_HID_PS2_KEYBOARDS) {
+			struct find_context context{ &events, hid };
+
+			uacpi_find_devices(hid, [](void *ctx, uacpi_namespace_node *node) {
```
d-tatianin, line 69: Here you probably want `uacpi_find_devices_at`, which accepts a list of HIDs to look for

This one specifically won't work because a keyboard might match multiple HIDs at the same time and you don't do any deduplication here.

kernel/thor/system/acpi/acpi.cpp
```diff
@@ -36,8 +40,65 @@ coroutine<frg::expected<Error>> AcpiObject::handleRequest(LaneHandle lane) {
...
+		auto req = bragi::parse_head_only<managarm::acpi::AcpiGetResourcesRequest>(reqBuffer, *kernelAlloc);
+
+		managarm::acpi::AcpiGetResourcesReply<KernelAlloc> resp(*kernelAlloc);
+
+		uacpi_resources *resources = nullptr;
+		auto status = uacpi_get_current_resources(node, &resources);
+		if(status == UACPI_STATUS_OK) {
+			resp.set_error(managarm::acpi::Errors::SUCCESS);
+
+			auto ret = uacpi_for_each_resource(resources,
```
d-tatianin, line 53: `uacpi_for_each_device_resource` shorthand will work better here

kernel/thor/system/acpi/acpi.cpp
```diff
@@ -36,8 +40,65 @@ coroutine<frg::expected<Error>> AcpiObject::handleRequest(LaneHandle lane) {
...
+			[](void *ctx, uacpi_resource *res){
+				auto resp = reinterpret_cast<managarm::acpi::AcpiGetResourcesReply<KernelAlloc> *>(ctx);
+
+				switch(res->type) {
+					case UACPI_RESOURCE_TYPE_END_TAG:
+						break;
+					case UACPI_RESOURCE_TYPE_IO:
```
d-tatianin, line 60: This should support at least `UACPI_RESOURCE_TYPE_IO` and `UACPI_RESOURCE_TYPE_FIXED_IO`

kernel/thor/system/acpi/acpi.cpp
```diff
@@ -36,8 +40,65 @@ coroutine<frg::expected<Error>> AcpiObject::handleRequest(LaneHandle lane) {
...
+						for(auto i = res->io.minimum; i <= res->io.maximum; i++) {
+							resp->add_io_ports(i);
+						}
+						break;
+					case UACPI_RESOURCE_TYPE_IRQ:
```
d-tatianin, line 65: This should support at least `UACPI_RESOURCE_TYPE_IRQ` and `UACPI_RESOURCE_TYPE_EXTENDED_IRQ`

kernel/thor/system/acpi/ps2.cpp
```diff
@@ -0,0 +1,119 @@
+#include <bragi/helpers-all.hpp>
+#include <bragi/helpers-frigg.hpp>
+#include <thor-internal/acpi/acpi.hpp>
+#include <thor-internal/main.hpp>
+#include <thor-internal/fiber.hpp>
+#include <uacpi/utilities.h>
+
+namespace {
+
+size_t next_keyboard_id = 0;
```
d-tatianin, line 10: Maybe we should have a generic sequence allocator?

d-tatianin:
As a general thought, maybe instead of calling `find` for every possible device, we could do a similar idea to [this](https://wiki.osdev.org/UACPI#Treat_ACPI_Namespace_as_a_Bus)?

Doesn't have to be done in this PR, just an idea for improving both design & performance of this

no92:
Updated the PR to fix the comments raised above and to prefer use of the hw protocol.

avdgrinten (approved):

PR #730, opened 2024-10-30
eir: add UEFI boot protocol

marv7000:
This PR adds UEFI as a boot protocol to eir. It's architecture independent and introduces a "protos" directory for future independent protocols.

### DIFF
--- a/kernel/eir/arch/arm/load64.S
+++ b/kernel/eir/arch/arm/load64.S
@@ -30,4 +30,6 @@ eirEnterKernel:
 	mov x0, x4
 	br x2
 
+#ifndef __clang__
 	.section .note.GNU-stack,"",%progbits
+#endif
--- a/kernel/eir/arch/x86/load64.S
+++ b/kernel/eir/arch/x86/load64.S
@@ -26,4 +26,6 @@ eirEnterKernel:
 	movabs $0xFFFFFE8000010000, %rdi
 	jmp *%rsi
 
+#ifndef __clang__
 	.section .note.GNU-stack,"",%progbits
+#endif
--- a/kernel/eir/meson.build
+++ b/kernel/eir/meson.build
@@ -21,7 +21,7 @@ libgcc = cxx.find_library('gcc', required: not build_uefi)
 eir_c_args = []
 eir_cpp_args = ['-fno-threadsafe-statics', '-DFRG_DONT_USE_LONG_DOUBLE']
 eir_link_args = ['-nostdlib']
-eir_dependencies = [libgcc, cxxshim_dep, frigg, libarch]
+eir_dependencies = [libgcc, frigg, libarch]
 
 if get_option('kernel_kasan')
 	eir_c_args += [
--- a/kernel/eir/meson.build
+++ b/kernel/eir/meson.build
@@ -16,7 +16,7 @@ eir_includes = [include_directories(
 eir_includes += uacpi_includes
 
 cxx = meson.get_compiler('cpp')
-libgcc = cxx.find_library('gcc')
+libgcc = cxx.find_library('gcc', required: not build_uefi)
 
 eir_c_args = []
 eir_cpp_args = ['-fno-threadsafe-statics', '-DFRG_DONT_USE_LONG_DOUBLE']
@@ -41,4 +41,9 @@ elif arch == 'riscv64'
 	subdir('arch/riscv')
 elif arch == 'x86_64'
 	subdir('arch/x86')
+endif
+
+if build_uefi
+	eir_cpp_args = []
+	subdir('protos/uefi')
 endif
--- a/kernel/eir/protos/uefi/entry.cpp
+++ b/kernel/eir/protos/uefi/entry.cpp
@@ -1,0 +1,351 @@
+#include <arch/io_space.hpp>
+#include <assert.h>
+#include <eir-internal/arch.hpp>
+#include <eir-internal/debug.hpp>
+#include <eir-internal/generic.hpp>
+#include <frg/array.hpp>
+#include <frg/cmdline.hpp>
+#include <stdbool.h>
+
+#include "efi.hpp"
+#include "helpers.hpp"
+
+static_assert(sizeof(char16_t) == sizeof(wchar_t), "Strings are not UTF-16-ish, are you missing -fshort-wchar?");
+extern "C" void eirEnterKernel(uintptr_t, uint64_t, uint64_t) __attribute__((sysv_abi));
+
+namespace eir {
+
+const efi_system_table *st = nullptr;
+const efi_boot_services *bs = nullptr;
+efi_handle handle = nullptr;
+
+namespace {
+
+void uefiBootServicesLogHandler(const char c) {
+	if(bs) {
+		if(c == '\n') {
+			frg::array<char16_t, 3> newline = {u"\r\n"};
+			st->con_out->output_string(st->con_out, newline.data());
+			return;
+		}
+
+		frg::array<char16_t, 2> converted = {static_cast<char16_t>(c), 0};
+		st->con_out->output_string(st->con_out, converted.data());
+	}
+}
+
+} // namespace
+
+extern "C" efi_status eirUefiMain(const efi_handle h, const efi_system_table *system_table) {
+	// Set the system table so we can use loggers early on.
+	st = system_table;
+	bs = st->boot_services;
+	handle = h;
+
+	logHandler = uefiBootServicesLogHandler;
+
+	// Reset the watchdog timer.
+	EFI_CHECK(bs->set_watchdog_timer(0, 0, 0, nullptr));
+	EFI_CHECK(st->con_out->clear_screen(st->con_out));
+
+	// Get a handle to this binary in order to get the command line.
+	efi_guid protocol = EFI_LOADED_IMAGE_PROTOCOL_GUID;
+	efi_loaded_image_protocol *loadedImage = nullptr;
+	EFI_CHECK(bs->handle_protocol(handle, &protocol, reinterpret_cast<void **>(&loadedImage)));
+
+	// Convert the command line to ASCII.
+	char *cmdLine = nullptr;
+	{
+		EFI_CHECK(bs->allocate_pool(EfiLoaderData,
+			(loadedImage->load_options_size / sizeof(uint16_t)) + 1,
+			reinterpret_cast<void **>(&cmdLine)));
+		size_t i = 0;
+		for(; i < loadedImage->load_options_size / sizeof(char16_t); i++) {
+			auto c = reinterpret_cast<char16_t *>(loadedImage->load_options)[i];
+			// we only use printable ASCII characters, everything else gets discarded
+			if(c >= 0x20 && c <= 0x7E) {
+				cmdLine[i] = c;
+			} else {
+				cmdLine[i] = '\0';
+			}
+		}
+		// Null-terminate the buffer.
+		cmdLine[i] = '\0';
+	}
+	assert(cmdLine);
+
+	bool eir_gdb_ready_val = true;
+	frg::string_view thor_path = "managarm\\thor";
+	frg::string_view initrd_path = "managarm\\initrd.cpio";
+
+	frg::array args = {
+		// allow for attaching GDB to eir
+		frg::option{"eir.efidebug", frg::store_false(eir_gdb_ready_val)},
+		frg::option{"bochs", frg::store_true(log_e9)},
+		frg::option{"eir.thor", frg::as_string_view(thor_path)},
+		frg::option{"eir.initrd", frg::as_string_view(initrd_path)},
+	};
+	frg::parse_arguments(cmdLine, args);
+
+	eir::infoLogger() << "eir: command line='" << cmdLine << "'" << frg::endlog;
+
+	// this needs to be volatile as GDB sets this to true on attach
+	volatile bool eir_gdb_ready = eir_gdb_ready_val;
+
+	if(!eir_gdb_ready_val) {
+		// exfiltrate our base address for use with gdb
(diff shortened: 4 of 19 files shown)


no92:
Depends on managarm/frigg#62.

kernel/eir/meson.build
```diff
@@ -1,8 +1,11 @@
 eir_generic_sources = files(
 	'../common/libc.cpp',
 	'../common/font-8x16.cpp',
-	'generic/main.cpp',
-	'generic/debug.cpp')
+	'generic/main.cpp')
+
+if not build_uefi
+	eir_generic_sources += files('generic/debug.cpp')
+endif
```
avdgrinten, line 8: Instead of overriding `debug.cpp`, `debug.cpp` should offer a way for the protocol to install a custom logging sink. For example, by simply storing a global `void (*logUsingBootProto)(const char *)`.

kernel/eir/meson.build
```diff
@@ -13,6 +16,22 @@ eir_includes = [include_directories(
...
+	eir_dependencies = [frigg, libarch]
+	if arch == 'aarch64'
+		subdir('arch/arm')
+	elif arch == 'riscv64'
+		subdir('arch/riscv')
+	elif arch == 'x86_64'
+		subdir('arch/x86')
+	endif
+	subdir('protos/uefi')
+	subdir_done()
```
avdgrinten, line 32: Why is this done instead of continuing with the remainder of the meson.build?

kernel/eir/protos/uefi/eir-internal/arch/types.hpp
```diff
@@ -0,0 +1,11 @@
+#pragma once
```
avdgrinten, line 1: An `arch/` header should not be in `protos/`.

kernel/memory.txt
```diff
@@ -31,11 +31,6 @@ FFFF'F000'0000'0000
 	Log ring buffer memory
 	Referenced in eir/main.cpp, thor/generic/core.cpp
 
-FFFF'F000'0000'0000
-	Length: 256MiB
-	Log ring buffer memory
-	Referenced in eir/main.cpp, thor/generic/core.cpp
-
```
avdgrinten, line 38: I assume that this is outdated?

meson_options.txt
```diff
@@ -1,15 +1,20 @@
...
+    type : 'boolean',
     value : false
 )
 
-option('build_drivers', 
-    type : 'boolean', 
+option('build_uefi',
+    type : 'boolean',
     value : false
 )
```
avdgrinten, line 9: Can't we deduce this from the target triple?

meson.build
```diff
@@ -44,7 +45,7 @@ if build_kernel
 	add_project_link_arguments(['-Wl,--orphan-handling=warn', '-Wl,-znoexecstack'], language: ['c', 'cpp'])
 endif
 
-if build_kernel or build_drivers
+if build_kernel or build_uefi or build_drivers
 	add_languages('c', 'cpp', native: false)
 endif
```
avdgrinten, line 50: A lot of the meson.build logic can be simplified if we assume that `build_kernel` is set whenever `build_uefi` is set.

kernel/eir/protos/uefi/meson.build
```diff
@@ -0,0 +1,25 @@
...
+	'freestnd-c-hdrs-' + host_machine.cpu_family(),
+	required: true,
+	fallback: ['freestnd-c-hdrs', 'freestnd_c_hdrs_dep'],
+)
+
+freestnd_cxx_hdrs_dep = dependency(
+	'freestnd-cxx-hdrs-' + host_machine.cpu_family(),
+	required: true,
+	fallback: ['freestnd-cxx-hdrs', 'freestnd_cxx_hdrs_dep'],
+)
```
avdgrinten, line 17: Shouldn't eir in general use these dependencies (and not just the uefi target)?

avdgrinten (commented):
Reviewed the build changes. I did not review the EFI code yet.

kernel/eir/arch/arm/load64.S
```diff
@@ -30,4 +30,6 @@ eirEnterKernel:
 	mov x0, x4
 	br x2
 
+#ifndef __clang__
 	.section .note.GNU-stack,"",%progbits
+#endif
```
avdgrinten, line 35: Why is this gated behind non-Clang instead of non-UEFI?

kernel/eir/arch/x86/arch.cpp
```diff
@@ -178,6 +178,12 @@ void initProcessorEarly() {
 	// Keep in sync with the SMP trampoline in thor.
 	uint64_t pat = 0x00'00'01'00'00'00'04'06;
 	common::x86::wrmsr(0x277, pat);
+
+	// Clear Write Protect bit.
+	uint64_t cr0;
+	asm volatile ("mov %%cr0, %0" : "=r" (cr0));
+	cr0 &= ~(1 << 16);
+	asm volatile ("mov %0, %%cr0" : : "r" (cr0));
```
avdgrinten, line 186: Why is this necessary?

kernel/eir/arch/x86/load64.S
```diff
@@ -22,8 +22,11 @@ eirEnterKernel:
 	or $0x00010000, %rax # Enable WP
 	mov %rax, %cr0
 
+	cli
```
avdgrinten, line 25: Do UEFI apps run with interrupts enabled? What if an interrupt happens before this `cli`?

kernel/eir/protos/uefi/entry.cpp
```diff
@@ -0,0 +1,327 @@
...
+#include "helpers.hpp"
+
+static_assert(sizeof(char16_t) == sizeof(wchar_t), "Strings are not UTF-16-ish, are you missing -fshort-wchar?");
+extern "C" void eirEnterKernel(uintptr_t, uint64_t, uint64_t) __attribute__((sysv_abi));
+
+namespace eir {
+
+const efi_system_table *st = nullptr;
+const efi_boot_services *bs = nullptr;
+efi_handle handle = nullptr;
```
avdgrinten, line 20: Are these variables local to the file? If yes, they should be in an anonymous namespace. Otherwise, it would be good to pick more descriptive names.

kernel/eir/protos/uefi/entry.cpp
```diff
@@ -0,0 +1,327 @@
...
+	frg::parse_arguments(cmdLine, args);
+
+	eir::infoLogger() << "eir: command line='" << cmdLine << "'" << frg::endlog;
+
+	// this needs to be volatile as GDB sets this to true on attach
+	volatile bool __eir_gdb_ready = eir_gdb_ready_val;
+
+	if(!eir_gdb_ready_val) {
+		constexpr arch::scalar_register<uint8_t> offset{0};
+		auto port = arch::global_io.subspace(0xCB7);
```
avdgrinten, line 78: Please add a comment what this magic port number does.

kernel/eir/protos/uefi/entry.cpp
```diff
@@ -0,0 +1,327 @@
...
+	efi_file_protocol *kernelFile = nullptr;
+	EFI_CHECK(fsOpen(&kernelFile, asciiToUcs2(thor_path)));
+	auto kernel_info = fsGetInfo(kernelFile);
+	efi_physical_addr thor = 0;
+	EFI_CHECK(bs->allocate_pages(AllocateAnyPages, EfiLoaderData, (kernel_info->file_size >> 12) + 1, &thor));
+	fsRead(kernelFile, kernel_info->file_size, 0, thor);
+
+	efi_file_protocol *initrdFile = nullptr;
+	EFI_CHECK(fsOpen(&initrdFile, asciiToUcs2(initrd_path)));
+	efi_file_info *initrdInfo = fsGetInfo(initrdFile);
```
avdgrinten, line 108: In the future, we want to store `thor` inside the initrd CPIO. Does it make sense to implement this model right away? CC @qookei

kernel/eir/protos/uefi/entry.cpp
```diff
@@ -0,0 +1,327 @@
...
+	EFI_CHECK(bs->allocate_pages(AllocateAnyPages, EfiLoaderData, (initrdInfo->file_size >> 12) + 1, &initrd));
+	EFI_CHECK(fsRead(initrdFile, initrdInfo->file_size, 0, initrd));
+
+	// Get the frame buffer.
+	efi_graphics_output_protocol *gop = nullptr;
+
+	{
+		efi_guid gop_protocol = EFI_GRAPHICS_OUTPUT_PROTOCOL_GUID;
+		EFI_CHECK(bs->locate_protocol(&gop_protocol, nullptr, reinterpret_cast<void **>(&gop)));
+		assert(gop->mode->info->version == 0);
```
avdgrinten, line 121: We should not use `assert()` to check for runtime errors.

kernel/eir/protos/uefi/entry.cpp
```diff
@@ -0,0 +1,327 @@
...
+	// Exit boot services.
+	EFI_CHECK(bs->exit_boot_services(handle, mapKey));
+	bs = nullptr;
+
+	frg::array<InitialRegion, 4> reservedRegions = {};
+	size_t nReservedRegions = 0;
+
+	reservedRegions[nReservedRegions++] = {reinterpret_cast<uintptr_t>(loadedImage->image_base), loadedImage->image_size};
+	reservedRegions[nReservedRegions++] = {reinterpret_cast<uintptr_t>(thor), kernel_info->file_size};
+	reservedRegions[nReservedRegions++] = {reinterpret_cast<uintptr_t>(initrd), initrdInfo->file_size};
```
avdgrinten, line 162: What about the stuff that is allocated above? In particular, `cmdLine`, the memory map and allocations in helpers.cpp?

kernel/eir/protos/uefi/helpers.hpp
```diff
@@ -0,0 +1,28 @@
...
+void EFI_CHECK(efi_status s, std::source_location loc = std::source_location::current());
+
+namespace eir {
+
+extern const efi_system_table *st;
+extern const efi_boot_services *bs;
+extern efi_handle handle;
+
+extern bool __eir_log_debugcon;
+extern bool __eir_log_e9;
```
avdgrinten, line 20: Why do we have double underscore here?

kernel/eir/arch/arm/load64.S
marv7000: That syntax isn't supported in Clang's integrated assembler, and LLVM defaults to `-z noexecstack`. It is unrelated to UEFI.

kernel/eir/meson.build
marv7000: IIRC libgcc or cxxshim caused build failures and/or are incompatible with the UEFI target.

avdgrinten: Then these libraries should be guarded by `if not build_uefi`. The `subdir_done()` also skips `KERNEL_KASAN` and `KERNEL_LOG_ALLOCATIONS` and disables a bunch of other meson vars that architectures may want to use.

kernel/eir/arch/x86/arch.cpp
no92: We did this to work around something before, but this isn't necessary any more AFAICT.

kernel/eir/arch/x86/load64.S
no92: We do this because thor assumes it's getting launched with the interrupt flag cleared. UEFI runs with it enabled, though. From the spec, revision 2.10A, page 28:

> Interrupts are enabled-though no interrupt services are supported other than the UEFI boot services timer functions (All loaded device drivers are serviced synchronously by “polling.”)

kernel/eir/protos/uefi/entry.cpp
no92: They are global, and calling them `st` or `gST` is very common.

kernel/eir/protos/uefi/entry.cpp
no92: Note to self: this port is used for exfiltrating the base address of `eir-uefi` so that it can be communicated to `vm-util.py`'s `gdb --uefi` subcommand.

kernel/eir/protos/uefi/entry.cpp
no92: Thing about this is that we have no way of handling an updated modeinfo struct correctly, if that were to ever happen. The spec currently only defines version 0, so asserting seems like a fair choice?

avdgrinten: I don't have a problem with hard failures, but we should use `panicLogger()` and not `assert()` in these cases. `assert()` should be used to check for contract violations, not runtime errors.

avdgrinten: Note that this also applies to other `assert()`s in the same file (not all of them though, for example `assert()`ing that `cmdLine` is non-null is fine as the code above is supposed to guarantee this condition.

meson_options.txt
no92: We could check whether the target triplet's OS part is `windows` or `uefi`, but that seems a bit more complicated and prone to break than just having a meson_option for it.

marv7000: What about a new option for which protocol to build for? Currently, all archs assume they are always using e.g. mb2. With an extra option you could pass e.g. `uefi`, `mb2` or `limine`.

kernel/eir/protos/uefi/entry.cpp
no92: As discussed, this will be implemented later down the line.

no92:
Rebased the PR, addressed the comments raised.

avdgrinten (approved):

qookei (approved):

PR #731, opened 2024-11-01
meta: add clang-format

64:


Dennisbonke:
Implemented a while back.

PR #733, opened 2024-11-03
drivers/usb-serial: add `serial.dump` cmdline to dump kernel logs

no92:


avdgrinten:
Can we put the dumping logic into a library that all UART drivers can use? For example, also virtio-console?

avdgrinten:
Note that `virtio-console` already contains similar logic to dump other kinds of logs.

avdgrinten (approved):

PR #734, opened 2024-11-03
eir: load thor from initrd, drop multiboot1

no92:


kernel/eir/generic/main.cpp
```diff
@@ -351,6 +353,20 @@ address_t mapBootstrapData(void *p) {
...
+	auto initrd_end = reinterpret_cast<uintptr_t>(cpio_range.eof());
+	eir::infoLogger() << "Initrd ends at " << (void *)initrd_end << frg::endlog;
+
+	for(auto entry : cpio_range) {
+		if(entry.name == "thor") {
+			kernel_image = entry.data;
+		}
+	}
+
+	assert(kernel_image.data() && kernel_image.size());
```
avdgrinten, line 367: `panicLogger` instead of `assert`.

kernel/eir/arch/x86/multiboot2/multiboot2.cpp
```diff
@@ -228,7 +228,7 @@ extern "C" void eirMultiboot2Main(uint32_t info, uint32_t magic){
 	assert(mmap_start);
 	assert(mmap_end > mmap_start);
 
-	assert(n_modules >= 2);
+	assert(n_modules >= 1);
```
avdgrinten, line 231: Should we demand `== 1`? Also, this should use the `panicLogger`.

kernel/eir/arch/x86/multiboot2/multiboot2.cpp
```diff
@@ -304,7 +304,7 @@ extern "C" void eirMultiboot2Main(uint32_t info, uint32_t magic){
 		info_ptr->acpiRsdp = reinterpret_cast<uint64_t>(rsdpPtr);
 	}
 
-	info_ptr->numModules = n_modules - 1;
+	info_ptr->numModules = n_modules;
```
avdgrinten, line 307: We can probably even drop `numModules` and always just pass a single module.

PR #736, opened 2024-11-04
Enable more lints in clang-tidy

no92:


posix/subsystem/src/requests.cpp
```diff
@@ -1750,7 +1750,8 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 						co_await sendErrorResponse(managarm::posix::Errors::ALREADY_EXISTS);
 						continue;
 					}else{
-						auto fileResult = co_await tail->getTarget()->open(
+						auto target = tail->getTarget();
+						auto fileResult = co_await target->open(
```
avdgrinten, line 1754: Which lint does this fix?

posix/subsystem/src/subsystem/usb/usb.cpp
```diff
@@ -434,11 +434,12 @@ async::detached run() {
 				continue;
 
 			auto entity = co_await mbus_ng::Instance::global().getEntity(event.id);
+			auto entity_mbus_id = entity.id();
```
avdgrinten, line 437: Same here?

posix/subsystem/src/requests.cpp
avdgrinten: Discussed offline -- this is use after move.

PR #739, opened 2024-11-09
amd pcne

Lizzie841:
draft

### DIFF
--- a/drivers/nic/amdpcnet/src/pcnet.cpp
+++ b/drivers/nic/amdpcnet/src/pcnet.cpp
@@ -1,0 +1,432 @@
+#include <nic/pcnet/pcnet.hpp>
+#include <arch/dma_pool.hpp>
+#include <arch/dma_structs.hpp>
+#ifdef __x86_64__
+#include <arch/io_space.hpp>
+#endif
+#include <arch/mem_space.hpp>
+#include <arch/register.hpp>
+#include <arch/variable.hpp>
+#include <async/recurring-event.hpp>
+#include <async/oneshot-event.hpp>
+#include <helix/ipc.hpp>
+#include <helix/timer.hpp>
+#include <helix/memory.hpp>
+#include <protocols/hw/client.hpp>
+#include <netserver/nic.hpp>
+#include <core/queue.hpp>
+#include <queue>
+#include <string.h>
+
+// debug options
+namespace {
+
+constexpr bool logDriverStuff = false;
+
+}
+
+namespace {
+
+// AMD PCNET has two register sets, indexed by the same RAP32/RAP16 register
+// - RDP is for CSRs
+// - BDP is for BCRs
+//
+// Simple as that!
+inline constexpr arch::scalar_register<uint32_t> PCNET_MAC_1(0x0);
+inline constexpr arch::scalar_register<uint32_t> PCNET_MAC_2(0x4);
+
+inline constexpr arch::scalar_register<uint16_t> PCNET16_RDP(0x10);
+inline constexpr arch::scalar_register<uint16_t> PCNET16_RAP(0x12);
+inline constexpr arch::scalar_register<uint16_t> PCNET16_RST(0x14);
+inline constexpr arch::scalar_register<uint16_t> PCNET16_BDP(0x16);
+
+inline constexpr arch::scalar_register<uint32_t> PCNET32_RDP(0x10);
+inline constexpr arch::scalar_register<uint32_t> PCNET32_RAP(0x14);
+inline constexpr arch::scalar_register<uint32_t> PCNET32_RST(0x18);
+inline constexpr arch::scalar_register<uint32_t> PCNET32_BDP(0x1c);
+
+inline constexpr uint32_t MAX_FRAME_SIZE = 1520;
+
+struct Request {
+	Request(size_t size) : index(0, size) { };
+	QueueIndex index;
+	async::oneshot_event event;
+	arch::dma_buffer_view frame;
+};
+
+struct [[gnu::packed]] Descriptor {
+	uint32_t addr;
+	uint16_t length;
+	uint16_t status;
+	uint32_t msg_length;
+	uint32_t res;
+};
+static_assert(sizeof(Descriptor) == 16);
+
+struct [[gnu::packed]] InitializerDescriptor {
+	uint16_t mode;
+	uint8_t rx_len;
+	uint8_t tx_len;
+	//4
+	uint8_t mac[6];
+	uint16_t reserved_0;
+	//12
+	uint8_t ladr[8];
+	//20
+	uint32_t rx_paddr;
+	//24
+	uint32_t tx_paddr;
+};
+static_assert(sizeof(InitializerDescriptor) == 28);
+
+struct PcNetNic;
+
+template<bool IsTransmit>
+struct PcNetQueue {
+	static const uint32_t descriptor_count = IsTransmit ? 8 : 32;
+
+	std::vector<std::shared_ptr<Request>> requests;
+	arch::dma_array<Descriptor> descriptors;
+	QueueIndex next_index;
+	std::vector<arch::dma_buffer> buffers;
+
+	PcNetQueue() : next_index(0, descriptor_count) {
+
+	}
+	void setLength(uint32_t i, uint32_t len) {
+		descriptors[i].length = uint16_t((-len) & 0x0fff) | 0xf000;
+	}
+	void init(PcNetNic& nic);
+};
+
+struct PcNetNic : nic::Link {
+	PcNetNic(protocols::hw::Device dev);
+	virtual async::result<size_t> receive(arch::dma_buffer_view);
+	virtual async::result<void> send(const arch::dma_buffer_view);
+	virtual ~PcNetNic() = default;
+
+	async::result<void> init();
+	async::detached processIrqs();
+protected:
+	async::result<uint32_t> pollDevice();
+	uint32_t csr_read(uint32_t n);
+	void csr_write(uint32_t n, uint32_t m);
+	uint32_t bcr_read(uint32_t n);
+	void bcr_write(uint32_t n, uint32_t m);
+	void print_status();
+
+	arch::contiguous_pool dmaPool_;
+	protocols::hw::Device device_;
+	helix::UniqueDescriptor irq_;
+	helix::Mapping mmio_mapping_;
+	arch::mem_space mmio_;
+	arch::dma_buffer initializer_;
+	PcNetQueue<true> tx_;
+	PcNetQueue<false> rx_;
+};
+
+template<bool IsTransmit>
+void PcNetQueue<IsTransmit>::init(PcNetNic& nic) {
+	descriptors = arch::dma_array<Descriptor>(nic.dmaPool(), descriptor_count);
+	if(logDriverStuff)
+		std::cout << "drivers/pcnet: setting " << descriptor_count << " buffers at " << (void*)helix_ng::ptrToPhysical(descriptors.data()) << " of " << (IsTransmit ? "TX" : "RX") << std::endl;
+	
+	for(uint32_t i = 0; i < descriptor_count; ++i) {
+		auto buf = arch::dma_buffer(nic.dmaPool(), MAX_FRAME_SIZE);
+		memset(buf.data(), 0, buf.size());
+		uintptr_t addr = helix_ng::ptrToPhysical(buf.data());
+		uint16_t len = uint16_t(buf.size());
+		//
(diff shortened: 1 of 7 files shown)


drivers/nic/amdpcnet/src/amdpcnet.cpp
```diff
@@ -0,0 +1,445 @@
...
+#include <helix/timer.hpp>
+#include <helix/memory.hpp>
+#include <protocols/hw/client.hpp>
+#include <netserver/nic.hpp>
+#include <core/queue.hpp>
+#include <queue>
+#include <string.h>
+
+// debug options
+namespace {
```
no92, line 22: Same as above.

drivers/nic/amdpcnet/src/amdpcnet.cpp
```diff
@@ -0,0 +1,445 @@
...
+
+template<bool IsTransmit>
+struct AmdPcNetQueue {
+	static const uint32_t descriptor_count = IsTransmit ? 8 : 32;
+
+	std::vector<std::shared_ptr<Request>> requests;
+	arch::dma_array<Descriptor> descriptors;
+	QueueIndex next_index;
+	std::vector<arch::dma_buffer> buffers;
+	//
```
no92, line 90: Why is this a comment and not just an empty line?

drivers/nic/amdpcnet/src/amdpcnet.cpp
```diff
@@ -0,0 +1,445 @@
...
+	}
+	void setLength(uint32_t i, uint32_t len) {
+		descriptors[i].length = uint16_t((-len) & 0x0fff) | 0xf000;
+	}
+	void init(AmdPcNetNic& nic);
+};
+
+struct AmdPcNetNic : nic::Link {
+	AmdPcNetNic(protocols::hw::Device dev);
+	virtual async::result<size_t> receive(arch::dma_buffer_view) override;
```
no92, line 102: `virtual` and `override` one one line is unnecessary (possibly incorrect?), especially when the class isn't `final` IIRC.

drivers/nic/amdpcnet/src/amdpcnet.cpp
```diff
@@ -0,0 +1,445 @@
...
+}
+
+async::result<void> AmdPcNetNic::init() {
+	// Setup for PCI - then select the first MMIO bar
+	irq_ = co_await device_.accessIrq();
+	co_await device_.enableBusmaster();
+	auto info = co_await device_.getPciInfo();
+	size_t bar_index = 0;
+	while(true) {
+		assert(bar_index < 6 && "drivers/amdpcnet: unable to locate MMIO BAR!");
```
no92, line 179: This will fail with older PCnet models that do not provide a MMIO bar. Instead, it would be fairly easy to handle the port I/O case, especially given the way how the device does register access.

drivers/nic/amdpcnet/src/amdpcnet.cpp
```diff
@@ -0,0 +1,445 @@
...
+	mmio_.store(PCNET32_BDP, m);
+}
+
+async::result<void> AmdPcNetNic::init() {
+	// Setup for PCI - then select the first MMIO bar
+	irq_ = co_await device_.accessIrq();
+	co_await device_.enableBusmaster();
+	auto info = co_await device_.getPciInfo();
+	size_t bar_index = 0;
+	while(true) {
```
no92, line 178: This would be better replaced by a for loop; we can only ever get 6 BARs, and the hw protocols only provides as many.

drivers/nic/amdpcnet/src/amdpcnet.cpp
```diff
@@ -0,0 +1,445 @@
...
+		assert(bar_index < 6 && "drivers/amdpcnet: unable to locate MMIO BAR!");
+		if(info.barInfo[bar_index].ioType == protocols::hw::IoType::kIoTypeMemory) {
```
no92, line 180: I don't quite recall whether we should be checking this or `hostType`? CC @qookei

drivers/nic/amdpcnet/src/amdpcnet.cpp
```diff
@@ -0,0 +1,445 @@
...
+		std::cout << "drivers/amdpcnet: selected pci bar " << bar_index << std::endl;
+	}
+	auto &barInfo = info.barInfo[bar_index];
+	assert(barInfo.ioType == protocols::hw::IoType::kIoTypeMemory);
+	auto bar = co_await device_.accessBar(bar_index);
+	mmio_mapping_ = {bar, barInfo.offset, barInfo.length};
+	mmio_ = mmio_mapping_.get();
+
+	// Now we are ready to start sending shit
+	// Reset the fucker (in both modes, 16-bit and 32-bit)
```
no92, line 195: Tone.

drivers/nic/amdpcnet/src/amdpcnet.cpp
```diff
@@ -0,0 +1,445 @@
...
+	// Cleverly, RST is on +0x18 for 32-bit mode, so reading from there causes
+	// no (important) side effects! wohoo thanks pcnet i love you pcnet
```
no92, line 197: Tone.

drivers/nic/amdpcnet/src/amdpcnet.cpp
```diff
@@ -0,0 +1,445 @@
...
+	mmio_.load(PCNET32_RST);
+	mmio_.load(PCNET16_RST);
+	co_await helix::sleepFor(1'000'000); // sleep for 1 microsecond
```
no92, line 200: The sleeping part might be obvious :^)

drivers/nic/amdpcnet/src/amdpcnet.cpp
```diff
@@ -0,0 +1,445 @@
...
+	tx_.init(*this);
+
+	//
+	// RECEIVE
+	rx_.init(*this);
+
+	//
+	// INITIALIZE
+	initializer_ = arch::dma_buffer(dmaPool(), sizeof(InitializerDescriptor));
+	auto* init = (InitializerDescriptor*)initializer_.data();
```
no92, line 231: We right-align the `*`.

drivers/nic/amdpcnet/src/amdpcnet.cpp
```diff
@@ -0,0 +1,445 @@
...
+	init->mode = 0;
+	init->reserved_0 = 0;
+	init->ladr[0] = 0;
```
no92, line 234: Seems like a perfect place for a `memset`, or even better, `std::fill`. There is no padding to worry here, the struct is packed.

drivers/nic/amdpcnet/src/amdpcnet.cpp
```diff
@@ -0,0 +1,445 @@
...
+	init->ladr[1] = 0;
+	init->ladr[2] = 0;
+	init->ladr[3] = 0;
+	init->ladr[4] = 0;
+	init->ladr[5] = 0;
+	init->ladr[6] = 0;
+	init->ladr[7] = 0;
+	//
+	init->mac[0] = mac_[0];
```
no92, line 243: `memecpy` would do the job here, too.

drivers/nic/amdpcnet/src/amdpcnet.cpp
```diff
@@ -0,0 +1,445 @@
...
+
+	//
+	// Initialize step
+	if(logDriverStuff) {
+		std::cout << "drivers/amdpcnet: step-initialize" << std::endl;
+	}
+	csr_write(0, (1 << 0) | (1 << 6));
+
+	if(logDriverStuff) {
+		std::cout << "drivers/amdpcnet: 2434234" << std::endl;
```
no92, line 287: Is this a debugging leftover?

drivers/nic/amdpcnet/src/amdpcnet.cpp
```diff
@@ -0,0 +1,445 @@
...
+	}
+
+	processIrqs();
+}
+
+AmdPcNetNic::AmdPcNetNic(protocols::hw::Device device)
+	: nic::Link(1500, &dmaPool_), device_ { std::move(device) }
+{
+	//promiscuous_ = true;
```
no92, line 296: Why are these commented out?

drivers/nic/amdpcnet/src/amdpcnet.cpp
```diff
@@ -0,0 +1,445 @@
...
+	uint64_t sequence = 0;
+	while(true) {
+		auto await = co_await helix_ng::awaitEvent(irq_, sequence);
+		HEL_CHECK(await.error());
+		sequence = await.sequence();
+
+		__sync_synchronize();
+
+		if(logDriverStuff) {
+			std::cout << "drivers/amdpcnet: got the fucking irq shit #" << sequence << "!!!" << std::endl;
```
no92, line 360: Tone.

drivers/nic/amdpcnet/src/amdpcnet.cpp
```diff
@@ -0,0 +1,445 @@
...
+			new_csr0 |= (1 << 1); //set start bit
+			new_csr0 |= (1 << 8); //mask off IDON
+		}
+		new_csr0 |= (1 << 6);
+		csr_write(0, new_csr0);
+
+		if(new_csr0 != csr0 && logDriverStuff) {
+			std::cout << "drivers/amdpcnet: CSR0(old)" << (void*)(uintptr_t)csr0 << " != (new)" << (void*)(uintptr_t)new_csr0 << " !!!" << std::endl;
+		}
+		if(logDriverStuff) {
```
no92, line 420: Might be nice to put this in a separate function so this can be reused.

drivers/nic/amdpcnet/src/amdpcnet.cpp
no92: The port I/O case would need to be `#ifdef`ed away to only x86 though.

servers/netserver/meson.build
```diff
@@ -19,6 +19,7 @@ dep = [ fs_proto_dep, mbus_proto_dep, svrctl_proto_dep, nic_virtio_dep,
 
 if host_machine.cpu_family() == 'x86_64'
 	dep += nic_freebsd_e1000_dep
+	dep += nic_amdpcnet_dep
```
no92, line 22: Is this necessarily limited to x86_64? This should work just fine on any other (little-endian) arch, right?

no92 (changes requested):
Had a look over the code, haven't checked against the (surprisingly long) spec yet. While I can't boot managarm on VirtualBox, this currently does not work in VMWare due to the issues pointed out below. QEMU works fine for me, though.

A few general notes:
- Single-line if bodies can omit the curly braces
- after opening/before closing a namespace, we generally try to place a new line
- the comments' tone should be suitable for upstreaming :^)

I'll check this against the spec when I'm not as tired as now.

servers/netserver/meson.build
Nivaturimika: sure moved it above

drivers/nic/amdpcnet/src/amdpcnet.cpp
Nivaturimika: added print_status()

no92 (changes requested):
Generally, this looks quite decent, bar some small comments. I don't quite have the time to do a full pass on the spec right now, so take this as the first chunk of that.

drivers/nic/amdpcnet/src/pcnet.cpp
```diff
@@ -0,0 +1,432 @@
...
+struct [[gnu::packed]] Descriptor {
+	uint32_t addr;
+	uint16_t length;
+	uint16_t status;
+	uint32_t msg_length;
+	uint32_t res;
+};
+static_assert(sizeof(Descriptor) == 16);
+
+struct [[gnu::packed]] InitializerDescriptor {
```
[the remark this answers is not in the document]
Nivaturimika: it is set implicitly on reset

drivers/nic/amdpcnet/src/pcnet.cpp
```diff
@@ -0,0 +1,432 @@
...
+
+	// TODO: The kick here should not be required.
+	HEL_CHECK(helAcknowledgeIrq(irq_.getHandle(), kHelAckKick, 0));
+	uint64_t sequence = 0;
+	while(true) {
+		auto await = co_await helix_ng::awaitEvent(irq_, sequence);
+		HEL_CHECK(await.error());
+		sequence = await.sequence();
+
+		__sync_synchronize();
```
qookei, line 340: Why is this done?

PR #741, opened 2024-11-11
Let x86_64 variants of Eir call into generic eirMain()

avdgrinten:
This is a first step towards unifying different variants of Eir.


no92 (approved):
LGTM, locally tested in the UEFI case and virtual boot (both BIOS and UEFI)

PR #744, opened 2024-11-13
eir: Add limine protocol

marv7000:
This PR adds Limine support to eir.
Depends on https://github.com/managarm/bootstrap-managarm/pull/405


kernel/eir/protos/limine/entry.cpp
```diff
@@ -0,0 +1,152 @@
...
+extern "C" void eirLimineMain(void){
+	eirRunConstructors();
+
+	log_e9 = true; // TODO
+
+	assert(kernel_file_request.response);
+	assert(module_request.response);
+	assert(module_request.response->module_count > 0);
+
+	// Reserve region for the kernel
```
no92, line 132: We can probably just remove these commented lines.

kernel/eir/protos/limine/entry.cpp
```diff
@@ -0,0 +1,152 @@
...
+		info_ptr->moduleInfo = mapBootstrapData(initrd_module);
+	}
+};
+
+} // namespace
+
+extern "C" void eirLimineMain(void){
+	eirRunConstructors();
+
+	log_e9 = true; // TODO
```
no92, line 126: This should be dropped or replaced with cmdline parsing (see UEFI for how that would work).

kernel/eir/protos/limine/entry.cpp
```diff
@@ -0,0 +1,152 @@
...
+};
+
+initgraph::Task setupMemoryRegions{&globalInitEngine,
+	"limine.setup-memory-regions",
+	initgraph::Requires{getReservedRegionsKnownStage()},
+	initgraph::Entails{getMemoryRegionsKnownStage()},
+	[] {
+		assert(memmap_request.response);
+
+		cmdline = frg::string_view {kernel_file_request.response->kernel_file->cmdline};
```
no92, line 86: If we want to do command line parsing early on, this would need to be moved into the main function.

no92 (commented):
First, quick pass. Separating the multiboot2 and `general*_link.x` changes out into a separate commit would be nice, and squashing the rest into one, too. Switching all physical addresses to the `physaddr_t` type would be nice but IMO doesn't have to happen in this PR.

no92 (approved):
LGTM

PR #746, opened 2024-11-14
Generalize the ASID and page table manipulation code

qookei:


kernel/thor/generic/thor-internal/arch-generic/asid.hpp
```diff
@@ -0,0 +1,297 @@
...
+}
+
+
+// Switch to given page table on the given ASID. Potentially
+// invalidate the TLB entries for the ASID that's being used.
+void switchToPageTable(PhysicalAddr root, int asid, bool invalidate);
+
+// Switch away from the current user page tables and invalidate the
+// given ASID.
+void switchAwayFromPageTable(int asid);
```
avdgrinten, line 281:
```suggestion
// Switch away from the current user page tables to kernel-only page tables.
// Also invalidate the given ASID.
// This is called when the currently active page tables need to be destroyed.
void switchAwayFromPageTable(int asid);
```

avdgrinten: Also, is there a difference in efficiency compared to switching without invalidation (i.e., changing the prototype to `switchAwayFromPageTable()`) and calling `invalidateAsid()` afterwards? That may make the per-arch implementation a bit simpler.

kernel/thor/arch/arm/paging.cpp
```diff
@@ -991,7 +234,8 @@ bool ClientPageSpace::updatePageAccess(VirtualAddr pointer) {
 	tbl3[index3].store(bits);
 
 	// TODO: perform proper shootdown to update mapping
-	invalidatePage(reinterpret_cast<void *>(pointer));
+	// FIXME: wrong ASID, but it was already broken...
+	invalidatePage(0, reinterpret_cast<void *>(pointer));
```
avdgrinten, line 238: Isn't this just the current active ASID? Since this function is called on page fault to ensure that the true permissions match the permissions that we want.

Also, the previous comment said that this needs to perform shootdown, but is that really true? The only transition in this function is RO -> RW, so the worst that can happen is that some other cores fault because they think that the page is still RO (unless aarch64 re-reads the tables upon a fault?).

Note that the other direction (set RO) does need shootdown.

avdgrinten: This function is also underdocumented. It would be nice to add a comment to state what it does / what the return values mean.

kernel/thor/arch/x86/paging.cpp
```diff
@@ -3,688 +3,136 @@
...
+void invalidateAsid(int asid) {
+	if(asid == globalBindingId || !getCpuData()->havePcids) {
+		assert(asid == globalBindingId || asid == 0);
 
-void GlobalPageBinding::bind() {
-	assert(!intsAreEnabled());
+		uint64_t pml4;
+		asm volatile ("mov %%cr3, %0" : "=r"(pml4));
+		pml4 &= 0x000FFFFFFFFFF000; // Clear the first bit to invalidate the PCID.
+		asm volatile ("mov %0, %%cr3" : : "r"(pml4) : "memory");
```
avdgrinten, line 35: This case looks strange to me.

1. If PCIDs are not enabled, clearing bits is not necessary. Even in the PCID case, I don't think one can read by the invalidate bit.
2. For the global binding, this looks wrong since global pages are not invalidated by writing to CR3 (not matter whether the invalidation bit is set). Do we even ever need to invalidate the global binding?

kernel/thor/arch/x86/paging.cpp
```diff
@@ -3,688 +3,136 @@
...
 }
 
-KernelPageSpace::KernelPageSpace(PhysicalAddr pml4_address)
-: _rootTable{pml4_address} { }
-
-bool KernelPageSpace::submitShootdown(ShootNode *node) {
-	assert(!(node->address & (kPageSize - 1)));
-	assert(!(node->size & (kPageSize - 1)));
+// TODO(qookie): Can't we raise this?
+static constexpr int maxPcidCount = 8;
```
avdgrinten, line 103: No action needed for this PR, but one consideration here is that we sometimes loop over all PCIDs and that becomes inefficient if we make use of too many PCIDs. We probably have to improve the algorithms a bit to avoid looping before we enable a lot of PCIDs.

kernel/thor/arch/x86/paging.cpp
qookei: > Even in the PCID case, I don't think one can read by the invalidate bit.

That's true IIRC. This implementation was just copied from the previous `invalidateFullTlb()` function.

> Do we even ever need to invalidate the global binding?

It seems I missed the fact that GlobalPageBinding::shootdown and KernelPageSpace::submitShootdown did not do the optimization for large shootdowns and instead always invalidated them page-by-page. Fixing that means we can just `assert(asid != globalBindingId);` here.

kernel/thor/generic/asid.cpp
```diff
@@ -0,0 +1,296 @@
...
+		auto irqLock = frg::guard(&irqMutex());
+		auto lock = frg::guard(&mutex_);
+
+		auto unshotBindings = numBindings_;
+
+		auto &globalBinding = getCpuData()->asidData->globalBinding;
+		auto &bindings = getCpuData()->asidData->bindings;
+
+		// Perform synchronous shootdown.
+		if(globalBinding.boundSpace().get() == this) {
```
avdgrinten, line 267: Since this breaks if the same space is bound to `globalBinding` and one of the `bindings` we should add assertions that verify that this never happens.

kernel/thor/generic/thor-internal/arch-generic/cursor.hpp
```diff
@@ -0,0 +1,242 @@
...
+		return doRealizeLevel_(accessors_[level], accessors_[level - 1], level - 1);
+	}
+
+	static void doRealizePts_(void *ctxt) {
+		static_cast<PageCursor *>(ctxt)->realizeLevel_(lastLevel);
+	}
+
+	void realizePts_() {
+		detail::runWithLockedSpace(space_, doRealizePts_, this);
+	}
```
avdgrinten, line 231: This virtual call should be avoided if possible.

kernel/thor/generic/thor-internal/arch-generic/cursor.hpp
```diff
@@ -0,0 +1,242 @@
...
+
+		auto subPtPtr = Policy::pteTableAddress(ptEnt);
+		subPt = PageAccessor{subPtPtr};
+		return true;
+	}
+
+	bool reloadLevel_(size_t level) {
+		if(accessors_[level]) /*[[likely]]*/
+			return true;
+		if(level - 1 != initialLevel_ && !reloadLevel_(level - 1))
```
avdgrinten, line 193: `assert(level != initialLevel_)` instead of first branch.

avdgrinten (approved):

PR #748, opened 2024-11-15
eir: UEFI improvements

no92:
Depends on managarm/bootstrap-managarm#407.


kernel/eir/arch/x86/arch.cpp
```diff
@@ -189,6 +189,9 @@ void initProcessorPaging(void *kernel_start, uint64_t &kernel_entry) {
 	eir::infoLogger() << "eir: Allocated " << (allocatedMemory >> 10) << " KiB"
 			" after setting up paging" << frg::endlog;
 
+	// PE doesn't support linker scripts, this needs to be worked around by UEFI
+	// see the `uefi.map-eir-image` task
+#if !defined(EIR_NATIVE_PE) || EIR_NATIVE_PE == 0
```
avdgrinten, line 194: Testing both `defined(EIR_NATIVE_PE)` and `EIR_NATIVE_PE == 0` seems excessive for a bool flag. Why not just test whether it is defined or not?

kernel/eir/generic/elf_init.cpp
```diff
@@ -1,5 +1,7 @@
 #include <eir-internal/debug.hpp>
 
+#if !defined(EIR_NATIVE_PE) || EIR_NATIVE_PE == 0
```
avdgrinten, line 3: Instead putting `#if`s around `eirRunConstructors()`, we can just put the PE code also into this file (and rename it). That way, `eirRunConstructors()` always does the right thing (no matter if you're compiling to PE or not).

kernel/eir/protos/uefi/meson.build
```diff
@@ -3,10 +3,56 @@ eir_uefi_sources = [
...
+	]
+	eir_uefi_elf_name = 'eir-uefi'
+	eir_uefi_linker_script = files()
+else
+	eir_uefi_cpp_args += ['-DEIR_NATIVE_PE=0']
+	eir_uefi_ld_args += [
+		'-Wl,-entry,eirUefiMain',
+		'-Wl,--defsym=PECOFF_HEADER_SIZE=' + meson.get_external_property('pecoff_header_size').to_string()
+	]
+	eir_uefi_elf_name = 'eir-uefi-debug'
```
avdgrinten, line 28: Why is this called `-debug`? Maybe `-intermediate-elf` or something similar that makes it clear that this is not the final binary?

Also `eir_uefi_elf_name` -> `eir_uefi_exe_name` since it's not an ELF in the native case?

kernel/eir/protos/uefi/meson.build
```diff
@@ -3,10 +3,56 @@ eir_uefi_sources = [
 	'helpers.cpp',
 ]
 
-executable('eir-uefi', eir_uefi_sources + eir64_sources, eir_generic_main,
+eir_uefi_cpp_args = ['-ffreestanding', '-fshort-wchar', '-fno-stack-protector']
+eir_uefi_ld_args = ['-nostdlib']
+eir_uefi_link_depends = []
+
+native_pe_toolchain = meson.get_external_property('native_pe_toolchain')
```
avdgrinten, line 10: Please add a comment on why somebody would want to build a non-native PE.

kernel/eir/protos/uefi/non-pe-native-link.x
```diff
@@ -0,0 +1,92 @@
+ENTRY(eirUefiMain)
+
+SECTIONS {
+	. = PECOFF_HEADER_SIZE;
+
+	eirImageFloor = .;
```
avdgrinten, line 6: If symbols are not supported, we can drop this (and also eirImageCeiling).

avdgrinten (approved):

PR #749, opened 2024-11-15
Make logging re-entrant

avdgrinten:
This overhauls the logging infrastructure to allow re-entrant logging. In particular, this enables logging from NMI, MCE and other contexts that cannot be masked. To achieve this, we add a mechanism for code sequences that can be interrupted (and cancelled) by NMIs and similar exceptions. The new mechanism is similar to the existing user access region (UAR) infrastructure. We use the new mechanism to implement an interruptible memcpy() to a per-CPU logging ring buffer.

### DIFF
--- a/kernel/thor/arch/arm/thor-internal/arch/debug.hpp
+++ b/kernel/thor/arch/arm/thor-internal/arch/debug.hpp
@@ -5,7 +5,12 @@
 namespace thor {
 
 struct UartLogHandler : public LogHandler {
+	constexpr UartLogHandler() {
+		takesUrgentLogs = true;
+	}
+
 	void emit(Severity severity, frg::string_view msg) override;
+	void emitUrgent(Severity severity, frg::string_view msg) override;
 
 	void printChar(char c);
 };
--- a/kernel/thor/generic/debug.cpp
+++ b/kernel/thor/generic/debug.cpp
@@ -1,6 +1,7 @@
 #include <thor-internal/cpu-data.hpp>
 #include <thor-internal/debug.hpp>
 #include <thor-internal/arch/stack.hpp>
+#include <thor-internal/ring-buffer.hpp>
 
 namespace thor {
 
@@ -18,6 +19,12 @@ namespace {
 	>> globalLogList;
 } // anonymous namespace
 
+void LogHandler::emitUrgent(Severity severity, frg::string_view msg) {
+	if (!takesUrgentLogs)
+		panic();
+	emit(severity, msg);
+}
+
 void enableLogHandler(LogHandler *sink) {
 	if (!globalLogList)
 		globalLogList.initialize();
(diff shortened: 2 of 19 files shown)


kernel/thor/arch/arm/thor-internal/arch/debug.hpp
```diff
@@ -5,6 +5,10 @@
 namespace thor {
 
 struct UartLogHandler : public LogHandler {
+	constexpr UartLogHandler() {
+		takesUrgentLogs = true;
+	}
+
```
qookei, line 11: No `emitUrgent` override like in `PIOLogHandler`? Or alternatively, why have an override in `PIOLogHandler` if it only adds an `URGENT: ` prefix?

kernel/thor/generic/debug.cpp
```diff
@@ -34,9 +41,140 @@ void disableLogHandler(LogHandler *sink) {
...
+		return s & CpuData::RS_EMITTING;
+	}
+
+	// Assumption: !intsAreEnabled().
+	bool tryStartEmitting() {
+		auto *cpuData = getCpuData();
+		unsigned int s = cpuData->reentrantLogState.load(std::memory_order_relaxed);
+		while (true) {
+			if (s) {
+				bool cas = cpuData->reentrantLogState.compare_exchange_strong(
```
qookei, line 95: If we're looping based on failure maybe we could use `compare_exchange_weak`? Would lead to slightly better code gen on AArch64.

kernel/thor/generic/debug.cpp
```diff
@@ -49,11 +187,17 @@ namespace {
 			};
 
 			auto emit = [&] (char c) {
+				if (!stagedLength) {
+					// Put log metadata in front of actual log message.
+					LogMetadata md{.severity = severity};
+					memcpy(stagingBuffer, &md, sizeof(LogMetadata));
+					stagedLength = sizeof(LogMetadata);
```
qookei, line 194: Should the staging buffer perhaps be made `sizeof(LogMetadata) + logLineLength` bytes long?

kernel/thor/generic/debug.cpp
```diff
@@ -34,9 +41,140 @@ void disableLogHandler(LogHandler *sink) {
...
+
+		// Only start emitting logs if we are not a reentrant context.
+		if (!tryStartEmitting())
+			return;
+
+		do {
+			while (true) {
+				auto lock = frg::guard(&logMutex);
+
+				char buffer[128];
```
qookei, line 60: Shouldn't this be the same size as the staging buffer in `LogProcessor`? As this is currently, since `logLineLength` is 256, we could potentially dequeue a partial line and lose the rest of it, no?

kernel/thor/generic/thor-internal/ring-buffer.hpp
```diff
@@ -229,4 +229,123 @@ struct SingleContextRecordRing {
...
+
+	size_t effectiveSize(size_t recordSize) {
+		return (headerSize + recordSize + recordAlign - 1) & ~(recordAlign - 1);
+	}
+
+	int shift_ = 16;
+	char buffer_[1 << 16];
+	// tailPtr_ and headPtr_ may be modified by reentrant contexts. Access through atomic_ref.
+	uint64_t tailPtr_{0};
+	uint64_t headPtr_{0};
```
qookei, line 348: If this is not accessed from assembly or anything, why not just use `std::atomic<uint64_t>` directly?

kernel/thor/generic/debug.cpp
avdgrinten: We could do that; however 256 chars should be plenty even if there is a small header. In the future, I'd like to merge `LogProcessor` and the logger itself (such that the logger logs directly to the staging buffer). Then this problem also goes away.

kernel/thor/generic/thor-internal/ring-buffer.hpp
avdgrinten: A pointer to this is passed to `iseqStore64`, so we can't just use `std::atomic` (or we'd have to assume that the memory layout of `atomic<uint64_t>` matches `uint64_t`).

PR #534, opened 2023-07-09
RISC-V support (super-duper ultra WIP)

ElectrodeYT:
Technically links, but everything is currently a assert

PR #752, opened 2024-11-16
Add RISC-V support to thor

avdgrinten:
This PR makes thor compile, even though it (obviously) does not work yet.

We also use clang-format for the newly added files in `kernel/thor/arch/riscv` and add the appropriate configuration files for that purpose.

Original code by @ElectrodeYT, rebased by @no92 and rebased again by me.

Supersedes #534.

### DIFF
--- a/kernel/thor/arch/riscv/debug.cpp
+++ b/kernel/thor/arch/riscv/debug.cpp
@@ -1,0 +1,34 @@
+#include <arch/mem_space.hpp>
+#include <arch/register.hpp>
+#include <thor-internal/arch/debug.hpp>
+
+namespace thor {
+
+constinit FirmwareLogHandler firmwareLogHandler;
+
+void setupDebugging() { enableLogHandler(&firmwareLogHandler); }
+
+void FirmwareLogHandler::emit(Severity severity, frg::string_view msg) {
+	(void)severity;
+	for (size_t i = 0; i < msg.size(); ++i)
+		printChar(msg[i]);
+	printChar('\n');
+}
+
+void FirmwareLogHandler::sbiCall1(int ext, int func, SbiWord arg0) {
+	register SbiWord rExt asm("a7") = ext;
+	register SbiWord rFunc asm("a6") = func;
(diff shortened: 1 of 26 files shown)


kernel/thor/arch/riscv/link.x
```diff
@@ -0,0 +1,87 @@
+ENTRY(thorRtEntry)
+
+SECTIONS {
+	/* thor starts a bit higher than normal, to make space for eir on EFI systems. */
+	. = 0xffffffff90000000;
```
no92, line 5: Is this necessary? UEFI eir is identity mapped, so it shouldn't interfere with thor no matter what.

kernel/thor/arch/riscv/debug.cpp
```diff
@@ -0,0 +1,34 @@
+#include <arch/aarch64/mem_space.hpp>
```
marv7000, line 1: Is that on purpose?

kernel/thor/arch/riscv/link.x
avdgrinten: Good point. This is a residue from Electrode's PR which used a different `eir`. We should be able to change it back.

kernel/thor/arch/riscv/thor-internal/arch/stack.hpp
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#include <stdint.h>
+
+namespace thor {
+
+template <typename F>
+inline void walkThisStack(F) {
+	// TODO
```
no92, line 9: Should this use the `unimplementedOnRiscv` stub too?

avdgrinten: This is an optional debugging function that aarch64 also doesn't implement.

marv7000 (approved):
LGTM

kernel/thor/arch/riscv/link.x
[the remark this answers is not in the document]
avdgrinten: Good point, I'll add that when we start adding proper functionality.

PR #753, opened 2024-11-17
eir/riscv: Add ISA checks for RVA22 compliance

marv7000:
This PR adds a check in `initProcessorEarly` for riscv64 to confirm that the machine we're running on is actually RVA22 compliant and can execute thor.


kernel/eir/arch/riscv/cpu.cpp
```diff
@@ -1,8 +1,80 @@
...
+#include <eir-internal/debug.hpp>
 
 namespace eir {
 
+namespace {
+
+// All bits need to be enabled to ensure RVA22 compliance
+struct RiscVBits {
+	bool i, m, a, f, d, c, zicsr, zicntr, ziccif, ziccrse, ziccamoa, zicclsm, za64rs, zihpm, zihintpause, zic64b, zicbom, zicbop, zicboz;
+};
```
avdgrinten, line 13: I think it would be nicer to use an `enum` to assign an index to these extensions. Then we can use a boolean array with `n` entries instead of `n` separate `bool`s.

That would also simplify the `if (!(bits.i && bits.m && bits.a && bits.f && bits.d && bits.c && bits.zicsr && bits.zicntr && bits.ziccif && bits.ziccrse && bits.ziccamoa && bits.zicclsm && bits.za64rs && bits.zihpm && bits.zihintpause && bits.zic64b && bits.zicbom && bits.zicbop && bits.zicboz))` (e.g., you could use a loop or `std::all_of` to check against a static array of extensions needed for RVA22).

kernel/common/dtb.hpp
```diff
@@ -192,6 +194,17 @@ struct DeviceTreeProperty {
...
 
+	frg::optional<frg::string_view> asString(size_t index = 0) {
+		size_t total = 0;
+		const char* off = reinterpret_cast<const char*>(data_.data());
+		for (size_t i = 0; i < index; i++) {
+			total += strlen(off + total) + 1;
+			if (total >= data_.size())
+				return frg::null_opt;
+		}
+		return frg::string_view{off + total};
```
no92, line 205: This does not do a bounds check - I guess we'd need to do something like `strnlen(off + total, data_.size() - total)`.

kernel/common/dtb.hpp
```diff
@@ -192,6 +194,17 @@ struct DeviceTreeProperty {
 		return v.load();
 	}
 
+	frg::optional<frg::string_view> asString(size_t index = 0) {
+		size_t total = 0;
+		const char* off = reinterpret_cast<const char*>(data_.data());
```
no92, line 199: We right-align pointers :^)

kernel/eir/arch/riscv/cpu.cpp
```diff
@@ -1,8 +1,80 @@
...
+	if (dtb) {
+		DeviceTree dt(physToVirt<void>(dtb));
+
+		// Get the first "/cpus/cpu@..."
+		frg::optional<DeviceTreeNode> cpuNode;
+		dt.rootNode().discoverSubnodes([](auto node) {
+			return frg::string_view(node.name()) == "cpus";
+		}, [&](auto cpus) {
+			cpus.discoverSubnodes([](auto node) {
+				return frg::string_view(node.name()).starts_with("cpu@");
```
no92, line 31: This logic is probably fine for now where we'll assume all cores support the same set of base extensions, but it might be nice to look for the correct one using the HART ID that gets provided by the boot protocol somehow - on UEFI, it's a configuration table (`RISCV_EFI_BOOT_PROTOCOL`, GUID `{ 0xccd15fec, 0x6f73, 0x4eec, { 0x83, 0x95, 0x3e, 0x69, 0xe4, 0xb9, 0x40, 0xbf } }`)

kernel/eir/arch/riscv/cpu.cpp
```diff
@@ -1,8 +1,80 @@
...
+			if (extension == "zic64b") bits.zic64b = true;
+			if (extension == "zicbom") bits.zicbom = true;
+			if (extension == "zicbop") bits.zicbop = true;
+			if (extension == "zicboz") bits.zicboz = true;
+		}
+		if (!(bits.i && bits.m && bits.a && bits.f && bits.d && bits.c &&
+			bits.zicsr && bits.zicntr && bits.ziccif && bits.ziccrse &&
+			bits.ziccamoa && bits.zicclsm && bits.za64rs && bits.zihpm &&
+			bits.zihintpause && bits.zic64b && bits.zicbom && bits.zicbop && bits.zicboz)) {
+			panicLogger() << "Processor doesn't support all RVA22 extensions!" << frg::endlog;
```
no92, line 75: `required`

PR #754, opened 2024-11-17
eir: replace previous RISC-V code with UEFI stubs

no92:
This is enough to get past UEFI's ExitBootServices() call, but RISC-V initialization is unimplemented for now.


no92:
Rebased and updated the changes to make use of the recently merged RISC-V additions.

PR #760, opened 2024-11-21
Assorted bugfixes from udev porting

no92:


no92:
Comment addressed, requesting re-review.

Dennisbonke (approved):
LGTM

posix/subsystem/src/requests.cpp
```diff
@@ -2935,14 +2936,23 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+
+			if(req.fd() == -1) {
+				auto file = createSignalFile(req.sigset(),
+						req.flags() & managarm::posix::OpenFlags::OF_NONBLOCK);
+				auto fd = self->fileContext()->attachFile(file,
+						req.flags() & managarm::posix::OpenFlags::OF_CLOEXEC);
+				resp.set_fd(fd);
+			} else {
+				auto file = self->fileContext()->getFile(req.fd());
+				assert(file);
```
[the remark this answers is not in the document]
avdgrinten: This was marked as fixed but it's not actually fixed.

posix/subsystem/src/signalfd.cpp
```diff
@@ -9,71 +9,59 @@
...
+	memcpy(data, &si, sizeof(struct signalfd_siginfo));
+	co_return sizeof(struct signalfd_siginfo);
+}
 
-	helix::BorrowedDescriptor getPassthroughLane() override {
-		return _passthrough;
-	}
+async::result<frg::expected<Error, PollWaitResult>>
+OpenFile::pollWait(Process *process, uint64_t inSeq, int pollMask,
+		async::cancellation_token cancellation) {
```
avdgrinten, line 44: A proper implementation needs to take `_mask` into account in `pollWait()`. For example, we could poll for any signal but only return once a signal in `_mask` is active.

posix/subsystem/src/process.cpp
```diff
@@ -1222,26 +1222,26 @@ async::result<void> Process::terminate(TerminationState state) {
 	parent->signalContext()->issueSignal(SIGCHLD, info);
 }
 
-async::result<int> Process::wait(int pid, bool nonBlocking, TerminationState *state, ResourceUsage *stats) {
+async::result<std::shared_ptr<Process>> Process::wait(int pid, bool nonBlocking, TerminationState *state, ResourceUsage *stats) {
```
avdgrinten: Mentioned on Discord: we should probably return the siginfo instead (e.g., by adding it to `TerminationState`).

posix/subsystem/src/process.cpp
```diff
@@ -1222,8 +1222,10 @@ async::result<void> Process::terminate(TerminationState state) {
 	parent->signalContext()->issueSignal(SIGCHLD, info);
 }
 
-async::result<std::shared_ptr<Process>> Process::wait(int pid, bool nonBlocking, TerminationState *state, ResourceUsage *stats) {
+async::result<std::shared_ptr<Process>> Process::wait(int pid, WaitFlags flags, TerminationState *state, ResourceUsage *stats) {
 	assert(pid == -1 || pid > 0);
+	assert(flags & ~waitNonBlocking);
+	assert(!(flags & ~(waitNonBlocking | waitExited | waitStoppedBySignal)));
```
avdgrinten, line 1228: The first assertion seems broken?

posix/subsystem/src/process.cpp
```diff
@@ -1232,6 +1234,11 @@ async::result<std::shared_ptr<Process>> Process::wait(int pid, bool nonBlocking,
 		for(auto it = _notifyQueue.begin(); it != _notifyQueue.end(); ++it) {
 			if(pid > 0 && pid != it->pid())
 				continue;
+			if(std::get_if<TerminationByExit>(&it->_state) && !(flags & (waitExited | waitNonBlocking)))
+				continue;
+			if(std::get_if<TerminationBySignal>(&it->_state) && !(flags & (waitStoppedBySignal | waitNonBlocking)))
+				continue;
```
avdgrinten, line 1240: Why does `waitNonBlocking` affect which processes we return?

posix/subsystem/src/process.cpp
no92: The first assert ensures that there are flags set besides `waitNonBlocking`. Combined with the second assert it ensures that at least one state change type has been requested.

posix/subsystem/src/signalfd.cpp
no92: As discussed on Discord, this needs more work, including in libasync, so this is out of scope for this specific PR.

PR #767, opened 2024-11-24
thor: Add a generic mechanism for declaring CPU-local variables

qookei:
With this mechanism new code (both generic and architecture-specific) can add new CPU-local variables by just doing:

extern PerCpu\<MyType\> myLocal;
THOR_DEFINE_PERCPU(myLocal);

instead of having to add the variable as a member to either PlatformCpuData or CpuData.

This should hopefully help untangle the header dependencies for cpu-data.hpp a bit, and help move some large stuff away from the head of CpuData (e.g. GDT and IDT on x86_64).

CpuData (and {Assembly,Platform}CpuData) is still kept around for stuff we want at/near the start of the CPU-local data (e.g. the self pointer, stuff accessed from assembly code, hot members). But for most things, the new mechanism should be used.

### DIFF
--- a/kernel/thor/generic/cpu-data.cpp
+++ b/kernel/thor/generic/cpu-data.cpp
@@ -1,0 +1,84 @@
+#include <thor-internal/cpu-data.hpp>
+#include <thor-internal/physical.hpp>
+#include <thor-internal/arch-generic/paging.hpp>
+#include <thor-internal/kasan.hpp>
+#include <thor-internal/elf-notes.hpp>
+
+namespace thor {
+
+// Define a ELF note so that eir can find the per-CPU region and map
+// the KASAN shadow for it.
+
+// HACK: We define a different struct than eir's PerCpuRegion because
+// `reinterpret_cast<uint64_t>(&symbol)` is not constinit. We can't
+// just use `void *` in the eir struct because eir might be a 32-bit
+// binary. The two structs should have the same layout though, since
+// thor is always 64-bit.
+
+// Keep in sync with <eir/interface.hpp>!
+struct OurPerCpuRegion {
+	void *perCpuStart;
+	void *perCpuEnd;
+};
+
+extern ManagarmElfNote<OurPerCpuRegion> perCpuRegionNote;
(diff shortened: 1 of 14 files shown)


kernel/thor/generic/cpu-data.cpp
```diff
@@ -0,0 +1,63 @@
...
+	// CPUs in parallel.
+	curPos = reinterpret_cast<void *>(base + size);
+	numCpus++;
+
+	for(size_t pg = 0; pg < size; pg += kPageSize) {
+		auto page = physicalAllocator->allocate(kPageSize);
+		assert(page != PhysicalAddr(-1) && "OOM");
+
+		KernelPageSpace::global().mapSingle4k(base + pg, page, page_access::write, CachingMode::null);
+	}
```
avdgrinten, line 51: KASAN mappings need to be updated here.

kernel/thor/generic/cpu-data.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+	auto context = reinterpret_cast<void *>(base);
+	initializePerCpuDataFor(context);
+
+	return context;
+}
+
+size_t getCpuCount() {
+	return numCpus;
+}
```
avdgrinten, line 61: Isn't this off by one? `numCpus` is initialized to 0 and only incremented in `addNewPerCpuData().

kernel/thor/generic/thor-internal/cpu-data.hpp
```diff
@@ -64,13 +66,102 @@ struct CpuData : public PlatformCpuData {
...
+	inline PerCpu<Type> Name					\
+
+
+// Define a per-CPU variable without an initializer. Care has to be
+// taken to call Name.initializeInContext(context) prior to accessing
+// it from the given context. This is mainly intended for
+// architecture-specific fields that have to be initialized prior to
+// the allocator being available.
+#define DEFINE_PERCPU_UNINIT(Name, Type)		\
+	DEFINE_PERCPU_UNINIT_PRIV(Name, Type, "")	\
```
avdgrinten, line 136: Nitpick: I'd prefer `DEFINE_PERCPU_UNINIT` -> `DEFINE_PERCPU_UNINITIALIZED`.

avdgrinten (approved):

PR #782, opened 2024-12-08
thor: Support Sstc timer on RISC-V

avdgrinten:
The Sstc extension allows S-mode to set a timer comparison CSR. This avoids the need to call into SBI. In fact, some boards implement Sstc but not the TIME extension for SBI.

### DIFF
--- a/kernel/klibc/eir/interface.hpp
+++ b/kernel/klibc/eir/interface.hpp
@@ -66,6 +66,8 @@ enum class RiscvExtension {
 	f,
 	i,
 	m,
+	// S extensions.
+	sstc,
 	// Z extensions.
 	za64rs,
 	zic64b,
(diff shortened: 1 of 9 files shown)


kernel/klibc/eir/interface.hpp
```diff
@@ -64,7 +147,9 @@ constexpr unsigned int memoryLayout = 0x1000'0000;
 // 0x11xx'xxxx range reserved for arch-specific notes in Thor.
 // 0x1100'0xxx range reserved for x86.
 // 0x1100'1xxx range reserved for aarch64.
+// 0x1100'2xxx range reserved for aarch64.
```
qookei, line 153: riscv64?

avdgrinten: Good catch.

PR #784, opened 2024-12-10
kernel/eir: Use generic main on AArch64

Qwinci:
Use generic eir main on aarch64 + share dtb memory discovery code with RISC-V to avoid duplicating code (the code is also slightly improved, `memory` nodes don't necessarily have to have an unit associated with them like the current code assumed).


avdgrinten (commented):
Looks good, thank you for converting Eir to the new inigraph code! It's great to have all architectures unified now.

avdgrinten (approved):

qookei (approved):

PR #796, opened 2024-12-22
Fixes in preparation for NVMe-over-fabrics

no92:


servers/netserver/src/ip/tcp4.cpp
```diff
@@ -690,14 +698,6 @@ async::result<void> Tcp4Socket::flushOutPackets_() {
 				continue;
 			}
 
-			// Construct and transmit the TCP packet.
```
avdgrinten, line 693: Nit: this comment should probably remain here (and not be moved up).

protocols/fs/src/server.cpp
```diff
@@ -937,6 +937,238 @@ async::detached handlePassthrough(smarter::shared_ptr<void> file,
 	}
 }
 
+async::detached handleMessages(smarter::shared_ptr<void> file,
+		const FileOperations *file_ops,
+		bragi::preamble preamble,
+		helix_ng::RecvInlineResult &recv_req,
```
avdgrinten, line 943: Taking the `recv_req` by reference is problematic since this coro is detached.

avdgrinten: Taking it by value and moving it should be fine.

avdgrinten (approved):

PR #798, opened 2024-12-23
Initial batch of systemd work

Dennisbonke:
Just some small bits and bops here and there. Most notably comes with infrastructure for fstatfs.

### DIFF
--- a/posix/subsystem/src/fs.cpp
+++ b/posix/subsystem/src/fs.cpp
@@ -25,6 +25,10 @@ async::result<frg::expected<Error, FileStats>> FsNode::getStats() {
 	throw std::runtime_error("getStats() is not implemented for this FsNode");
 }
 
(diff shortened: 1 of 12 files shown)


posix/subsystem/src/fs.cpp
```diff
@@ -25,6 +25,10 @@ async::result<frg::expected<Error, FileStats>> FsNode::getStats() {
 	throw std::runtime_error("getStats() is not implemented for this FsNode");
 }
 
+async::result<frg::expected<Error, FsFileStats>> FsNode::getFsstats() {
+	throw std::runtime_error("getFsstats() is not implemented for this FsNode");
```
avdgrinten, line 29: This should log and return an error, similar to `symlink()` etc.

Dennisbonke:
Superseded by #815

PR #800, opened 2024-12-26
thor: Avoid calling uACPI functions while not on a fiber

qookei:
Calling uACPI functions from work queues leads to issues like entering a work queue recursively.

### DIFF
--- a/kernel/thor/system/acpi/thor-internal/acpi/acpi.hpp
+++ b/kernel/thor/system/acpi/thor-internal/acpi/acpi.hpp
@@ -7,6 +7,8 @@
 #include <uacpi/uacpi.h>
 #include <uacpi/utilities.h>
 #include <initgraph.hpp>
+#include <async/queue.hpp>
+#include <async/oneshot-event.hpp>
 
 namespace thor {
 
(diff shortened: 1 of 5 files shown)


kernel/thor/system/acpi/thor-internal/acpi/acpi.hpp
```diff
@@ -140,4 +140,31 @@ struct AcpiObject final : public KernelBusObject {
 	async::oneshot_event completion = {};
 };
 
+
+struct AcpiFiberWork {
+	void (*fn)(void *ev, void *work);
+	void *ev, *work;
```
no92, line 146: Does `ev` need to be a `void *` pointer?

qookei: True, that could be a `async::oneshot_event *`.

avdgrinten (approved):

kernel/thor/system/acpi/thor-internal/acpi/acpi.hpp
```diff
@@ -140,4 +140,32 @@ struct AcpiObject final : public KernelBusObject {
 	async::oneshot_event completion = {};
 };
 
+
+struct AcpiFiberWork {
+	void (*fn)(async::oneshot_event *ev, void *work);
+	async::oneshot_event *ev;
+	void *work;
+};
```
avdgrinten, line 148: Nit: it arguably nicer to let the queue store a `AcpiFiberWork *` while turning the `async::oneshot_event *` here into a `async::oneshot_event`. Then the `AcpiFiberWork` can be put onto the stack without requiring an external oneshot event. In this case, it doesn't really make a difference though.

PR #802, opened 2024-12-27
NVMe-over-fabrics support, part 2

no92:
Various fixes in preparation for NVMe-oF.

### DIFF
--- a/servers/netserver/src/ip/tcp4.cpp
+++ b/servers/netserver/src/ip/tcp4.cpp
@@ -701,7 +701,7 @@ async::result<void> Tcp4Socket::flushOutPackets_() {
 			auto chunk = std::min({
 				bytesAvailable - flushPointer,
 				windowPointer - flushPointer,
-				size_t{1000} // TODO: Perform path MTU discovery.
+				size_t{1280} // TODO: Perform path MTU discovery.
 			});
 
 			std::vector<char> buf;
--- a/servers/netserver/src/main.cpp
+++ b/servers/netserver/src/main.cpp
@@ -9,6 +9,8 @@
 
 #include <async/result.hpp>
 #include <bragi/helpers-std.hpp>
+#include <core/cmdline.hpp>
+#include <frg/cmdline.hpp>
 #include <hel.h>
 #include <hel-syscalls.h>
 #include <helix/ipc.hpp>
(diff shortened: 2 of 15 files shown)


servers/netserver/src/ip/tcp4.cpp
```diff
@@ -646,7 +646,7 @@ async::result<void> Tcp4Socket::flushOutPackets_() {
 				.checksum = 0,
 				.urgentPointer = 0,
 			};
-			header->flags.store(TcpHeader::headerWords(sizeof(TcpHeader) / 4)
+			header->flags.store(TcpHeader::headerWords(buf.size() / 4)
```
avdgrinten, line 649: We should probably drop this commit since we do not send any options right now. If we send options, it needs to be a bit more general (for example, in this case it only works for the `SYN` packet).

servers/netserver/src/main.cpp
```diff
@@ -380,6 +382,59 @@ async::result<protocols::svrctl::Error> bindDevice(int64_t base_id) {
...
+		frg::option{"netserver.gateway", frg::as_string_view(gateway)},
+	};
+	frg::parse_arguments(cmdline.c_str(), args);
+
+	auto convert_ip = [](frg::string_view &str, in_addr *addr) -> bool {
+		char strbuf[INET6_ADDRSTRLEN];
+		strncpy(strbuf, str.data(), str.size());
+		strbuf[str.size()] = '\0';
+		return inet_pton(AF_INET, strbuf, addr) == 1;
+	};
```
avdgrinten, line 408: Why do we copy to a temporary buffer here?

servers/netserver/src/main.cpp
```diff
@@ -380,6 +382,59 @@ async::result<protocols::svrctl::Error> bindDevice(int64_t base_id) {
...
+	auto device = baseDeviceMap.at(base_id);
+
+	Cmdline cmdHelper;
+	auto cmdline = co_await cmdHelper.get();
+	frg::string_view station = "";
+	frg::string_view subnet = "";
+	frg::string_view gateway = "";
+
+	frg::array args = {
+		frg::option{"netserver.station", frg::as_string_view(station)},
```
avdgrinten, line 397: Calling this `netserver.ip` would be less confusing.

servers/netserver/src/main.cpp
```diff
@@ -380,6 +382,59 @@ async::result<protocols::svrctl::Error> bindDevice(int64_t base_id) {
 			co_return err;
 	}
 
+	assert(baseDeviceMap.contains(base_id));
+
+	auto device = baseDeviceMap.at(base_id);
```
avdgrinten, line 388: This code will assign the same IP to all NICs. To avoid this, we could move the command line parsing to stage 1 and use a netlink socket to set up the IP. This might be cleaner, what do you think? Could also be done in a follow up PR.

servers/netserver/src/main.cpp
no92: `inet_pton` takes a (null-terminated) C string, which frg::string_view is not.

servers/netserver/src/main.cpp
no92: It would probably be nice to check whether the device is the one specified in the kernel cmdline I guess.

PR #805, opened 2024-12-30
NVMe-over-fabrics, part 4: the rest

no92:


kernel/thor/system/pci/pci_discover.cpp
```diff
@@ -78,6 +78,13 @@ void PciBus::runRootBus() {
 			assert(ret);
 			device->mbusId = ret.value();
 			device->mbusPublished.raise();
+
+			if(device->acpiNode) {
+				Properties updateProps;
+				updateProps.stringProperty("unix.subsystem", frg::string<KernelAlloc>{*kernelAlloc, "acpi"});
+				updateProps.decStringProperty("acpi.physical_node", device->mbusId, 1);
+				co_await device->acpiNode->updateProperties(updateProps);
+			}
```
avdgrinten, line 87: Doesn't this erase other properties of the ACPI node?

avdgrinten: Also, I dislike the dependencies here: the PCI code updates an ACPI node. Why doesn't the PCI node instead publish what ACPI node it belongs to?

drivers/block/nvme/src/fabric/tcp.cpp
```diff
@@ -0,0 +1,318 @@
...
+
+	auto send_err = co_await file_->sendto(&connect_req, sizeof(connect_req), 0, nullptr, 0);
+	assert(send_err);
+	assert(send_err.value() == sizeof(connect_req));
+
+	spec::tcp::ICResp resp{};
+	auto recv_err = co_await file_->recvfrom(reinterpret_cast<void *>(&resp), sizeof(resp), 0, nullptr, 0);
+	assert(recv_err);
+	assert(recv_err.value() == sizeof(resp));
+	assert(resp.ch.pduType == spec::tcp::PduType::ICResp);
```
avdgrinten, line 39: These asserts should be error returns instead (when in doubt, we can probably just return EIO). Also, it would easily be possible to handle the case where `sendto`/`recvfrom` return partial success.

avdgrinten: Also: is it possible to use the Command infrastructure here? That'd simplify the code since there would be only a single send and receive path.

drivers/block/nvme/src/fabric/tcp.cpp
```diff
@@ -0,0 +1,318 @@
...
+
+	file_ = std::make_unique<protocols::fs::File>(std::move(sock_err.value()));
+
+	sockaddr_in sockaddr{};
+	sockaddr.sin_family = AF_INET;
+	sockaddr.sin_port = htons(port_);
+	sockaddr.sin_addr.s_addr = addr_.s_addr;
+
+	auto connect_err = co_await file_->connect(reinterpret_cast<struct sockaddr *>(&sockaddr), sizeof(sockaddr));
+	assert(connect_err == protocols::fs::Error::none);
```
avdgrinten, line 58: This function should also handle errors. In particular, failure to connect should not crash the NVMe driver.

drivers/block/nvme/src/fabric/tcp.cpp
```diff
@@ -0,0 +1,318 @@
...
+
+	memcpy(connectData.hostIdentifier, uuid_.data(), sizeof(connectData.hostIdentifier));
+	auto nqn = std::format("nqn.2014-08.org.nvmexpress:uuid:{:02x}{:02x}{:02x}{:02x}-{:02x}{:02x}-{:02x}{:02x}-{:02x}{:02x}-{:02x}{:02x}{:02x}{:02x}{:02x}{:02x}",
+		uuid_[0], uuid_[1], uuid_[2], uuid_[3], uuid_[4], uuid_[5], uuid_[6], uuid_[7],
+		uuid_[8], uuid_[9], uuid_[10], uuid_[11], uuid_[12], uuid_[13], uuid_[14], uuid_[15]);
+	strncpy(connectData.hostNqn, nqn.c_str(), sizeof(connectData.hostNqn));
+
+	cmd->setupBuffer(arch::dma_buffer_view{nullptr, &connectData, sizeof(connectData)}, spec::DataTransfer::SGL);
+	auto res = co_await submitCommand(std::move(cmd));
+	assert((res.first & 0x0FFE) == 0);
```
avdgrinten, line 86: What do we check here? This should probably also be an error return.

core/include/core/device-path.hpp
```diff
@@ -0,0 +1,128 @@
...
+#include <format>
+#include <iostream>
+#include <ranges>
+#include <stdint.h>
+#include <string>
+#include <string.h>
+#include <vector>
+
+struct DevicePathParser {
+	DevicePathParser(std::string str) : str_{str} {
```
avdgrinten, line 14: Since this constructor processes external input, it should not use `assert` to validate the data. Maybe the constructor should be replaced by a function that can return an error?

core/include/core/device-path.hpp
```diff
@@ -0,0 +1,128 @@
...
+				memcpy(&n.data[4], &uid_num, sizeof(uid_num));
+
+				nodes_.push_back(n);
+			} else if(p.starts_with("Pci(")) {
+				Node n{
+					.type = DevicePathType::Hardware,
+					.subtype = 1
+				};
+				auto slot = p.substr(4, p.find_first_of(',', 4) - 4);
+				size_t prefix_offset = slot.starts_with("0x") ? 2 : 0;
```
avdgrinten, line 41: Are the number always hex, regardless of 0x prefix?

drivers/block/nvme/src/fabric/tcp.cpp
no92: Using the Command infra is not possible here, as these commands are not NVMe commands, but commands specific to the transport binding.

drivers/block/nvme/src/fabric/tcp.cpp
```diff
@@ -0,0 +1,318 @@
...
+		while(received < ch->pduLength) {
+			recv_err = co_await file_->recvfrom(recvbuf + received, ch->pduLength - received, 0, nullptr, 0);
+			assert(recv_err);
+			received += recv_err.value();
+		}
+
+		ch = reinterpret_cast<spec::tcp::PduCommonHeader *>(recvbuf);
+
+		switch(ch->pduType) {
+			case spec::tcp::PduType::CapsuleResp: {
```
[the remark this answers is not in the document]
no92: The spec makes no mention of such inline data in `CapsuleResp`, all it can deliver is a CQE and header digests (which we don't enable or implement yet).

kernel/thor/system/pci/pci_discover.cpp
no92: `updateProperties` does not erase other properties, it only creates or updates the properties passed to it.

no92: For reference, the updated object looks like this in the relevant `lsmbus` dump:
```
Entity "acpi-object" (ID 3):
        acpi.uid: 0
        unix.subsystem: acpi
        acpi.path: \_SB_.PCI0
        acpi.hid: PNP0A03
        acpi.instance: 0
        acpi.physical_node: 11
```

avdgrinten (approved):
LGTM when the final comments are fixed.

core/include/core/device-path.hpp
```diff
@@ -11,15 +11,21 @@
...
 
 		for(auto p : parts) {
 			if(p.starts_with("PciRoot(")) {
 				auto uid = p.substr(8, p.size() - 9);
 				uint32_t uid_num = 0;
-				auto res = std::from_chars(uid.begin(), uid.end(), uid_num);
-				assert(res.ec != std::errc::invalid_argument && res.ec != std::errc::result_out_of_range);
+				auto uid_val = integerFromString<uint32_t>(uid);
+				if(uid_val)
+					uid_num = uid_val.value();
```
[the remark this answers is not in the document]
no92: If this fails, the default value should be zero as the UID parameter is optional.

PR #819, opened 2025-01-12
thor: Add load balancing

avdgrinten:
Thor already had code to migrate threads from one CPU to another but it did not migrate automatically. Add an algorithm that tries to balance the load periodically.

### DIFF
--- a/kernel/thor/generic/load-balancing.cpp
+++ b/kernel/thor/generic/load-balancing.cpp
@@ -1,0 +1,223 @@
+#include <frg/unique.hpp>
+#include <thor-internal/load-balancing.hpp>
+#include <thor-internal/timer.hpp>
+
+namespace thor {
+
+namespace {
+
+constexpr bool debugLb = false;
+
+// Basic settings.
+constexpr bool enableLb = true;
+constexpr uint64_t lbInterval = 100'000'000;
+
+// Load decay factor (scale is hardcoded to 8 below) and decay interval.
+constexpr uint64_t lbDecay = 184;
+constexpr uint64_t lbDecayInterval = 1'000'000'000;
+
+frg::eternal<LoadBalancer> loadBalancer;
+
+} // namespace
+
+THOR_DEFINE_PERCPU(lbNode);
+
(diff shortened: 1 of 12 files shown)


kernel/thor/generic/load-balancing.cpp
```diff
@@ -0,0 +1,221 @@
...
+			uint64_t newLoad = thisNode->totalLoad;
+			for (size_t i = 0; i < getCpuCount(); ++i) {
+				auto *toCpu = getCpuData(i);
+				if (cpu != toCpu)
+					balanceBetween_(&lbNode.get(toCpu), thisNode, newLoad, idealLoad);
+			}
+		}
+
+		// Balance load again after some time has passed.
+		if (!cpu->cpuIndex)
```
qookei, line 146: Maybe add a comment noting that the BSP instead falls through and blocks on the barrier wait instead of sleeping? (also it might be better if only the BSP would sleep instead? since that would mean less pending timers)

kernel/thor/generic/load-balancing.cpp
```diff
@@ -0,0 +1,221 @@
...
+				infoLogger() << "Moving thread with load " << cb->load_
+						<< " from CPU " << srcNode->cpu->cpuIndex
+						<< " to CPU " << dstNode->cpu->cpuIndex << frg::endlog;
+
+			// Move ownership from srcNode to dstNode.
+			assert(cb->node_ == srcNode);
+			srcNode->tasks.erase(currentIt);
+			cb->node_ = dstNode;
+			cb->_assignedCpu.store(dstNode->cpu, std::memory_order_relaxed);
+			stolenTasks.push_back(cb);
```
qookei, line 205: This doesn't account for the thread's affinity mask.

no92:
This is blocked on managarm/libasync#25.

kernel/thor/generic/load-balancing.cpp
avdgrinten: That is what the code already does, isn't it? I can add a comment to clarify this though.

PR #821, opened 2025-01-14
posix: Initial cgroupv2 implementation

Dennisbonke:


posix/subsystem/src/cgroupfs.hpp
```diff
@@ -0,0 +1,169 @@
...
+
+
+	VfsType getType() override {
+		return VfsType::symlink;
+	}
+};
+
+} // namespace cgroupfs
+
+std::shared_ptr<FsLink> getCgroupfs();
```
avdgrinten, line 169: Is there only a single cgroupfs? What if you mount cgroup2 twice on Linux? Does it show the same cgroups?

Dennisbonke:
Superseded by #846

PR #822, opened 2025-01-16
Implement `PR_SET_PDEATHSIG`

no92:


Dennisbonke (approved):
LGTM

posix/subsystem/src/process.cpp
```diff
@@ -1213,6 +1213,17 @@ async::result<void> Process::terminate(TerminationState state) {
 		(void)result;
 	}
 
+	for(auto &c : _children) {
+		c->_parent = parent;
```
avdgrinten, line 1217: Processes should be re-parented to the init process or to subreapers (one we implement them) -- but not to the grandparent. This can be implemented by walking up the `_parent` pointers until we hit one of these.

avdgrinten: Also, we need to add `c` to the new parent's `_children` list.

testsuites/posix-tests/src/parent-dead-signal.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+namespace {
+
+int child_end = -1;
+
+void handle_signal(int sig) {
+	printf("Received signal %d. Success!\n", sig);
+	int data = 42;
+	write(child_end, &data, sizeof(data));
+	exit(0);
```
avdgrinten, line 18: We can check that the ppid is now `1`.

no92:
Applied the suggested changes.

PR #823, opened 2025-01-17
memfd read/write support

no92:
This get used by systemd-journald.


posix/subsystem/src/memfd.cpp
```diff
@@ -97,3 +97,11 @@ MemoryFile::writeAll(Process *, const void *data, size_t length) {
 	_offset += length;
 	co_return length;
 }
+
+async::result<frg::expected<Error, size_t>>
+MemoryFile::readSome(Process *, void *data, size_t max_length) {
+	auto read_len = std::min(_fileSize - _offset, max_length);
```
avdgrinten, line 103: This needs a check for `_offset <= _fileSize` since `seek()` may position the offset outside of the file.

posix/subsystem/src/memfd.cpp
```diff
@@ -76,3 +80,20 @@ MemoryFile::addSeals(int seals) {
...
+async::result<frg::expected<Error, size_t>>
+MemoryFile::writeAll(Process *, const void *data, size_t length) {
+	if(_seals & F_SEAL_WRITE)
+		co_return Error::insufficientPermissions;
+
+	auto end_size = _offset + length;
+	if(end_size > _fileSize) {
+		auto e = _resizeFile(end_size);
+		if(!e)
+			co_return Error::insufficientPermissions;
```
avdgrinten, line 93: Why not simply `FRG_CO_TRY` here? Converting all errors to `insufficientPermissions` looks wrong.

posix/subsystem/src/memfd.cpp
```diff
@@ -34,21 +33,24 @@ MemoryFile::allocate(int64_t offset, size_t size) {
 	/* if the file size isn't enough */
 	if(_seals & F_SEAL_GROW)
 		co_return protocols::fs::Error::insufficientPermissions;
```
avdgrinten, line 35: This is now handled by `_resizeFile`, isn't it?

posix/subsystem/src/memfd.cpp
no92: `_resizeFile` returns a `protocols::fs::Error`, while this functioon returns an `Error`; `EPERM` is also the only permissible error code for resizing.

no92:
Addressed the comments and added a new commit that zeroes out new memory when growing the buffer.

PR #824, opened 2025-01-17
netserver ICMP sockets

no92:
Depends on #825.

### DIFF
--- a/servers/netserver/src/ip/icmp.cpp
+++ b/servers/netserver/src/ip/icmp.cpp
@@ -1,0 +1,291 @@
+#include "icmp.hpp"
+#include "ip4.hpp"
+
+#include <arch/bit.hpp>
+#include <arpa/inet.h>
+#include <async/basic.hpp>
+#include <async/queue.hpp>
+#include <async/recurring-event.hpp>
+#include <async/result.hpp>
+#include <core/clock.hpp>
+#include <cstring>
+#include <format>
+#include <netinet/ip.h>
+#include <netinet/ip_icmp.h>
+#include <protocols/fs/server.hpp>
+#include <sys/epoll.h>
+#include <sys/socket.h>
+#include <sys/time.h>
+
+struct IcmpPacket {
+	struct Header {
+		uint8_t type;
+		uint8_t code;
+		uint16_t checksum;
+		uint32_t rest_of_header;
+	} header;
+	static_assert(sizeof(Header) == 8);
+
+	arch::dma_buffer_view payload() const {
+		return packet->payload();
+	}
+
+	bool parse(smarter::shared_ptr<const Ip4Packet> packet) {
+		if (packet->payload().size() < sizeof(header)) {
+			return false;
+		}
+
+		auto now = clk::getRealtime();
+		TIMESPEC_TO_TIMEVAL(&recvTimestamp, &now);
+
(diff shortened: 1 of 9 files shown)


servers/netserver/src/ip/icmp.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+struct stl_allocator {
+	void *allocate(size_t size) {
+		return operator new(size);
+	}
+
+	void deallocate(void *p, size_t) {
+		return operator delete(p);
+	}
+};
```
avdgrinten, line 28: This exists in frigg.

servers/netserver/src/ip/icmp.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+public:
+	async::queue<IcmpPacket, stl_allocator> queue_;
+	Icmp *parent_;
+
+	async::recurring_event _statusBell;
+	uint64_t _currentSeq;
+	uint64_t _inSeq;
+
+	size_t id_;
```
avdgrinten, line 256: What is this for?

servers/netserver/src/ip/icmp.cpp
```diff
@@ -0,0 +1,290 @@
...
+			const char *creds, uint32_t flags,
+			void *data, size_t len,
+			void *addr_ptr, size_t addr_size,
+			std::vector<uint32_t> fds, struct ucred) {
+		using arch::endian;
+		(void) creds;
+		(void) flags;
+		(void) fds;
+
+		assert(!flags);
```
avdgrinten, line 151: This should log + return error.

servers/netserver/src/ip/icmp.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+		co_return len;
+	}
+
+	static async::result<frg::expected<protocols::fs::Error, protocols::fs::PollWaitResult>>
+	pollWait(void *obj, uint64_t past_seq, int mask, async::cancellation_token cancellation) {
+		auto self = static_cast<IcmpSocket *>(obj);
+		(void)mask; // TODO: utilize mask.
+
+		assert(past_seq <= self->_currentSeq);
```
avdgrinten, line 178: This should return an error.

servers/netserver/src/ip/icmp.cpp
avdgrinten: If its only purpose is to provide a key in the `sockets` map, we should just replace that map by a data structure that doesn't need a key (intrusive linked list?).

servers/netserver/src/ip/icmp.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+	static async::result<RecvResult> recvmsg(void *obj,
+			const char *creds,
+			uint32_t flags, void *data, size_t len,
+			void *addr_buf, size_t addr_size, size_t max_ctrl_len) {
+		using arch::endian, arch::convert_endian;
+		(void) creds;
+		(void) flags;
+
+		assert(!flags);
```
avdgrinten, line 88: This should log + return error.

avdgrinten (commented):
Local variable names should be camelCase. Otherwise, LGTM if the comments are fixed.

avdgrinten (approved):

PR #828, opened 2025-01-20
posix: add umask support

pitust:
This adds the `sys_umask` mlibc sysdep, and makes it work with the usual system calls. Hopefully, unless I screwed it up.

### DIFF
--- a/posix/subsystem/src/debug-options.hpp
+++ b/posix/subsystem/src/debug-options.hpp
@@ -2,6 +2,7 @@
 
 namespace {
 	constexpr bool logRequests = false;
+	constexpr bool logSpammyRequests = false;
 	constexpr bool logPaths = false;
 	constexpr bool logSignals = false;
 	constexpr bool logCleanup = false;
--- a/posix/subsystem/src/process.hpp
+++ b/posix/subsystem/src/process.hpp
@@ -141,13 +141,16 @@ struct FsContext {
 
 	ViewPath getRoot();
 	ViewPath getWorkingDirectory();
+	int getUmask();
 
 	void changeRoot(ViewPath root);
 	void changeWorkingDirectory(ViewPath root);
+	int changeUmask(int mask);
 
 private:
 	ViewPath _root;
 	ViewPath _workDir;
+	int _umask;
 };
 
 struct FileDescriptor {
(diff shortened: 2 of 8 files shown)


posix/subsystem/src/process.hpp
```diff
@@ -141,13 +141,16 @@ struct FsContext {
...
+	int getUmask();
 
 	void changeRoot(ViewPath root);
 	void changeWorkingDirectory(ViewPath root);
+	int changeUmask(int mask);
 
 private:
 	ViewPath _root;
 	ViewPath _workDir;
+	int _umask;
```
no92, line 153: New code should not have members start with an underscore, something like `umask_` is preferable.

protocols/posix/posix.bragi
```diff
@@ -625,3 +626,13 @@ head(128):
...
 }
+
+message UmaskRequest 101 {
+head(128):
+	int32 newmask;
+}
+
+message UmaskResponse 102 {
+head(128):
+	int32 oldmask;
```
no92, line 637: It might be nice to leave a comment about the fact that a umask call never fails, and no error needs to be reported.

protocols/fs/fs.bragi
```diff
@@ -180,7 +180,7 @@ head(128):
 		// used by PT_IOCTL, PT_SET_OPTION.
 		tag(8) int64 command;
 
-		//used by FLOCK
+		// used by FLOCK
```
no92, line 183: Unrelated (formatting) changes don't have to go into this commit.

posix/subsystem/src/un-socket.cpp
```diff
@@ -549,6 +549,8 @@ struct OpenFile : File {
 						conversation,
 						helix_ng::sendBuffer(ser.data(), ser.size())
 					);
+
+					HEL_CHECK(send_resp.error());
```
no92, line 553: Due to us not implementing `/dev/tty` correctly yet, this check is deliberately not done. It also breaks some userspace programs. Also, this change is unrelated to the rest of the commit.

posix/subsystem/src/requests.cpp
```diff
@@ -965,6 +965,14 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 				continue;
 			}
 
+			auto target = std::get<std::shared_ptr<FsLink>>(result)->getTarget();
+			auto chmodResult = co_await target->chmod(req->mode() & ~self->fsContext()->getUmask() & 0777);
```
no92, line 969: It would be more readable to have braces around one of the `&&` for readability.

posix/subsystem/src/debug-options.hpp
```diff
@@ -2,6 +2,7 @@
 
 namespace {
 	constexpr bool logRequests = false;
+	constexpr bool logSpammyRequests = false;
```
no92, line 5: Unrelated changes should go in their own commit.

PR #829, opened 2025-01-20
kernel: allow the managarm kernel to work correctly on Apple M1 in KVM

pitust:
The Apple M1 timer seems to be weird and breaks if you set a deadline of `0xffff'ffff'ffff'ffff`. Work around this by setting a lower value. Also, add memory barriers after changing page tables because it is a requirement on arm.

### DIFF
--- a/kernel/eir/arch/arm/arch.cpp
+++ b/kernel/eir/arch/arm/arch.cpp
@@ -154,7 +154,7 @@ void initProcessorEarly() {
 	uint64_t aa64mmfr0;
 	asm volatile("mrs %0, id_aa64mmfr0_el1" : "=r"(aa64mmfr0));
 
-	if (aa64mmfr0 & (0xF << 28))
+	if (aa64mmfr0 & (uint64_t(0xF) << 28))
 		eir::panicLogger() << "PANIC! This CPU doesn't support 4K memory translation granules"
 		                   << frg::endlog;
 
--- a/kernel/thor/arch/arm/timer.cpp
+++ b/kernel/thor/arch/arm/timer.cpp
@@ -77,7 +77,7 @@ struct VirtualGenericTimer : IrqSink, AlarmTracker {
 	}
 
 	void disarm() {
-		asm volatile ("msr cntv_cval_el0, %0" :: "r"(0xFFFFFFFFFFFFFFFF));
+		asm volatile ("msr cntv_cval_el0, %0" :: "r"(0xFFFF'FFFF'FFFF'FFFFULL));
 	}
 };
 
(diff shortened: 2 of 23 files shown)


kernel/eir/arch/arm/arch.cpp
```diff
@@ -154,7 +154,7 @@ void initProcessorEarly() {
 	uint64_t aa64mmfr0;
 	asm volatile("mrs %0, id_aa64mmfr0_el1" : "=r"(aa64mmfr0));
 
-	if (aa64mmfr0 & (0xF << 28))
+	if (aa64mmfr0 & (uint64_t(0xF) << 28))
```
no92, line 157: It might be cleaner to use suffixes (`ULL`) instead of casting.

kernel/thor/arch/arm/timer.cpp
```diff
@@ -103,7 +103,10 @@ void armPreemption(uint64_t nanos) {
 }
 
 void disarmPreemption() {
-	asm volatile ("msr cntp_cval_el0, %0" :: "r"(0xFFFFFFFFFFFFFFFF));
+	// although it *should* be correct to set this to 0xFFFF'FFFF'FFFF'FFFFULL,
+	// this does not work on apple platforms for whatever reason. work around this
+	// by setting a smaller value that still works.
```
no92, line 108: Is there a way to detect this behavior in order to limit the application of this quirk to the affected platforms? Also, do you happen to know how other OSes deal with this?

kernel/thor/system/pci/pci_dtb.cpp
```diff
@@ -117,9 +117,6 @@ DtbPciIrqRouter::DtbPciIrqRouter(PciIrqRouter *parent_, PciBus *associatedBus_,
 			if (!childIrq.read(index))
 				panicLogger() << "Failed to read pin index from interrupt-map" << frg::endlog;
 
-			if (parentAddress.numCells())
```
no92, line 120: Why is this check dropped? AFAICT the handling for this hasn't been changed in the PR.

kernel/thor/generic/thor-internal/arch-generic/asid.hpp
```diff
@@ -294,6 +294,10 @@ void invalidateAsid(int asid);
 // (globalBindingId for the global page tables).
 void invalidatePage(int asid, const void *address);
 
+// Wait for all preceding page table updates to finish. This is required
+// after mapping new page table entries for the newly created entries to be
+// observable.
+void pageTableUpdateBarrier();
```
avdgrinten, line 300: Can we do this when submitting a shootdown request? That would make the code much cleaner.

qookei: I'm pretty sure this is also needed in cases a shootdown doesn't happen (mapping a page to a previously unmapped VA), since it's synchronizing the PT walker loads with our PT writes.

IMO this could be a member function of the cursor policy, and the generic code could invoke it after every write to the page tables.

avdgrinten: Yes, that makes sense. In this case, we should integrate it into the mapping functions.

PR #846, opened 2025-02-13
Systemd init part1

Dennisbonke:
Various fixes here and there, working towards systemd init.

### DIFF
--- a/posix/subsystem/src/requests.cpp
+++ b/posix/subsystem/src/requests.cpp
@@ -29,6 +29,7 @@
 #include "eventfd.hpp"
 #include "signalfd.hpp"
 #include "tmp_fs.hpp"
+#include "cgroupfs.hpp"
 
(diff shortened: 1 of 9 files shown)


posix/subsystem/src/requests.cpp
```diff
@@ -2307,6 +2346,9 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 				} else if(result.error() == Error::notDirectory) {
 					co_await sendErrorResponse(managarm::posix::Errors::NOT_A_DIRECTORY);
 					continue;
+				} else if(result.error() == Error::illegalOperationTarget) {
```
no92, line 2349: It would be nice to use our new error conversion helpers instead of adding new cases here.

no92:
Squashing the three cgroups commits into one would be great, too.

PR #847, opened 2025-02-13
posix: Fix a netlink send to closed socket bug

Dennisbonke:

### DIFF
--- a/posix/subsystem/src/netlink/nl-socket.cpp
+++ b/posix/subsystem/src/netlink/nl-socket.cpp
@@ -43,6 +43,23 @@ OpenFile::OpenFile(int protocol, bool nonBlock)
 		_protocol{protocol}, ops_(globalProtocolOpsMap.at(protocol)), _currentSeq{1},
 		_inSeq{0}, _socketPort{0}, _passCreds{false}, nonBlock_{nonBlock} { }
 
+void OpenFile::handleClose() {
+	_isClosed = true;
+	_statusBell.raise();
+	_cancelServe.cancel();
+	for(int i = 0; i < MAX_SUPPORTED_GROUP_ID; i++) {
+		if(joinedGroups_ & (1 << i)) {
+			// Remove the netlink socket from the subscription vector
+			auto it = globalGroupMap.find({_protocol, i + 1});
+			assert(it != globalGroupMap.end());
+			auto group = it->second.get();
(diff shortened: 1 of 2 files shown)


posix/subsystem/src/netlink/nl-socket.cpp
```diff
@@ -323,10 +341,13 @@ async::result<frg::expected<protocols::fs::Error>> OpenFile::setSocketOption(int
 		auto val = *reinterpret_cast<int *>(optbuf.data());
 		std::cout << "posix: Join netlink group "
 					<< _protocol << "." << val << std::endl;
+		if(val >= 32)
```
no92, line 344: I would rather replace the 32 with a constant (e.g. `MAX_SUPPORTED_NETLINK_GROUP`) so that we can later correctly implement support for group IDs >= 32.

Leaving a TODO comment here would help for that, too.

posix/subsystem/src/netlink/nl-socket.hpp
```diff
@@ -116,6 +112,9 @@ struct OpenFile : File, core::netlink::NetlinkFile {
 
 	// BPF filter
 	std::optional<std::vector<char>> filter_ = std::nullopt;
+
+	// Group subscriptions
+	uint32_t joinedGroups_ = 0;
```
no92, line 117: Please leave a TODO to remind us to adapt this for Group ID >= 32 support.

posix/subsystem/src/netlink/nl-socket.cpp
```diff
@@ -43,6 +43,23 @@ OpenFile::OpenFile(int protocol, bool nonBlock)
 		_protocol{protocol}, ops_(globalProtocolOpsMap.at(protocol)), _currentSeq{1},
 		_inSeq{0}, _socketPort{0}, _passCreds{false}, nonBlock_{nonBlock} { }
 
+void OpenFile::handleClose() {
+	_isClosed = true;
+	_statusBell.raise();
+	_cancelServe.cancel();
+	for(int i = 0; i < 32; i++) {
```
no92, line 50: Same here, please replace the 32 with a constant.

no92 (changes requested):
The logic by itself is fine, but I'd like to be more explicit about what needs to change in the near future to properly support Group IDs >= 32, as they are needed for delivering broadcasts to userspace for some protocols, notably rtnetlink.

PR #849, opened 2025-02-13
posix: Add /proc/sys/kernel/random/boot_id

Dennisbonke:


posix/subsystem/src/procfs.cpp
```diff
@@ -444,6 +448,37 @@ async::result<void> ArchNode::store(std::string) {
...
+	size_t n = 0;
+	while(n < 16) {
+		size_t chunk;
+		HEL_CHECK(helGetRandomBytes(uuid + n, 16 - n, &chunk));
+		n += chunk;
+	}
+
+	auto a = *reinterpret_cast<uint32_t*>(&uuid[0]);
+	auto b = *reinterpret_cast<uint16_t*>(&uuid[4]);
+	auto c = *reinterpret_cast<uint16_t*>(&uuid[6]);
```
avdgrinten, line 462: The access through `reinterpret_cast` is UB. This should `memcpy` the data out instead.

avdgrinten (commented):
LGTM except for one issue.

no92:
Fixed the comment raised before and added a missing newline in the attribute.

PR #850, opened 2025-02-13
posix: Unix socket SOCK_DGRAM and related fixes

Dennisbonke:
A bunch of SOCK_DGRAM related fixes, might not be completely right but let's review it anyway.


no92 (commented):
I'm not 100 % sure but I think the handling of sockets opened by `socketpair()` is not correct here. It would be nice to verify the behavior of that on Linux and write tests for our posix testsuite to verify correct behavior, but that is probably out of scope for this PR.

avdgrinten (approved):

PR #851, opened 2025-02-13
posix: Assorted fixes for tmpfs MemoryFile

Dennisbonke:
Systemd choked on this, and I have my doubts if that assert is even needed. Also, 0 links makes no sense, hardcoding to 1 sounds better (but implementing it properly is of course even better).

### DIFF
--- a/posix/subsystem/src/tmp_fs.cpp
+++ b/posix/subsystem/src/tmp_fs.cpp
@@ -109,7 +109,7 @@ public:
 
 private:
 	int64_t _inodeNumber;
-	int _numLinks = 0;
+	int _numLinks = 1;
 	mode_t _mode = 0;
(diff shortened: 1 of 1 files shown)


posix/subsystem/src/tmp_fs.cpp
```diff
@@ -650,7 +650,7 @@ MemoryFile::truncate(size_t size) {
 
 async::result<frg::expected<protocols::fs::Error>>
 MemoryFile::allocate(int64_t offset, size_t size) {
-	assert(!offset);
+	// assert(!offset);
```
avdgrinten, line 653: This should be removed instead of commented out.

no92 (approved):
LGTM

PR #853, opened 2025-02-15
AArch64 and DTB fixes and cleanup

qookei:
Fix PPI IrqPins being shared between CPUs, and remove ARM-specific code from the core DTB code.

### DIFF
--- a/kernel/thor/arch/arm/thor-internal/arch/gic.hpp
+++ b/kernel/thor/arch/arm/thor-internal/arch/gic.hpp
@@ -2,12 +2,13 @@
 
 #include <initgraph.hpp>
 #include <thor-internal/arch-generic/cpu.hpp>
+#include <thor-internal/dtb/irq.hpp>
 #include <thor-internal/irq.hpp>
 #include <arch/mem_space.hpp>
 
 namespace thor {
 
-struct Gic {
+struct Gic : dt::IrqController {
 	virtual void sendIpi(int cpuId, uint8_t id) = 0;
 	virtual void sendIpiToOthers(uint8_t id) = 0;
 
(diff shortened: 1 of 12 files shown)


kernel/thor/arch/arm/thor-internal/arch/gic.hpp
```diff
@@ -37,6 +38,64 @@ struct Gic {
...
+			panicLogger() << "Failed to read GIC interrupt index" << frg::endlog;
+		uint32_t flags;
+		if (!irqSpecifier.readSlice(flags, 2, 1))
+			panicLogger() << "Failed to read GIC interrupt flags" << frg::endlog;
+		frg::optional<uint32_t> ppiHandle = 0;
+		if (!irqSpecifier.readSlice(*ppiHandle, 3, 1))
+			ppiHandle = frg::null_opt;
+
+		// TODO(qookie): Handle extended PPI and SPI.
+		assert(type == 0 || type == 1);
```
avdgrinten, line 59: It makes more sense to use `panicLogger()` here such that this is also checked in release mode.

avdgrinten: Also, how does this work with the timer interrupt (which is a PPI and which goes through this code path AFAICT)?

qookei: For PPIs `setupIrq` and `getPin` return the current CPU's pin instead of a global instance.

avdgrinten: Yes, but why doesn't the assertion trigger? I probably misunderstood what this assertion checks for.

qookei: Ah, type 2 and 3 are extended PPIs and SPIs, which have indices > 1024, so they don't appear in QEMU.

avdgrinten (approved):

PR #856, opened 2025-02-17
thor: implement kernel API changes for uACPI 1.0.1

no92:
Depends on managarm/bootstrap-managarm#445.

### DIFF
--- a/kernel/thor/system/acpi/glue.cpp
+++ b/kernel/thor/system/acpi/glue.cpp
@@ -23,6 +23,162 @@
 
 using namespace thor;
 
+namespace {
+
+#ifdef THOR_ARCH_SUPPORTS_PIO
+template<typename T>
+uacpi_status uacpi_kernel_raw_io_write(
+	uacpi_io_addr address, T in_value
+) {
+	uint16_t p = address;
+
+	switch (sizeof(T)) {
+	case 1: {
+		uint8_t v = in_value;
+		asm volatile ("outb %0, %1" : : "a"(v), "d"(p));
+		break;
+	}
+	case 2: {
+		uint16_t v = in_value;
+		asm volatile ("outw %0, %1" : : "a"(v), "d"(p));
+		break;
+	}
+	case 4: {
+		uint32_t v = in_value;
+		asm volatile ("outl %0, %1" : : "a"(v), "d"(p));
+		break;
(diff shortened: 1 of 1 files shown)


kernel/thor/system/acpi/glue.cpp
```diff
@@ -23,6 +23,160 @@
 
 using namespace thor;
 
+namespace {
+
+#ifdef THOR_ARCH_SUPPORTS_PIO
+uacpi_status uacpi_kernel_raw_io_write(
+	uacpi_io_addr address, uacpi_u8 byte_width, uacpi_u64 in_value
```
d-tatianin, line 30: can't this be a template as well, same as the read?

kernel/thor/system/acpi/glue.cpp
```diff
@@ -259,78 +371,43 @@ void uacpi_kernel_pci_device_close(uacpi_handle) {
...
+}
 
-	return UACPI_STATUS_OK;
+uacpi_status uacpi_kernel_pci_write32(
+    uacpi_handle device, uacpi_size offset, uacpi_u32 value
+) {
+	return uacpi_kernel_raw_pci_write(device, offset, value);
 }
 
+
```
d-tatianin, line 410: this seems irrelevant

kernel/thor/system/acpi/glue.cpp
```diff
@@ -23,6 +23,160 @@
...
+	}
+	default:
+		return UACPI_STATUS_INVALID_ARGUMENT;
+	}
+
+	return UACPI_STATUS_OK;
+}
+
+template<typename T>
+uacpi_status uacpi_kernel_raw_io_read(
```
d-tatianin, line 58: should probably be marked static and renamed?

d-tatianin (changes requested):

kernel/thor/system/acpi/glue.cpp
no92: It already is in an anonymous namespace, so this should not be a problem.

kernel/thor/system/acpi/glue.cpp
no92: Same as with the read.

qookei (approved):

PR #859, opened 2025-02-19
posix/init: switch to systemd udev

no92:
Depends on managarm/bootstrap-managarm#436.

### DIFF
--- a/posix/init/src/stage2.cpp
+++ b/posix/init/src/stage2.cpp
@@ -28,14 +28,23 @@ int main() {
 
 	waitpid(xbps_reconfigure, nullptr, 0);
 
+	if(access("/etc/udev/hwdb.bin", F_OK)) {
+		std::cout << "init: systemd-hwdb update" << std::endl;
+		auto hwdb_update = fork();
+		if(!hwdb_update) {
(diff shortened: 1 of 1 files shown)


posix/init/src/stage2.cpp
```diff
@@ -28,14 +28,23 @@ int main() {
...
 	waitpid(xbps_reconfigure, nullptr, 0);
 
+	if(access("/etc/udev/hwdb.bin", F_OK)) {
+		std::cout << "init: systemd-hwdb update" << std::endl;
+		auto hwdb_update = fork();
+		if(!hwdb_update) {
+			execl("/usr/bin/systemd-hwdb", "systemd-hwdb", "update", nullptr);
+		}else assert(hwdb_update != -1);
+		waitpid(hwdb_update, nullptr, 0);
+	}
```
avdgrinten, line 38: Shouldn't this be done as a post install script that's triggered by `xbps-reconfigure`?

no92: This cannot be done at build-time, but needs to be done at boot-time. systemd has a unit file for this, but we do it manually here.

avdgrinten: Post install scripts run during the first boot (see the `xbps-reconfigure` invocation just above this chunk).

PR #860, opened 2025-02-19
drivers/nvme: request a number of I/O queues as mandated by spec

no92:
The NVMe specification requires a `Set Features` command with Feature ID 0x07 (Number of Queues) to be issued before setting up I/O queues. This has been the case since version 1.0 of the spec. For reference, see section 7.6.1 "Initialization" of the NVMe 1.3d spec (p. 115), point 8.


marv7000 (approved):
LGTM

PR #861, opened 2025-02-20
NVMe MSI support

no92:


protocols/hw/include/protocols/hw/client.hpp
```diff
@@ -36,6 +36,7 @@ struct PciInfo {
 	ExpansionRomInfo expansionRomInfo;
 	std::vector<Capability> caps;
 	unsigned int numMsis;
+	bool msiX;
```
avdgrinten, line 39: `bool msiX{false};`? Technically unnecessary but may be a good idea.

drivers/block/nvme/src/controller.cpp
```diff
@@ -91,10 +86,43 @@ async::detached PciExpressController::handleIrqs() {
...
+		if(!isMsiX)
+			regs_.store(regs::intms, 1 << queueId);
+
+		HEL_CHECK(awaitResult.error());
+		irqSequence_ = awaitResult.sequence();
+
+		int found = static_cast<PciExpressQueue *>(q->get())->handleIrq();
+
+		if(!found)
+			std::cout << std::format("nvme: Queue ID {} has no completions despite raised MSI", queueId) << std::endl;
```
avdgrinten, line 120: Does this happen in practice? Does it add log spam?

no92: I did not see this much.

PR #862, opened 2025-02-22
posix: Rework FIFO code to allow read-write mode

avdgrinten:
The first commit merges `ReaderFile` and `WriterFile` into a single `OpenFile` class. This should preserve the semantics of the code; I tested the changes by using shell pipelines.

The second commit allows `open()` on FIFOs to pass `O_RDWR`. This is a Linux extension. The read-write mode should probably get some additional testing before this is merged.


avdgrinten:
Added a test for the read-write mode behavior. This should be good to go.

no92 (approved):
LGTM

PR #863, opened 2025-02-22
Mount information in procfs

no92:
Implements `/proc/mounts` and `/proc/[pid]/mountinfo`, which should allow for running `mount` on managarm.

### DIFF
--- a/posix/subsystem/src/cgroupfs.hpp
+++ b/posix/subsystem/src/cgroupfs.hpp
@@ -109,7 +109,9 @@ protected:
 
 struct SuperBlock final : FsSuperblock {
 public:
-	SuperBlock() = default;
+	SuperBlock() {
+		deviceMinor_ = getUnnamedDeviceIdAllocator().allocate();
+	}
 
 	FutureMaybe<std::shared_ptr<FsNode>> createRegular(Process *) override;
 	FutureMaybe<std::shared_ptr<FsNode>> createSocket() override;
@@ -117,6 +119,17 @@ public:
 	async::result<frg::expected<Error, std::shared_ptr<FsLink>>>
 			rename(FsLink *source, FsNode *directory, std::string name) override;
 	async::result<frg::expected<Error, FsFileStats>> getFsstats() override;
+
+	std::string getFsType() override {
+		return "cgroup2";
+	}
+
+	dev_t deviceNumber() override {
+		return makedev(0, deviceMinor_);
+	}
+
+private:
+	unsigned int deviceMinor_;
 };
 
 struct DirectoryNode final : FsNode, std::enable_shared_from_this<DirectoryNode> {
--- a/posix/subsystem/src/extern_fs.cpp
+++ b/posix/subsystem/src/extern_fs.cpp
@@ -147,7 +147,7 @@ struct Node : FsNode {
 
 
 public:
-	Node(uint64_t inode, helix::UniqueLane lane, Superblock *sb = nullptr)
+	Node(uint64_t inode, helix::UniqueLane lane, Superblock *sb)
 	: FsNode{sb}, _inode{inode}, _lane{std::move(lane)} { }
 
 protected:
@@ -361,8 +361,8 @@ private:
 	}
 
 public:
-	SymlinkNode(uint64_t inode, helix::UniqueLane lane)
-	: Node{inode, std::move(lane)} { }
+	SymlinkNode(Superblock *sb, uint64_t inode, helix::UniqueLane lane)
+	: Node{inode, std::move(lane), sb} { }
 };
 
 struct Link : FsLink {
@@ -950,7 +950,7 @@ std::shared_ptr<Node> Superblock::internalizePeripheralNode(int64_t type,
 		node = std::make_shared<RegularNode>(this, id, std::move(lane));
 		break;
 	case managarm::fs::FileType::SYMLINK:
-		node = std::make_shared<SymlinkNode>(id, std::move(lane));
+		node = std::make_shared<SymlinkNode>(this, id, std::move(lane));
 		break;
 	default:
 		throw std::runtime_error("extern_fs: Unexpected file type");
--- a/posix/subsystem/src/extern_fs.cpp
+++ b/posix/subsystem/src/extern_fs.cpp
@@ -18,7 +18,7 @@ struct Node;
 struct DirectoryNode;
 
 struct Superblock final : FsSuperblock {
-	Superblock(helix::UniqueLane lane);
+	Superblock(helix::UniqueLane lane, std::shared_ptr<UnixDevice> device);
 
 	FutureMaybe<std::shared_ptr<FsNode>> createRegular(Process *process) override;
 	FutureMaybe<std::shared_ptr<FsNode>> createSocket() override;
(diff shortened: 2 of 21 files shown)


posix/subsystem/src/vfs.hpp
```diff
@@ -56,23 +80,20 @@ struct MountView : std::enable_shared_from_this<MountView> {
 		}
 	};
 
+	std::set<std::shared_ptr<MountView>, Compare> &mounts() {
+		return _mounts;
+	}
```
avdgrinten, line 85: This should be `const`.

posix/subsystem/src/fs.cpp
```diff
@@ -28,6 +28,10 @@ struct AnonymousSuperblock : FsSuperblock {
 		};
 		co_return stats;
 	}
+
+	std::string getFsType() override {
+		return {};
+	}
```
avdgrinten, line 34: Can this ever be called? If not, warn if it is?

posix/subsystem/src/procfs.cpp
```diff
@@ -864,6 +867,52 @@ expected<std::string> SymlinkNode::readSymlink(FsLink *, Process *process) {
...
+		std::string devName = [&]() {
+			if(dev.second) {
+				return dev.getPath(proc->fsContext()->getRoot());
+			} else {
+				return fsType;
+			}
+		}();
+
+		auto mountPath = root ? "/" :
+			ViewPath{mount->getParent(), mount->getAnchor()}.getPath(proc->fsContext()->getRoot());
```
avdgrinten, line 888: Why are there two cases here?

posix/subsystem/src/requests.cpp
```diff
@@ -1441,6 +1443,7 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 			resp.set_mtime_nanos(stats.mtimeNanos);
 			resp.set_ctime_secs(stats.ctimeSecs);
 			resp.set_ctime_nanos(stats.ctimeNanos);
+			resp.set_mount_id(target_mount ? target_mount->mountId() : 0);
```
avdgrinten, line 1446: When is this `nullptr`?

protocols/posix/posix.bragi
```diff
@@ -258,6 +258,8 @@ head(128):
 
 		// returned by GET_RESOURCE_USAGE
 		tag(29) uint64 ru_user_time;
+
+		tag(32) uint32 mount_id;
```
avdgrinten, line 262: We should convert the request/response to a modern one.

posix/subsystem/src/procfs.cpp
```diff
@@ -963,6 +964,93 @@ async::result<void> MountInfoNode::store(std::string) {
...
+}
+
+helix::BorrowedDescriptor FdInfoDirectoryFile::getPassthroughLane() {
+	return _passthrough;
+}
+
+async::result<std::string> FdInfoNode::show(Process *) {
+	auto seekResult = co_await file_->seek(0, VfsSeek::relative);
+	auto pos = seekResult ? seekResult.value() : 0;
+	auto mountId = mountView_ ? mountView_->mountId() : 0;
```
avdgrinten, line 1043: When is `mountView_` nullptr?

posix/subsystem/src/procfs.cpp
```diff
@@ -963,6 +964,93 @@ async::result<void> MountInfoNode::store(std::string) {
...
+	co_return File::constructHandle(std::move(file));
+}
+
+async::result<frg::expected<Error, std::shared_ptr<FsLink>>> FdInfoDirectoryNode::getLink(std::string name) {
+	for(const auto &[fdnum, fd] : _process->fileContext()->fileTable()) {
+		if(name != std::to_string(fdnum))
+			continue;
+		auto pointee = std::make_shared<FdInfoNode>(fd.file->associatedMount(), fd.file);
+		co_return std::make_shared<Link>(shared_from_this(), name, pointee);
+	}
```
avdgrinten, line 1009: Instead of looping here, it would be more efficient to convert `name` to an integer that we can look up in `fileTable()`.

posix/subsystem/src/procfs.cpp
no92: Otherwise, in the `root == true` case, we would end up resolving this to `/realfs`.

posix/subsystem/src/vfs.hpp
```diff
@@ -18,26 +20,48 @@ inline constexpr ResolveFlags resolvePrefix = (1 << 4);
...
+		deviceLink_{std::move(deviceLink)}
+	{ }
+
+	uint64_t mountId() const {
+		return mountId_;
+	}
 
+	dev_t deviceNumber() const {
+		return makedev(0, mountId_);
+	}
```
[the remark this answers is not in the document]
no92: Linux uses major 0 for this, and allocates the minor. Because our mount IDs have similar constraints, we just reuse them to avoid introducing another identifier.

posix/subsystem/src/procfs.cpp
no92: This happens for some PID 1 files for instance.

posix/subsystem/src/requests.cpp
no92: This case can happen it the file belongs to a mount outside of the current chroot, such as something opened by PID 1 before pivoting to `/realfs`.

protocols/posix/posix.bragi
no92: While I agree that is the way forward, splitting out a 15-field message feels like quite the scope creep for this PR.

posix/subsystem/src/procfs.cpp
avdgrinten: Please add `// TODO: We should probably ensure that the MountView pointer is never null.`

posix/subsystem/src/procfs.cpp
avdgrinten: Alright, makes sense. Maybe this logic should get a comment.

posix/subsystem/src/fs.cpp
```diff
@@ -37,6 +54,11 @@ FsSuperblock *getAnonymousSuperblock() {
 	return &sb;
 }
 
+id_allocator<unsigned int> &getAnonymousMountIdAllocator() {
+	static id_allocator<unsigned int> anonymousMountIdAllocator{1};
+	return anonymousMountIdAllocator;
+}
```
avdgrinten, line 60: This does not allocate mount IDs. The Linux kernel documentation calls these numbers "unnamed devices". What about `getUnnamedDeviceIdAllocator()`? We should also add a comment `// This is used to allocate device IDs for non-device-based file systems such as tmpfs.`

posix/subsystem/src/cgroupfs.hpp
```diff
@@ -109,14 +109,27 @@ struct RegularNode : FsNode, std::enable_shared_from_this<RegularNode> {
...
+	std::string getFsType() override {
+		return "cgroup2";
+	}
+
+	dev_t deviceNumber() override {
+		return makedev(0, mountMinor_);
+	}
+
+private:
+	unsigned int mountMinor_;
```
avdgrinten, line 132: This has nothing to do with mounts, hence:
```suggestion
	unsigned int deviceMinor_;
```
Likewise for the other file systems.

posix/subsystem/src/extern_fs.cpp
```diff
@@ -39,6 +48,8 @@ struct Superblock final : FsSuperblock {
 	std::map<uint64_t, std::weak_ptr<DirectoryNode>> _activeStructural;
 	std::map<uint64_t, std::weak_ptr<Node>> _activePeripheralNodes;
 	std::map<std::tuple<uint64_t, std::string, uint64_t>, std::weak_ptr<FsLink>> _activePeripheralLinks;
+
+	UnixDevice *device_;
```
avdgrinten, line 52: We want a `shared_ptr<UnixDevice>` here.

posix/subsystem/src/procfs.cpp
```diff
@@ -242,6 +262,7 @@ std::shared_ptr<Link> DirectoryNode::createRootDirectory() {
 	the_node->_entries.insert(std::move(self_thread_link));
 
 	the_node->directMkregular("uptime", std::make_shared<UptimeNode>());
+	the_node->directMkregular("mounts", std::make_shared<MountsNode>());
```
avdgrinten, line 265: On Linux, `/proc/mounts` is a symlink to `self/mounts`.

PR #868, opened 2025-02-24
posix-tests: Enable build on Linux and fix/workaround various bugs

avdgrinten:
This enables us to build the POSIX testsuite on Linux to allow for easier comparisons to Managarm.

Merge together with https://github.com/managarm/bootstrap-managarm/pull/458.


Dennisbonke (commented):
LGTM, but are we sure all sysdeps that used NO_SUCH_FD are changed over now? If so, can't we just remove NO_SUCH_FD from the protocol files?

no92 (approved):
LGTM, but this probably requires mlibc changes as Dennis pointed out.

PR #870, opened 2025-03-01
Random fixes

no92:


kernel/thor/system/pci/thor-internal/pci/pci.hpp
```diff
@@ -194,6 +194,7 @@ struct PciBus : private KernelBusObject {
 
 		auto router = irqRouter->makeDownstreamRouter(newBus);
 		newBus->irqRouter = router;
+		newBus->mbusPublished.raise();
```
avdgrinten, line 197: This looks wrong to me. Why do we need this? Why is it even correct?

kernel/thor/system/pci/pci_discover.cpp
```diff
@@ -1291,12 +1291,24 @@ void checkPciBus(PciBus *bus, EnumFunc &&enumerateDownstream) {
 		checkPciDevice(bus, slot, enumerateDownstream);
 }
 
+namespace {
+
+void runBridge(PciBridge *bridge) {
```
avdgrinten, line 1296: Maybe rename to `runChildBridges()`.

drivers/block/nvme/src/controller.cpp
```diff
@@ -114,10 +114,7 @@ async::detached PciExpressController::handleMsis(helix::UniqueDescriptor irq, si
 		HEL_CHECK(awaitResult.error());
 		irqSequence_ = awaitResult.sequence();
 
-		int found = static_cast<PciExpressQueue *>(q->get())->handleIrq();
-
-		if(!found)
-			std::cout << std::format("nvme: Queue ID {} has no completions despite raised MSI", queueId) << std::endl;
+		static_cast<PciExpressQueue *>(q->get())->handleIrq();
```
avdgrinten, line 117: Maybe guard this behind a `constexpr bool`? Also, please open an issue. It should probably still be investigated (too many IRQs may degrade performance).

PR #874, opened 2025-03-06
thor: Fix missing 0 in pteAddress

marv7000:
This was probably just a typo.


avdgrinten:
It's probably better to write this as `0x000F'FFFF'FFFF'F000` to make it more obvious that it's correct. Also, this mistake was copied over to `ept.cpp` and `npt.cpp`. Can you also fix these files?

Thank you for finding this bug!

marv7000:
Sure, will do in 1-2h

avdgrinten (approved):

PR #875, opened 2025-03-07
thor: add Intel IOMMU support

no92:


kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,826 @@
...
+struct [[gnu::packed]] DeviceScope {
+	uint8_t type;
+	uint8_t length;
+	uint16_t reserved;
+	uint8_t enumeration_id;
+	uint8_t start_bus_number;
+};
+static_assert(sizeof(DeviceScope) == 6);
+
+namespace RootTable {
```
avdgrinten, line 63: We use `snake_case` for namespaces.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,826 @@
...
+	constexpr arch::field<uint64_t, uint8_t> faultReason{32, 8};
+	[[maybe_unused]]
+	constexpr arch::field<uint64_t, uint8_t> addressType{60, 2};
+	constexpr arch::field<uint64_t, bool> read{62, 1};
+	constexpr arch::field<uint64_t, bool> fault{63, 1};
+}
+
+namespace {
+
+size_t nextIommuId = 0;
```
avdgrinten, line 253: This is fine, but maybe it is easier to refer to IOMMUs using their base addresses?

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,826 @@
...
+		case 2: return "Context Entry not present";
+		case 3: return "Invalid Programming of Context Entry";
+		default: return "Reserved or Unhandled";
+	}
+}
+
+} // namespace
+
+struct IntelIommu final : Iommu, IrqSink {
+	IntelIommu(DmarDrhd *drhd)
```
avdgrinten, line 267: From the DRHD we only need base address and segment, don't we? It is probably cleaner to just take base address + segment.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,826 @@
...
+			VirtualAddr(registers), drhd->register_base,
+			page_access::write, CachingMode::mmio);
+
+		regs_ = arch::mem_space{registers};
+
+		cap_ = regs_.load(capability);
+		ecap_ = regs_.load(extendedCapability);
+		auto sagaw = cap_ & Capability::sagaw;
+		assert(sagaw);
+		sagaw_ = ((sizeof(unsigned int) * 8) - __builtin_clz(sagaw)) + 1;
```
avdgrinten, line 284: I don't think this is valid? For example, AFAICT the IOMMU could support 3 level tables and 5 level tables but not 4 level tables.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,826 @@
...
+				v & Version::major, v & Version::minor) << frg::endlog;
+		infoLogger() << frg::fmt("thor: DRHD for segment {}", segment_) << frg::endlog;
+		infoLogger() << frg::fmt("thor: cap 0x{:016x} ecap 0x{:016x}",
+			uint64_t{cap_}, uint64_t{ecap_}) << frg::endlog;
+
+		auto name = frg::string<KernelAlloc>{*kernelAlloc, "iommu"} +
+			frg::to_allocated_string(*kernelAlloc, id()) +
+			frg::string<KernelAlloc>{*kernelAlloc, "-msi"};
+
+		auto interrupt = allocateApicMsi(name);
```
avdgrinten, line 330: MSI allocation can fail, we need to check the result here.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,826 @@
...
+		IrqPin::attachSink(interrupt, this);
+
+		writeBufferFlush();
```
avdgrinten, line 333: Why is this needed? It would be good to add a comment.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,826 @@
...
+
+		regs_.store(faultEventControl, FaultEventControl::interruptMask(true));
+
+		regs_.store(faultEventAddress, interrupt->getMessageAddress() & ~0x3);
```
avdgrinten, line 337: `getMessageAddress()` must be 4 byte aligned anyway. We can assert that instead of doing `~0x3`.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,826 @@
...
+		regs_.store(faultEventUpperAddress, interrupt->getMessageAddress() >> 32);
+		regs_.store(faultEventData, interrupt->getMessageData());
+
+		regs_.store(faultStatus, FaultStatus::stickyBits);
+		regs_.store(faultEventControl, FaultEventControl::interruptMask(false));
+
+		writeBufferFlush();
```
avdgrinten, line 344: Why is this done (comment)?

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,826 @@
...
+		setRootEntryTable(rootTablePhys_);
+
+		if(cap_ & Capability::plmr || cap_ & Capability::phmr) {
+			regs_.store(protectedMemoryEnable, regs_.load(protectedMemoryEnable) / ProtectedMemoryEnable::epm(false));
+
+			while(regs_.load(protectedMemoryEnable) & ProtectedMemoryEnable::prs);
+		}
```
avdgrinten, line 351: Why is this done (comment)?

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,826 @@
...
+			for(size_t i = status & FaultStatus::faultRecordIndex; i < faultRecords_.size(); i++) {
+				auto flags = faultRecords_[i].flags.load();
+
+				if(!(flags & FaultRecording::fault))
+					continue;
+
+				auto reason = flags & FaultRecording::faultReason;
+				auto sourceId = SourceID{flags & FaultRecording::sourceIdentifier};
+
+				urgentLogger() << frg::fmt("thor: IOMMU fault {}, {} request from {:02x}:{:02x}:{:x} to 0x{:x}: {} (0x{:x})",
```
avdgrinten, line 372: This can use `warningLogger` (same for the other use of `urgentLogger` below).

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,826 @@
...
+					i, (flags & FaultRecording::read) ? "Read" : "Write",
+					sourceId.bus(), sourceId.device(), sourceId.function(),
+					faultRecords_[i].faultInfo.load(),
+					decodeFaultReason(reason), reason) << frg::endlog;
```
avdgrinten, line 376: Don't we need to clear the fault after logging it?

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,826 @@
...
+			contextTable = {reinterpret_cast<ContextTable::Entry *>(context.get()), 256};
+
+			rootEntry->entry.store(RootTable::present(true) | RootTable::contextEntry(contextPhys >> 12));
+			flush(rootEntry);
+		}
+
+		if(!contextTable.size()) {
+			context = {(rootEntry->entry.load() & RootTable::contextEntry) << 12};
+			contextTable = {reinterpret_cast<ContextTable::Entry *>(context.get()), 256};
+		}
```
avdgrinten, line 415: This can be an `}else{` instead of a second `if`.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,826 @@
...
+
+		contextEntry->low.store(
+			ContextTable::present(true) |
+			ContextTable::translationType(ContextTable::TranslationType::Passthrough)
+		);
+
+		contextEntry->high.store(
+			ContextTable::addressWidth(sagaw_ - 2) |
+			ContextTable::domainId(1)
+		);
```
avdgrinten, line 433: The writes should be swapped such that `present` is only set after the entry is fully written.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,826 @@
...
+
+		auto contextEntry = &contextTable[sourceId.devfn()];
+
+		contextEntry->low.store(
+			ContextTable::present(true) |
+			ContextTable::translationType(ContextTable::TranslationType::Passthrough)
+		);
+
+		contextEntry->high.store(
+			ContextTable::addressWidth(sagaw_ - 2) |
```
avdgrinten, line 431: For now, I would prefer to simply always use a fixed page table depth instead of making this dynamic. We can fail during initialization if our desired depth is not available.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,826 @@
...
+			ContextTable::domainId(1)
+		);
+
+		flush(contextEntry);
+		asm volatile("mfence" : : : "memory");
```
avdgrinten, line 436: The `mfence` is probably unnecessary.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,826 @@
...
+
+		if(initialized_) {
+			invalidateDeviceContext(0, sourceId);
+			invalidateDomainIotlb(1);
+
+			writeBufferFlush();
```
avdgrinten, line 442: Is this necessary? Why?

As far as I understand the spec, invalidation is stronger than write buffer flushing:
> For updates to remapping hardware structures that require context-cache, PASID-cache, IOTLB or
IEC invalidation operations to flush stale entries from the hardware caches, no additional action is
required to make the modifications visible to hardware. This is because, hardware performs an
implicit write-buffer-flushing as a pre-condition to context-cache, PASID-cache, IOTLB and IEC
invalidation operations.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,826 @@
...
+	}
+
+	void flush(void const *ptr, size_t len) {
+		if(ecap_ & ExtendedCapability::coherent)
+			return;
+
+		for(size_t i = 0; i < len; i += 64) {
+			asm volatile ("clflush (%0)" : : "r" (reinterpret_cast<uintptr_t>(ptr) + i) : "memory");
+		}
+	}
```
avdgrinten, line 461: It would be nice to move these functions to a header such that they can be reused by other parts of `thor`.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,826 @@
...
+
+	void runGlobalCommand(arch::bit_value<uint32_t> c) {
+		auto status = regs_.load(globalStatus) & arch::bit_mask(uint32_t(~GlobalStatus::clearedOnCompletion));
+		regs_.store(globalCommand, status | c);
+	}
```
avdgrinten, line 466: Why doesn't `runGlobalCommand()` check for completion?

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,826 @@
...
+		if((cap_ & Capability::rwbf) == 0)
+			return;
+
+		runGlobalCommand(GlobalStatus::writeBufferFlushStatus(true));
+
+		while(regs_.load(globalStatus) & GlobalStatus::writeBufferFlushStatus);
+	}
+
+	void invalidateGlobalContext() {
+		using namespace ContextCommand;
```
avdgrinten, line 478: I don't like `using namespace` here.

avdgrinten (commented):
Review of the IOMMU code itself. I still have to review the DMAR parsing and the PCI code.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,836 @@
...
+
+	if(drhd.flags & dmarDrhdFlagsPciIncludeAll) {
+		// we need to allow matching multiple root buses, as the spec explicitly
+		// allows that to happen
+		for(auto b : *pci::allRootBuses) {
+			if(b->segId != drhd.segment)
+				continue;
+
+			if(!b)
+				return nullptr;
```
avdgrinten, line 565: This seems weird. We dereference `b` first, then check it against nullptr? That check also seems useless since we shouldn't add nullptrs to `allRootBuses`.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,836 @@
...
+					memcpy(&func, device_scope.subspan(sizeof(DeviceScope) + pathOffset + 1).data(), sizeof(uint8_t));
+
+					pci::PciDevice *pciDev = nullptr;
+					for(auto d : bus->childDevices) {
+						if(d->slot == slot && d->function == func) {
+							pciDev = d;
+							break;
+						}
+					}
+					assert(pciDev);
```
avdgrinten, line 625: Likewise.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,836 @@
...
+				continue;
+
+			bus = b;
+			break;
+		}
+
+		for(size_t pathOffset = 0; pathOffset < (dev.length - 6); pathOffset += 2) {
+			switch(dev.type) {
+				case 1: {
+					assert(bus);
```
avdgrinten, line 613: This uses `assert()` to check the sanity of the DMAR table

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,836 @@
...
+							break;
+						}
+					}
+					assert(pciDev);
+
+					pciDev->associatedIommu = iommu;
+					break;
+				}
+				case 2: {
+					assert(bus);
```
avdgrinten, line 631: Likewise.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,836 @@
...
+							continue;
+						if(b->function != func)
+							continue;
+
+						bus = b->associatedBus;
+						b->associatedIommu = iommu;
+						break;
+					}
+
+					assert(bus);
```
avdgrinten, line 647: This check doesn't verify that the bus was indeed found: bus could simply still point to the old bus here.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,836 @@
...
+		if(dev.length >= device_scope.size())
+			break;
+
+		device_scope = device_scope.subspan(dev.length);
+	}
+
+	return iommu;
+}
+
+void handleRmrr(frg::span<uint8_t> remappingStructureTypes) {
```
avdgrinten, line 665: `assert`s in `handleRmrr` should be adjusted in the same way as in `handleDrhd`.

kernel/thor/arch/x86/vt_d.cpp
avdgrinten: Ideally, we just do not use the IOMMU if the table is broken.

kernel/thor/generic/thor-internal/iommu.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+struct Iommu {
+	Iommu(size_t id)
+	: id_{id} {}
+
+	size_t id() const {
+		return id_;
+	}
+
+	virtual void enableDevice(pci::PciEntity *dev) = 0;
```
avdgrinten, line 22: Do we want to take the `PciEntity` or `(bus, slot, function)`?

avdgrinten, line 22: Do we want to take the `PciEntity` or `(bus, slot, function)`? The latter is probably easier.

kernel/thor/system/pci/pci_discover.cpp
```diff
@@ -699,6 +699,37 @@ coroutine<frg::expected<Error>> PciEntity::handleRequest(LaneHandle lane) {
...
+			infoLogger() << "thor: Closing lane due to illegal HW request." << frg::endlog;
+			co_return Error::protocolViolation;
+		}
+
+		managarm::hw::SvrResponse<KernelAlloc> resp{*kernelAlloc};
+		resp.set_error(managarm::hw::Errors::SUCCESS);
+
+		PciDevice *dev = static_cast<PciDevice *>(this);
+
+		if(dev && dev->associatedIommu) {
```
avdgrinten, line 715: `dev` cannot be null here, can it?

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,836 @@
...
+							pciDev = d;
+							break;
+						}
+					}
+					assert(pciDev);
+
+					infoLogger() << frg::fmt("thor: PCI device {:04x}:{:02x}:{:02x}.{} has RMRR",
+						rmrr.segment, dev.start_bus_number, slot, func) << frg::endlog;
+
+					if(pciDev && pciDev->associatedIommu) {
```
avdgrinten, line 713: According to the assert above, `pciDev` cannot be null here.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,836 @@
...
+						pciDev->associatedIommu->enableDevice(pciDev);
+					} else {
+						auto bridge = pciDev->parentBus->associatedBridge;
+
+						while(bridge && bridge->parentBus && !bridge->associatedIommu)
+							bridge = bridge->parentBus->associatedBridge;
+
+						if(bridge && bridge->associatedIommu)
+							bridge->associatedIommu->enableDevice(bridge);
```
avdgrinten, line 722: Isn't it an error if `bridge` becomes null here (= if there is no associated IOMMU)? We can probably ignore the RMRR if that's the case, but we should emit a warning.

kernel/thor/arch/x86/vt_d.cpp
avdgrinten: We should potentially remove the assertion and add a warning (and ignore the RMRR).

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -0,0 +1,836 @@
...
+					infoLogger() << frg::fmt("thor: PCI bridge at {:04x}:{:02x}:{:02x}.{} to bus {} has RMRR",
+						rmrr.segment, dev.start_bus_number, slot, func, bus->busId) << frg::endlog;
+
+					auto bridge = bus->associatedBridge;
+
+					while(bridge && bridge->parentBus && !bridge->associatedIommu)
+						bridge = bridge->parentBus->associatedBridge;
+
+					if(bridge && bridge->associatedIommu)
+						bridge->associatedIommu->enableDevice(bridge);
```
avdgrinten, line 754: See comments about `bridge` == null above.

kernel/thor/system/pci/pci_discover.cpp
```diff
@@ -699,6 +699,37 @@ coroutine<frg::expected<Error>> PciEntity::handleRequest(LaneHandle lane) {
...
+			while(bridge && bridge->parentBus && !bridge->associatedIommu) {
+				bridge = bridge->parentBus->associatedBridge;
+			}
+
+			if(bridge && bridge->associatedIommu) {
+				bridge->associatedIommu->enableDevice(bridge);
+				bridge->associatedIommu->enableDevice(dev);
+			} else {
+				resp.set_error(managarm::hw::Errors::DEVICE_ERROR);
+			}
```
avdgrinten, line 729: Does this mean that this request returns an error if no IOMMU is present?

no92: Yes, or if the device isn't covered by an IOMMU. The client side does not care about this for now.

kernel/thor/generic/thor-internal/iommu.hpp
no92: I made the generic interface take a PciEntity to allow for future support of different IOMMU designs that might require more that those two parameters.

PR #876, opened 2025-03-07
thor: Make global log ring available early

avdgrinten:
This makes the global log buffer (which ultimately feeds into `dmesg`) available early during boot (= immediately after entering `thorMain()`).


no92 (approved):
LGTM

qookei (approved):

PR #877, opened 2025-03-09
posix/init: Make systemd boot be configurable from the kernel commandline

Dennisbonke:
This eases debugging systemd on Managarm.


no92 (approved):
LGTM

PR #879, opened 2025-03-09
Systemd init part3

Dennisbonke:
This PR adds the final touches to an initial boot with systemd.


no92 (approved):
LGTM

PR #882, opened 2025-03-12
posix: implement SO_TIMESTAMP for unix sockets

no92:


avdgrinten (approved):
LGTM but we may want to hold this until we know that it works with systemd.

no92:
I tested this with systemd and it works.

PR #888, opened 2025-03-14
wait-for-devices: Add utility to wait for udev devices

avdgrinten:
In contrast to Linux, not all devices are already up when userspace boots. In particular, from the perspective of userspace devices such as the graphics card or USB HID devices are hotplugged at runtime. Add a `wait-for-devices` utility that the init system can use to wait until certain devices are available.

### DIFF
--- a/utils/wait-for-devices/src/main.cpp
+++ b/utils/wait-for-devices/src/main.cpp
@@ -1,0 +1,157 @@
+#include <functional>
+#include <iostream>
+#include <print>
+#include <vector>
+
+#include <CLI/CLI.hpp>
+#include <fcntl.h>
+#include <libudev.h>
+#include <poll.h>
+#include <string.h>
+
+struct Predicate {
+	const char *name{nullptr};
+	std::function<bool(udev_device *)> detect;
+};
+
+std::vector<Predicate> pending;
+
+bool debug = false;
+
+int main(int argc, char **argv) {
+	bool wantGraphics{false};
+	bool wantMouse{false};
+	bool wantKeyboard{false};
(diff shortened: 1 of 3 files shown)


utils/wait-for-devices/src/main.cpp
```diff
@@ -0,0 +1,156 @@
...
+		udev_device_unref(udevDevice);
+	}
+
+	if (pending.empty())
+		return 0;
+
+	// Finally, wait until missing devices show up.
+	std::println(std::cout, "Waiting for missing devices to show up");
+	for (auto &predicate : pending)
+		std::println(std::cout, "    Missing: {}", predicate.name);
```
no92, line 121: Specifying `std::cout` here is unnecessary, as the definition 1 (see cppreference.com) implies printing to stdout.

utils/wait-for-devices/src/main.cpp
```diff
@@ -0,0 +1,156 @@
...
+			std::println(std::cerr, "udev_monitor_receive_device() failed");
+			exit(1);
+		}
+		auto syspath = udev_device_get_syspath(udevDevice);
+		if (debug)
+			std::println(std::cerr, "Monitor probes {}", syspath);
+
+		std::erase_if(pending, [&] (auto &predicate) {
+			if (predicate.detect(udevDevice)) {
+				std::println(std::cout, "Monitor found {} at {}", predicate.name, syspath);
```
no92, line 146: Same as above here.

no92 (commented):
LGTM, two minor code nits.

PR #889, opened 2025-03-16
posix: Implement EPOLLET and EPOLLONESHOT

no92:


posix/subsystem/src/epoll.cpp
```diff
@@ -12,6 +13,8 @@
 
 namespace {
 
+constexpr int epollFlags = EPOLLET | EPOLLONESHOT | EPOLLWAKEUP | EPOLLEXCLUSIVE;
```
avdgrinten, line 16: We only use `~epollFlags`. Doesn't it make sense to use a `epollEvents` instead that explicitly lists all possible events? `epollFlags` might still be useful to do error checking on the flags that we get from `epoll_ctl()`.

posix/subsystem/src/epoll.cpp
```diff
@@ -239,11 +242,13 @@ struct OpenFile : File {
...
 
+				auto itemEvents = item->eventMask & ~epollFlags;
+
 				// Discard non-alive items without returning them.
-				if(!(item->state & stateActive)) {
+				if(!(item->state & stateAlive) || !(item->state & stateActive)) {
 					if(logEpoll)
 						std::cout << "posix.epoll \e[1;34m" << structName() << "\e[0m: Discarding"
-								" inactive item \e[1;34m" << item->file->structName() << "\e[0m"
+								" dead item \e[1;34m" << item->file->structName() << "\e[0m"
```
avdgrinten, line 251: The print doesn't match the condition anymore.

posix/subsystem/src/epoll.cpp
```diff
@@ -273,8 +278,25 @@ struct OpenFile : File {
 							<< " is active" << std::endl;
 
 				// Abort early (i.e before requeuing) if the item is not pending.
-				auto status = std::get<1>(result) & (item->eventMask | EPOLLERR | EPOLLHUP);
-				if(!status) {
+				auto status = std::get<1>(result) & (itemEvents | EPOLLERR | EPOLLHUP);
+				if(status) {
+					// We have to increment the sequence again as concurrent waiters
+					// might have seen an empty _pendingQueue.
```
avdgrinten, line 284: This comment should be attached to the `_currentSeq++;` below in the same function.

posix/subsystem/src/epoll.cpp
```diff
@@ -273,8 +278,25 @@ struct OpenFile : File {
 							<< " is active" << std::endl;
 
 				// Abort early (i.e before requeuing) if the item is not pending.
```
avdgrinten, line 280: This comment is outdated.
```suggestion
				// Return pending items to the caller.
```

PR #890, opened 2025-03-17
inotify: Various fixes for inotify support

Dennisbonke:
More systemd bugfixes! This time adding support for `IN_ISDIR`, supporting blocking reads and fixing a bug where we potentially try to use the front of an empty queue.

### DIFF
--- a/posix/subsystem/src/inotify.cpp
+++ b/posix/subsystem/src/inotify.cpp
@@ -54,8 +54,8 @@ public:
 				smarter::shared_ptr<File>{file}, &File::fileOperations));
 	}
 
-	OpenFile()
-	: File{StructName::get("inotify"), nullptr, SpecialLink::makeSpecialLink(VfsType::regular, 0777)} { }
+	OpenFile(bool nonBlock = false)
+	: File{StructName::get("inotify"), nullptr, SpecialLink::makeSpecialLink(VfsType::regular, 0777)}, nonBlock_{nonBlock} { }
 
 	~OpenFile() override {
 		// TODO: Properly keep track of watches.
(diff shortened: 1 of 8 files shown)


posix/subsystem/src/inotify.cpp
```diff
@@ -29,14 +32,35 @@ struct OpenFile : File {
...
+			if(events & FsObserver::modifyEvent)
+				inotifyEvents |= IN_MODIFY;
+			if(events & FsObserver::accessEvent)
+				inotifyEvents |= IN_ACCESS;
+			if(events & FsObserver::closeWriteEvent)
+				inotifyEvents |= IN_CLOSE_WRITE;
+			if(events & FsObserver::closeNoWriteEvent)
+				inotifyEvents |= IN_CLOSE_NOWRITE;
+
+			auto preexists = std::ranges::find_if(file->_queue, [&name, inotifyEvents](auto a) {
```
avdgrinten, line 52: Just capture everything by reference here (IMO individual capture doesn't really make sense for short functions):
```suggestion
			auto preexists = std::ranges::find_if(file->_queue, [&](auto a) {
```

posix/subsystem/src/requests.cpp
```diff
@@ -3236,11 +3237,7 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 
 			assert(!(req->flags() & ~(managarm::posix::OpenFlags::OF_CLOEXEC | managarm::posix::OpenFlags::OF_NONBLOCK)));
```
avdgrinten, line 3238: We should also get rid of this assertion (and log + return an error).

avdgrinten:
Should we also add support to remove watches?

posix/subsystem/src/inotify.cpp
```diff
@@ -172,12 +175,30 @@ struct OpenFile : File {
...
+		if(!watches_.contains(wd))
+			return false;
+
+		auto watch = watches_.at(wd);
+		auto node = watch->node.lock();
+
+		if(node) {
+			node->removeObserver(watch.get());
+			node->notifyObservers(FsObserver::ignoredEvent, {}, 0);
+		}
```
avdgrinten, line 194: Is this really correct? Do _other_ inotify instances get `IN_IGNORED`?

PR #891, opened 2025-03-17
posix: Various getStats fixes and log cleanup

Dennisbonke:


.gitignore
```diff
@@ -1,3 +1,5 @@
 subprojects
 !subprojects/*.wrap
 *.xbstrap
+.clangd
```
no92, line 4: This belongs in your global gitignore or `.git/info/exclude`.

PR #897, opened 2025-03-18
thor: Rework IrqStrategy to make IrqPin logic more flexible

avdgrinten:
This makes it easier to support the PLIC (and in a future PR, also the IMSIC). In particular, level-triggered PLIC interrupts need to be completed when they exit service (and not when the synchronous IRQ handler completes).


no92 (approved):
LGTM

qookei (approved):

PR #905, opened 2025-03-21
netserver: Add support for getsockopt to tcp sockets

Dennisbonke:


servers/netserver/src/ip/tcp4.cpp
```diff
@@ -520,6 +520,24 @@ struct Tcp4Socket {
...
+		int layer, int number, std::vector<char> &optbuf) {
+		auto self = static_cast<Tcp4Socket *>(object);
+		(void)self;
+
+		if(layer == SOL_SOCKET && number == SO_TYPE) {
+			int out = SOCK_STREAM;
+			optbuf.resize(std::min(optbuf.size(), sizeof(out)));
+			memcpy(optbuf.data(), &out, optbuf.size());
+		} else {
+			std::cout << std::format("netserver: unhandled TCP socket getsockopt layer {} number {}\n",
```
no92, line 533: We can use `std::println` here.

no92 (commented):
One style nit, rest LGTM.

PR #910, opened 2025-03-24
posix: UNIX socket timeouts

no92:


hel/include/helix/timer.hpp
```diff
@@ -77,15 +77,22 @@ struct TimeoutCancellation {
...
-	HEL_CHECK(await.error());
+	auto async_id = await.asyncId();
+
+	{
+		async::cancellation_callback cb{cancel, [&] {
+			HEL_CHECK(helCancelAsync(helix::Dispatcher::global().acquire(),
+					async_id));
+		}};
+		co_await submit.async_wait();
+	}
```
avdgrinten, line 95: The `HEL_CHECK` is deleted here. What about returning a `bool` to indicate success? So something like
```cpp
if (await.error() == kHelErrCancelled)
    return false;
HEL_CHECK(await.error());
return true;
```

posix/subsystem/src/un-socket.cpp
```diff
@@ -60,6 +61,26 @@ struct OpenFile : File {
...
+
+	std::function<async::result<void>(async::cancellation_token)> raceSendTimeout =
+			[&](async::cancellation_token c) -> async::result<void> {
+		if(sendTimeout_) {
+			uint64_t ns = (sendTimeout_->tv_usec * 1'000) + (sendTimeout_->tv_sec * 1'000'000'000);
+			co_await helix::sleepFor(ns, c);
+		} else {
+			co_await async::suspend_indefinitely(c);
+		}
+	};
```
avdgrinten, line 83: `std::function` should be removed in favor of a normal function that takes an `async::cancellation_token` and that returns a `async::result<void>`. I get that there is still the need for a lambda similar to: `[&] (async::cancellation_token c) { return raceSendTimeout(c); }` at the call site but that's fine.

posix/subsystem/src/un-socket.cpp
```diff
@@ -710,6 +727,18 @@ struct OpenFile : File {
...
+				co_return protocols::fs::Error::illegalArguments;
+
+			int val = *reinterpret_cast<int *>(optbuf.data());
+
+			if(val >= 0)
+				peekOffset_ = val;
+			else if(val == -1)
+				peekOffset_ = std::nullopt;
+			else
+				co_return protocols::fs::Error::illegalArguments;
```
avdgrinten, line 741: Does this match Linux? From the man page, it seems that any negative value results in `std::nullopt` behavior.

posix/subsystem/src/un-socket.cpp
```diff
@@ -286,9 +286,14 @@ struct OpenFile : File {
...
-		auto chunk = std::min(packet->buffer.size() - packet->offset, max_length);
-		memcpy(data, packet->buffer.data() + packet->offset, chunk);
+
+		size_t packet_offset = packet->offset;
+		if((flags & MSG_PEEK) && peekOffset_)
+			packet_offset += peekOffset_.value();
+
+		auto data_length = packet->buffer.size() - packet_offset;
+		auto chunk = std::min(packet->buffer.size() - packet_offset, max_length);
+		memcpy(data, packet->buffer.data() + packet_offset, chunk);
```
avdgrinten, line 296: This is not correct if the peek offset crosses packet boundaries.

PR #911, opened 2025-03-25
posix: basic pidfd implementation

no92:
Depends on managarm/mlibc#1320.

### DIFF
--- a/posix/subsystem/src/common.hpp
+++ b/posix/subsystem/src/common.hpp
@@ -3,7 +3,6 @@
 #include <iostream>
 #include <format>
 #include <iterator>
-#include <unordered_map>
 
 #include <helix/ipc.hpp>
 
@@ -13,6 +12,10 @@ struct StructName {
 	static StructName get(const char *type) {
 		static uint64_t idCounter = 1;
 		return StructName{type, idCounter++};
+	}
+
+	std::string type() const {
+		return {_type};
 	}
 
 	friend std::ostream &operator<< (std::ostream &os, const StructName &sn) {
--- a/posix/subsystem/src/pidfd.cpp
+++ b/posix/subsystem/src/pidfd.cpp
@@ -1,0 +1,95 @@
+#include <async/recurring-event.hpp>
+#include <helix/ipc.hpp>
+#include <string.h>
+#include <sys/epoll.h>
+#include <sys/signalfd.h>
+
+#include "pidfd.hpp"
+#include "process.hpp"
+
+namespace pidfd {
+
+void OpenFile::serve(smarter::shared_ptr<OpenFile> file) {
+	helix::UniqueLane lane;
+	std::tie(lane, file->passthrough_) = helix::createStream();
+	async::detach(protocols::fs::servePassthrough(std::move(lane),
+			smarter::shared_ptr<File>{file}, &File::fileOperations, file->cancelServe_));
+}
+
+OpenFile::OpenFile(std::weak_ptr<Process> proc, bool nonBlock)
+: File{FileKind::pidfd,  StructName::get("pidfd"), nullptr,
+	SpecialLink::makeSpecialLink(VfsType::regular, 0777), defaultPipeLikeSeek},
+	nonBlock_{nonBlock}, process_{std::move(proc)} { }
+
+async::result<frg::expected<Error, size_t>>
+OpenFile::readSome(Process *, void *, size_t) {
+	co_return Error::illegalArguments;
+}
+
+async::result<frg::expected<Error, PollWaitResult>>
+OpenFile::pollWait(Process *, uint64_t inSeq, int pollMask,
+		async::cancellation_token cancellation) {
+	(void)pollMask; // TODO: utilize mask.
+
+	auto p = process_.lock();
+	if(!p)
+		co_return PollWaitResult(1, EPOLLIN);
+
+	if(inSeq > 1) {
+		co_return Error::illegalArguments;
+	} else if(inSeq == 1) {
+		co_await async::suspend_indefinitely(cancellation);
+	} else {
+		while(p->notifyType() != NotifyType::terminated && !cancellation.is_cancellation_requested()) {
+			if(!isOpen())
+				co_return Error::fileClosed;
+
+			co_await p->awaitNotifyTypeChange(cancellation);
+		}
+	}
+
+	co_return PollWaitResult(1, p->notifyType() == NotifyType::terminated ? EPOLLIN : 0);
+}
+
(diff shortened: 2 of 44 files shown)


posix/subsystem/src/common.hpp
```diff
@@ -15,6 +14,10 @@ struct StructName {
 		return StructName{type, idCounter++};
 	}
 
+	std::string type() const {
+		return {_type};
+	}
```
avdgrinten, line 19: While exposing this for debugging reasons is fine, it is not a good idea to use the `StructName` for "is instance of" queries. To achieve the later, we can add something like this:
```cpp
enum class FileKind {
    undefined,
    pidfd,
    epoll,
    // ...
};

class File {
    File(FileKind kind, /* ... */)
    : kind_{kind}, /* ... */

    FileKind kind() { return kind_; }

    // ...

private:
    FileKind kind_;
};
```

posix/subsystem/src/process.hpp
```diff
@@ -702,7 +702,10 @@ struct Process : std::enable_shared_from_this<Process> {
 	std::vector<std::shared_ptr<Process>> _children;
 
 	// The following intrusive queue stores notifications for wait().
+public:
 	NotifyType _notifyType;
+	async::recurring_event _notifyTypeChange;
```
avdgrinten, line 707: This should not be exposed like that. We can add a function to wait for a notification instead.

posix/subsystem/src/pidfd.cpp
```diff
@@ -0,0 +1,79 @@
...
+async::result<frg::expected<Error, PollWaitResult>>
+OpenFile::pollWait(Process *, uint64_t inSeq, int pollMask,
+		async::cancellation_token cancellation) {
+	(void)pollMask; // TODO: utilize mask.
+
+	auto p = process_.lock();
+	if(!p)
+		co_return PollWaitResult(1, EPOLLIN);
+
+	assert(inSeq == 0);
```
avdgrinten, line 39: This needs to return an error. IIRC it's `EINVAL` but please double check with other implementations.

posix/subsystem/src/requests.cpp
```diff
@@ -384,6 +384,16 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 				wait_pid = req->id();
 			} else if(req->idtype() == P_ALL) {
 				wait_pid = -1;
+			} else if(req->idtype() == P_PIDFD) {
+				auto fd = self->fileContext()->getFile(req->id());
+				if(!fd || fd->structName().type() != "pidfd") {
```
avdgrinten, line 389: See other comment about `structName()`. Also, doesn't the `!fd` case return a different error? I would expect `BADF` in that case but please double check against Linux.

posix/subsystem/src/un-socket.cpp
```diff
@@ -615,6 +617,26 @@ struct OpenFile : File {
...
+			int result = 0;
+
+			if(!pid) {
+				result = -ENODATA;
+			} else {
+				auto remoteProc = Process::findProcess(pid);
+
+				if(remoteProc) {
+					auto proc = Process::findProcess(_ownerPid);
+					assert(proc);
```
avdgrinten, line 634: What guarantees that this process exists?

posix/subsystem/src/requests.cpp
```diff
@@ -4033,6 +4033,50 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+		}else if(preamble.id() == managarm::posix::PidfdSendSignalRequest::message_id) {
+			auto req = bragi::parse_head_only<managarm::posix::PidfdSendSignalRequest>(recv_head);
+
+			if (!req) {
+				std::cout << "posix: Rejecting request due to decoding failure" << std::endl;
+				break;
+			}
+
+			auto fd = self->fileContext()->getFile(req->pidfd());
+			if(!fd || fd->structName().type() != "pidfd") {
```
avdgrinten, line 4047: `!fd` case is probably different from file != pidfd case.

posix/subsystem/src/un-socket.cpp
no92: Before connecting two end the _ownerPid is always set, see `connect`.

posix/subsystem/src/pidfd.cpp
```diff
@@ -0,0 +1,79 @@
...
+		co_return PollStatusResult{1, EPOLLIN};
+
+	bool terminated = p->_notifyType == NotifyType::terminated;
+	co_return PollStatusResult(terminated, terminated ? EPOLLIN : 0);
+}
+
+async::result<std::string> OpenFile::getFdInfo() {
+	auto p = process_.lock();
+	if(!p)
+		co_return "Pid:\t-1\n";
```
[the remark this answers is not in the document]
no92: Yes.

posix/subsystem/src/requests.cpp
no92: Linux returns EBADF in both cases, this is also checked in the posix-testsuite now.

no92:
Addressed the comments + rebased on master.

posix/subsystem/src/un-socket.cpp
```diff
@@ -615,6 +617,29 @@ struct OpenFile : File {
...
+					auto proc = Process::findProcess(_ownerPid);
+					if(proc) {
+						auto pidfd = createPidfdFile(remoteProc, false);
+						result = proc->fileContext()->attachFile(pidfd);
+					} else {
+						result = -EINVAL;
+					}
+				} else {
+					result = -ENODATA;
+				}
```
avdgrinten, line 642: I think this is still broken. If I do `socketpair()`, then `fork()`, this implementation of `SO_PEERPIDFD` will insert an FD into the original process even if it's called in the child. I suspect that it should insert the FD into the child instead.

avdgrinten: This can be fixed by adding `imbueCredentials()` to the `getsockopt` request.

posix/subsystem/src/un-socket.cpp
```diff
@@ -615,6 +617,28 @@ struct OpenFile : File {
...
+			} else {
+				auto remoteProc = Process::findProcess(pid);
+
+				if(remoteProc) {
+					if(process) {
+						auto pidfd = createPidfdFile(remoteProc, false);
+						result = process->fileContext()->attachFile(pidfd);
+					} else {
+						result = -EINVAL;
+					}
```
avdgrinten, line 638: Can `process` even be null?

PR #915, opened 2025-03-31
rust: Start working on Hel bindings

48cf:
This PR introduces simple Rust bindings for the Hel API and an example program using them. The plan is to eventually be able to write servers and drivers in Rust.


rust/hel/src/queue.rs
```diff
@@ -0,0 +1,360 @@
...
+                )
+            })?;
+        }
+    }
+
+    /// Returns a reference to the head futex of the queue.
+    fn head_futex(&mut self) -> &AtomicI32 {
+        let queue = self.mapping.as_mut().unwrap();
+
+        unsafe { AtomicI32::from_ptr(&mut queue.headFutex) }
```
pitust, line 314: This is an SB/TB violation: obtaining a `&mut hel_sys::HelQueue` means that there is no other place where `mut hel_sys::HelQueue` is held in the system, which clearly isn't true because the futex used by the kernel.

rust/hel/src/queue.rs
```diff
@@ -0,0 +1,360 @@
...
+        let queue = self.mapping.as_ref().unwrap();
+
+        unsafe { queue.indexQueue.as_slice(1 << self.ring_shift) }
+    }
+
+    /// Returns a mutable reference to the index queue.
+    fn index_queue_mut(&mut self) -> &mut [i32] {
+        let queue = self.mapping.as_mut().unwrap();
+
+        unsafe { queue.indexQueue.as_mut_slice(1 << self.ring_shift) }
```
pitust, line 328: This is a bit subtle, but it is also a TB/SB violation, because the `&mut hel_sys::HelQueue` reference is obtained, which is invalid because the kernel holds at least a `&hel_sys::HelQueue` to access the `headFutex`.

rust/hel/src/submission.rs
```diff
@@ -0,0 +1,156 @@
...
+pub struct SimpleResult;
+
+impl Completion for SimpleResult {
+    /// Parses a completion result from the queue element.
+    fn from_queue_element(element: QueueElement) -> Result<Self> {
+        let data = element.data();
+
+        assert!(data.len() == size_of::<hel_sys::HelSimpleResult>());
+
+        let result = unsafe { data.as_ptr().cast::<hel_sys::HelSimpleResult>().read() };
```
pitust, line 34: It may be better to use `read_unaligned` (it doesn't come with a performance penalty in practice, so might as well). Also, it would be nice to have a safety comment here. Also, I recommend using `bytemuck` for this kind  of thing (although here it's C FFI so I'm not sure if it would work...).

rust/hel/src/submission.rs
```diff
@@ -0,0 +1,156 @@
...
+
+                if let Err(err) = self
+                    .submission
+                    .submit(self.state.executor.queue_handle(), context)
+                {
+                    // Submission failed, return the error
+                    // We need to drop the reference count of the operation state
+                    // object here to avoid leaking it
+
+                    drop(unsafe { Rc::from_raw(context as *const OperationState) });
```
pitust, line 114: This drop is unsound because there is no guarantee that an error return from `submit` implies that no operation was submitted. Really, `submit` should be `unsafe` and `Submission` should be an `unsafe trait`.

pitust (changes requested):
I'll look through it more carefully tomorrow, it's quite possible I missed stuff.

rust/hel/src/queue.rs
avdgrinten: Is this fixed by simply using `&raw mut queue.headFutex` here?

rust/hel/src/queue.rs
avdgrinten: I guess we could do this by working with pointers instead and by using `std::ptr::write()` to avoid the need for a reference.

rust/hel/src/submission.rs
avdgrinten: There's also the `zerocopy` crate which allows accessing data that is stored in `[u8]`.

rust/hel/src/queue.rs
avdgrinten: Also, is this even a safety issue? From the memory model's perspective, the kernel may as well hold a `*hel_sys::HelQueue` which doesn't run into aliasing issues.

avdgrinten: Actually, it's true that the kernel may write to `HelQueue`. However, this problem can be avoided by only getting a mutable reference to the index queue (and not to the entire `HelQueue`). The index queue is never written by the kernel.

rust/hel/src/submission.rs
```diff
@@ -0,0 +1,165 @@
...
+        assert!(data.len() >= size_of::<hel_sys::HelSimpleResult>());
+
+        // SAFETY: The data is guaranteed to contain enough bytes
+        // to read a [`hel_sys::HelSimpleResult`]`, but it may
+        // not be aligned, so we use `read_unaligned` to read it.
+        let result = unsafe {
+            data.as_ptr()
+                .cast::<hel_sys::HelSimpleResult>()
+                .read_unaligned()
+        };
```
avdgrinten, line 43: The kernel guarantees that results within the queue are properly aligned.

rust/hel/src/submission.rs
```diff
@@ -0,0 +1,161 @@
...
+/// Common trait for completion result types.
+pub trait Completion {
+    /// Parses the result from the queue element.
+    fn from_queue_element(element: QueueElement) -> Result<Self>
+    where
+        Self: Sized;
+}
+
+/// Common trait for all submissions.
+pub unsafe trait Submission {
```
pitust, line 19: What are the safety invariants for `Submission`?

pitust, line 19: What are the safety invariants for `Submission`?

rust/hel/src/submission.rs
```diff
@@ -0,0 +1,161 @@
...
+    /// Submits a submission to the queue.
+    /// SAFETY: The implementation must ensure that no operation
+    /// is submitted if the return value is an error.
+    unsafe fn submit(&self, queue_handle: &Handle, context: usize) -> Result<()>;
```
pitust, line 23: 1. I'd write it as `/// SAFETY: An operation may only be submitted if this function does not return an error`, probably?
2. Put an empty comment line (`///`) between the first doc comment and the second doc comment.
3. what are the safety invariants for callers?

rust/hel/src/submission.rs
```diff
@@ -0,0 +1,161 @@
...
+                // once the submission is completed.
+
+                let state_cloned = self.state.clone();
+                let context = Rc::into_raw(state_cloned) as usize;
+
+                // SAFETY: We expect the implementation to uphold the
+                // safety contract of the [`Submission`] trait. The following
+                // [`drop`] in the error case will drop the reference count
+                // of the operation state object and release the memory.
+                if let Err(err) = unsafe {
```
pitust, line 115: Split it into two safety comments: one about the `self.submission.submit` (what are the constraints on that, exactly?), and one about the drop below (which is pretty much what this) is

pitust (commented):
Looks mostly fine now, I think. The safety comments are written in a slightly weird way, though.

PR #921, opened 2025-04-06
posix: Remove various exceptions, replace by error returns

avdgrinten:
This avoids crashes in posix-subsystem that can be triggered from user code.


no92 (approved):
LGTM

PR #923, opened 2025-04-07
posix: unify timerfd and setitimer timers

no92:


posix/subsystem/src/interval-timer.hpp
```diff
@@ -0,0 +1,43 @@
...
+namespace posix {
+
+struct IntervalTimer {
+	IntervalTimer(uint64_t initial, uint64_t interval)
+		: initial_(initial), interval_(interval) {
+	}
+
+	virtual ~IntervalTimer() {
+		cancelEvt_.cancel();
+	}
```
avdgrinten, line 16: This looks strange. The arm() coroutine has a shared pointer to this anyway, so `IntervalTimer` will never be destructed while it is running.

posix/subsystem/src/interval-timer.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+	void setTime(uint64_t initial, uint64_t interval);
+	void getTime(timespec &initial, timespec &interval);
+	void getTime(timeval &initial, timeval &interval);
+	void reset();
+
+	// Called when the timer expires. The success argument indicates whether the timer
+	// expired (true) or if the wait was cancelled (false). The return value indicates
+	// whether the timer should be cancelled (false) or should be continued (true).
+	virtual bool raise(bool success) = 0;
```
avdgrinten, line 29: Don't we _always_ want to stop the timer loop if the timer has been cancelled? In other words, does the return value make sense? It seems arbitrary that the `raise()` function makes the decision to stop the timer loop.

posix/subsystem/src/interval-timer.hpp
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	virtual ~IntervalTimer() {}
+
+	// Static function that arms the timer and keeps it alive until expiration or cancellation.
+	static async::detached arm(std::shared_ptr<IntervalTimer> timer);
+
+	void setTime(uint64_t initial, uint64_t interval);
+	void getTime(timespec &initial, timespec &interval);
+	void getTime(timeval &initial, timeval &interval);
```
avdgrinten, line 21: I don't like the asymmetry here. It would be nicer to have a `void getTime(uint64_t &initial, uint64_t &interval)` together with conversion functions `nanosToTimespec()` / `nanosToTimeval()` (maybe in the core library?).

posix/subsystem/src/interval-timer.hpp
```diff
@@ -0,0 +1,41 @@
...
+	void reset();
+
+	// Called when the timer expires. The success argument indicates whether the timer
+	// expired (true) or if the wait was cancelled (false). The return value indicates
+	// whether the timer should be cancelled (false) or should be continued (true).
+	virtual void raise(bool success) = 0;
```
avdgrinten, line 27: Comment is outdated.

posix/subsystem/src/interval-timer.cpp
```diff
@@ -0,0 +1,86 @@
...
+	initial.tv_usec = (left % 1'000'000'000) / 1'000;
+	interval.tv_sec = interval_ / 1'000'000'000;
+	interval.tv_usec = (interval_ % 1'000'000'000) / 1'000;
+}
+
+void IntervalTimer::reset() {
+	cancelEvt_.cancel();
+	nextExpiration_ = 0;
+	initial_ = 0;
+	interval_ = 0;
```
avdgrinten, line 83: This is probably not necessary since callers will just stop using the `IntervalTimer` after `reset()`.

posix/subsystem/src/interval-timer.cpp
```diff
@@ -0,0 +1,86 @@
...
+void IntervalTimer::getTime(timespec &initial, timespec &interval) {
+	uint64_t now;
+	HEL_CHECK(helGetClock(&now));
+
+	uint64_t left = 0;
+	if(initial_ > now)
+		left = initial_ - now;
+
+	initial.tv_sec = left / 1'000'000'000;
+	initial.tv_nsec = left % 1'000'000'000;
```
avdgrinten, line 60: `initial` is supposed to return the time to the next expiration.

PR #928, opened 2025-04-18
posix: sigtimedwait + POSIX timers

no92:


posix/subsystem/src/observations.cpp
```diff
@@ -516,6 +516,72 @@ async::result<void> observeThread(std::shared_ptr<Process> self,
...
+						infoPtr, sizeof(siginfo_t), &siginfo);
+					HEL_CHECK(store.error());
+				}
+
+				gprs[kHelRegOut0] = 0;
+				gprs[kHelRegOut1] = check->signalNumber;
+			} else if(!timeout) {
+				// a timeout of 0 is non-blocking
+				gprs[kHelRegOut0] = EAGAIN;
+				gprs[kHelRegOut1] = 0;
```
avdgrinten: The extra cases for the `timeout = 0` case are probably not necessary.

posix/subsystem/src/process.hpp
```diff
@@ -245,6 +245,16 @@ struct SignalItem {
 using PollSignalResult = std::tuple<uint64_t, uint64_t>;
 using CheckSignalResult = std::tuple<uint64_t, uint64_t>;
 
+struct CompileSignalInfo {
+	void operator() (const UserSignal &info) const {
+		//si->si_code = SI_USER;
+		si->si_pid = info.pid;
+		si->si_uid = info.uid;
+	}
```
avdgrinten: Nit: the implementation can stay in the `.cpp` file.

posix/subsystem/src/clocks.hpp
```diff
@@ -0,0 +1,45 @@
...
+#include <assert.h>
+#include <core/clock.hpp>
+#include <hel.h>
+#include <stdint.h>
+#include <time.h>
+#include <print>
+
+namespace posix {
+
+inline uint64_t convertToNanos(const timespec &ts, clockid_t clock, bool relative = false) {
```
avdgrinten: There's no reason to have the definition of this function in the header file.

posix/subsystem/src/process.hpp
```diff
@@ -645,7 +645,45 @@ struct Process : std::enable_shared_from_this<Process> {
...
+		int signo;
+	};
+
+	struct PosixTimer : posix::IntervalTimer {
+		PosixTimer(std::weak_ptr<Process> proc, PosixTimerContext *ctx, uint64_t initial, uint64_t interval)
+			: posix::IntervalTimer(initial, interval), process_{proc}, ctx_{ctx} {}
+
+		void raise(bool success) override {
+			if(!success)
+			return;
```
avdgrinten: Indentation.

posix/subsystem/src/process.hpp
```diff
@@ -645,7 +645,45 @@ struct Process : std::enable_shared_from_this<Process> {
...
+
+			if(ctx_->tid) {
+				auto proc = process_.lock();
+				if(!proc) {
+					cancel();
+					return;
+				}
+				proc->signalContext()->issueSignal(ctx_->signo, TimerSignal{ctx_->clockid});
+			}
```
avdgrinten: Why can `tid` be unset? Isn't it pretty useless to have a timer that doesn't notify anything?

posix/subsystem/src/process.hpp
```diff
@@ -645,7 +645,45 @@ struct Process : std::enable_shared_from_this<Process> {
...
+	struct PosixTimer;
+
+	struct PosixTimerContext {
+		clockid_t clockid;
+		std::shared_ptr<PosixTimer> timer = {};
+		std::optional<int> tid = std::nullopt;
+		int signo;
+	};
+
+	struct PosixTimer : posix::IntervalTimer {
```
avdgrinten, line 649: Why is `PosixTimerContext` a separate struct (compared to `PosixTimer`)? The lifetimes look a bit whacky especially since `PosixTimerContext` uses manual new + delete.

posix/subsystem/src/requests.cpp
```diff
@@ -4099,6 +4100,68 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+
+					timerContext->timer = std::make_shared<Process::PosixTimer>(
+						targetProcess, timerContext, valueNanos, intervalNanos);
+					timerContext->timer->arm(timerContext->timer);
+				} else {
+					resp.set_error(managarm::posix::Errors::ILLEGAL_ARGUMENTS);
+				}
+			} else {
+				resp.set_error(managarm::posix::Errors::ILLEGAL_ARGUMENTS);
+			}
```
avdgrinten, line 4141: Indentation can be reduced by handling this case first.

posix/subsystem/src/process.hpp
no92: There is `SIGEV_NONE`, for which the man page states:

> Don't asynchronously notify when the timer expires. Progress of the timer can be monitored using timer_gettime(2).

no92:
Addressed the comments and rebased.

posix/subsystem/src/process.hpp
```diff
@@ -625,7 +637,45 @@ struct Process : std::enable_shared_from_this<Process> {
...
+			}
+		}
+
+		void expired() override {
+			ctx_->timer = {};
+		}
+
+	private:
+		std::weak_ptr<Process> process_;
+		PosixTimerContext *ctx_;
```
avdgrinten: What guarantees that this raw pointer does not go out of scope? This should either be a weak pointer, or it should be removed by copying the necessary fields (`tid`, `signo` etc.) directly into `PosixTimer`.

PR #931, opened 2025-04-21
Add aarch64 support for eir-uefi

marv7000:
Depends on managarm/bootstrap-managarm#507


kernel/eir/arch/arm/meson.build
```diff
@@ -6,11 +6,14 @@ eir_sources = [
 		'fault.S',
 		'fault.cpp',
 		'generic.cpp',
-		'../../system/dtb/discovery.cpp'
```
avdgrinten, line 9: This file is required for non-UEFI builds.

kernel/eir/arch/arm/meson.build
```diff
@@ -6,11 +6,16 @@ eir_sources = [
...
 	),
 	eir_generic_sources
 ]
 
+if build_uefi
+	eir_sources += files('platform/virt/virt.cpp')
```
qookei, line 14: I think this is fine for now but maybe we should at least add a comment saying this will need to be reworked to support booting with UEFI on real machines.

qookei (approved):
Seems good to be apart from the fact eir-uefi only works on QEMU's `-M virt`, but that's fine for now IMO. (not 100% sure how we want to deal with this yet either)

.gitignore
```diff
@@ -3,3 +3,4 @@ subprojects
 *.xbstrap
 .clangd
 compile_commands.json
+target
```
no92, line 6: How is this produced? This seems like something from your setup, so that should go into your `.git/info/exclude`, for instance.

kernel/eir/arch/arm/platform/virt/virt.cpp
```diff
@@ -21,9 +24,18 @@ static initgraph::Task prepareSerialForThor{
     }
 };
 
+static initgraph::Task initSerial{
```
no92, line 27: I'm not sure this is a good idea; it would be nice to have serial output with `debugPrintChar` as early as possible, especially on non-UEFI.

kernel/eir/arch/arm/meson.build
```diff
@@ -12,6 +12,8 @@ eir_sources = [
 
 if build_uefi
 	eir_sources += files('platform/virt/virt.cpp')
+else
+	eir_sources += files('../../system/dtb/discovery.cpp')
```
no92, line 17: This change should probably be squashed with the `eir: Add aarch64 support` commit.

.gitignore
marv7000: This is the default build artifact directory for Rust and IMO should be ignored in any case

kernel/eir/arch/arm/platform/virt/virt.cpp
marv7000: The alternative is to hardcode this in eirUefiMain

qookei (approved):

Dennisbonke (approved):
LGTM

PR #935, opened 2025-05-01
hel: Fix argument types and parameter qualifiers

48cf:
This makes it possible to include the hel headers from C source files. Also added a `kHelActionNone` constant because there wasn't one and it's useful for my WIP Rust bindings.


no92 (approved):
LGTM

PR #936, opened 2025-05-04
libblockfs untangling

qookei:
This PR attempts to untangle libblockfs by separating out the Ext2-specific code into it's own files and providing generic implementations that can be used by other file systems in the future.


drivers/libblockfs/src/ext2/ops.cpp
```diff
@@ -0,0 +1,290 @@
...
+}
+
+async::result<protocols::fs::Error> chmod(std::shared_ptr<void> object, int mode) {
+	auto self = std::static_pointer_cast<ext2fs::Inode>(object);
+	auto result = co_await self->chmod(mode);
+
+	co_return result;
+}
+
+async::result<protocols::fs::TraverseLinksResult> traverseLinks(std::shared_ptr<void> object,
```
avdgrinten, line 180: Is there anything that prevents us from having a common `traverseLinks`?

drivers/libblockfs/src/fs.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+
+namespace blockfs {
+
+using FlockManager = protocols::fs::FlockManager;
+using Flock = protocols::fs::Flock;
+
+struct BaseFileSystem;
+
+struct BaseInode {
```
avdgrinten, line 20: The `Base...` classes should have deleted copy / move constructors and assignment.

drivers/libblockfs/src/fs.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+	FlockManager flockManager;
+	std::unordered_set<std::string> obstructedLinks;
+};
+
+struct BaseFile {
+	BaseFile(std::shared_ptr<void> inode, bool append)
+	: inode{inode}, append{append} { }
+
+	const std::shared_ptr<void> inode;
```
avdgrinten, line 40: This can be `shared_ptr<BaseInode>`.

drivers/libblockfs/src/fs.hpp
```diff
@@ -0,0 +1,84 @@
...
+struct BaseFileSystem {
+	virtual const protocols::fs::FileOperations *fileOps() = 0;
+	virtual const protocols::fs::NodeOperations *nodeOps() = 0;
+
+	virtual std::shared_ptr<void> accessRoot() = 0;
+	virtual std::shared_ptr<void> accessInode(uint32_t inode) = 0;
+	virtual async::result<std::shared_ptr<void>> createRegular(int uid, int gid) = 0;
+
+	virtual ~BaseFileSystem() = default;
+};
```
avdgrinten, line 57: We use virtual functions here but we also have a `FileSystem` concept. Why is that done? Not necessarily something that we need to change but we should probably documented when we prefer methods in the concept vs. when we prefer virtual functions.

drivers/libblockfs/src/fs.hpp
```diff
@@ -0,0 +1,84 @@
...
+	bool append;
+};
+
+struct BaseFileSystem {
+	virtual const protocols::fs::FileOperations *fileOps() = 0;
+	virtual const protocols::fs::NodeOperations *nodeOps() = 0;
+
+	virtual std::shared_ptr<void> accessRoot() = 0;
+	virtual std::shared_ptr<void> accessInode(uint32_t inode) = 0;
+	virtual async::result<std::shared_ptr<void>> createRegular(int uid, int gid) = 0;
```
avdgrinten, line 54: It would be good to think about which functions in the `Base...` classes and in the concepts can fail. Then we can make them return `frg::expected`. It is not that important to fix the error handling in ext2 but if we add proper return types to the base functions, it is easier for new FSes to add proper error handling.

avdgrinten, line 54: These shared_ptrs can be `shared_ptr<BaseInode>`.

drivers/libblockfs/src/fs.hpp
```diff
@@ -0,0 +1,84 @@
...
+};
+
+template <typename T>
+concept Inode =
+	std::derived_from<T, BaseInode>
+	&& requires (T ino, std::optional<timespec> ts, size_t sz) {
+		{ ino.fileSize() } -> std::same_as<size_t>;
+		{ ino.accessMemory() } -> std::same_as<helix::BorrowedDescriptor>;
+		{ ino.updateTimes(ts, ts, ts) }; // TODO(qookie): Verify somehow that this is awaitable
+		{ ino.resizeFile(sz) }; // TODO(qookie): Verify somehow that this is awaitable
```
avdgrinten, line 66: Maybe we should still document the return types here.

drivers/libblockfs/src/fs.hpp
qookei: We're using virtual functions here instead of the concept because `servePartition` is not templated on the filesystem type, and it'd require a bit of untangling to change that, since `servePartition` handles both the FS-specific request, as well as stuff like `DEV_MOUNT` and partition block device `ioctl`s.

qookei: Ideally we'd move this stuff into the concept to avoid the virtual calls though.

drivers/libblockfs/src/ext2/ops.cpp
qookei: Nothing is preventing that other than the fact I haven't done it yet. :^)

drivers/libblockfs/src/fs.hpp
avdgrinten: Alright. I guess just adding this as a comment to the code is good enough.

drivers/libblockfs/src/fs.hpp
qookei: Done using `async::awaits_to` that was added a while ago.

drivers/libblockfs/src/fs.hpp
qookei: Added a TODO comment that these should eventually become part of the concept instead.

PR #937, opened 2025-05-04
Add scsi abstraction

Qwinci:
Add an abstraction for scsi and use it for usb-storage. Also fill the parent property of BlockDevice to fix a regression introduced in 9760ccf552b7e96bec7515f6fb5fe545b77fe733 where if the device would have no parent it would create the device link in the subsystem's sysfs twice resulting in an assert fail.

### DIFF
--- a/drivers/libblockfs/include/scsi.hpp
+++ b/drivers/libblockfs/include/scsi.hpp
@@ -1,0 +1,90 @@
+#pragma once
+
+#include <blockfs.hpp>
+
+#include <arch/dma_structs.hpp>
+#include <async/recurring-event.hpp>
+#include <frg/list.hpp>
+
+#include <string>
+
+namespace scsi {
+
+enum class ErrorType {
+	success,
+	checkCondition,
+	conditionMet,
+	busy,
+	reservationConflict,
+	taskSetFull,
+	acaActive,
+	taskAborted,
+	deviceSpecific
+};
+
+struct Error {
+	constexpr bool operator==(const Error &other) const = default;
+
+	std::string toString();
+
+	ErrorType type;
+	uint32_t code;
+};
+
+struct CommandInfo {
+	arch::dma_buffer_view command;
+	arch::dma_buffer_view data;
+	bool isWrite;
+};
+
+Error statusToError(uint8_t status);
+
+struct Interface {
+	virtual ~Interface() = default;
+	virtual async::result<frg::expected<Error, size_t>> sendScsiCommand(const CommandInfo &info) = 0;
+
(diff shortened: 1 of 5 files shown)


drivers/libblockfs/include/scsi.hpp
```diff
@@ -0,0 +1,84 @@
+#include <blockfs.hpp>
```
avdgrinten, line 1: `#pragma once`.

drivers/libblockfs/src/scsi.cpp
```diff
@@ -0,0 +1,279 @@
...
+		co_return result.error();
+	}
+
+	data[0] = std::byteswap<uint32_t>(data[0] & 0xffffffff);
+	co_return data;
+}
+
+async::detached StorageDevice::runScsi() {
+	while (true) {
+		if (!queue_.empty()) {
```
avdgrinten, line 146: Nitpick: the indentation level can be decreased by handling the empty case first and doing `continue` to re-try.

drivers/libblockfs/src/scsi.cpp
```diff
@@ -0,0 +1,279 @@
...
+					command.lba[1] = (req->sector >> 16) & 0xff;
+					command.lba[2] = (req->sector >> 8) & 0xff;
+					command.lba[3] = req->sector & 0xff;
+					command.transferLength[0] = req->numSectors >> 8;
+					command.transferLength[1] = req->numSectors & 0xff;
+
+					commandLength = sizeof(Read10);
+					memcpy(commandData, &command, sizeof(Read10));
+				} else {
+					throw std::logic_error("SCSI storage does not currently support high LBAs!");
```
avdgrinten, line 181: `throw`s can be converted to `logPanic` from the `core` library.

drivers/libblockfs/src/scsi.cpp
```diff
@@ -0,0 +1,279 @@
...
+		.command{nullptr, &command, sizeof(command)},
+		.data{nullptr, &lunListLength, 4},
+		.isWrite = false
+	};
+	auto result = co_await sendScsiCommand(info);
+	if (!result) {
+		co_return result.error();
+	}
+
+	lunListLength = std::byteswap(lunListLength);
```
avdgrinten, line 125: It probably makes sense to use endian-aware types from libarch. That also handles big endian hosts correctly (even though we don't support them right now).

avdgrinten (approved):
LGTM after minor comments are fixed.

avdgrinten (approved):

PR #939, opened 2025-05-06
Fix aarch64 gdbservers

Qwinci:
Fix the aarch64 register sizes in thor gdbserver and add support for aarch64 register read in the posix one.


qookei (approved):

no92 (approved):
LGTM

PR #944, opened 2025-05-19
Nvidia driver, part 2: fix IPC issues

no92:
This removes all uses of `helix::action` and removes it from the headers. The last commit came to be by checking all instances of recvInline and aggressively resetting the result to avoid a full queue.

### DIFF
--- a/hel/include/helix/ipc.hpp
+++ b/hel/include/helix/ipc.hpp
@@ -104,18 +104,12 @@ struct OperationBase {
 	friend struct Dispatcher;
 
 	OperationBase()
-	: _asyncId(0), _element(nullptr) { }
+	: _asyncId(0) { }
 
 	virtual ~OperationBase() { }
 
-	void *element() {
-		return _element;
-	}
-
 protected:
 	int64_t _asyncId;
-public: // TODO: This should not be public.
-	void *_element;
 };
 
 struct Operation : OperationBase {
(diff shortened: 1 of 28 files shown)


hel/include/helix/ipc.hpp
```diff
@@ -362,300 +356,109 @@ inline ElementHandle::ElementHandle(const ElementHandle &other) {
...
-		return result()->error;
-	}
-
-	UniqueDescriptor descriptor() {
-		HEL_CHECK(error());
-		return std::move(_descriptor);
-	}
+	// HelManageResult *result() {
+	// 	return reinterpret_cast<HelManageResult *>(OperationBase::element());
+	// }
```
avdgrinten, line 403: This should be removed (and likewise below).

hel/include/helix/ipc.hpp
```diff
@@ -362,300 +356,109 @@ inline ElementHandle::ElementHandle(const ElementHandle &other) {
...
-		return result()->error;
+		return result_.error;
 	}
 
 private:
-	HelSimpleResult *result() {
-		return reinterpret_cast<HelSimpleResult *>(OperationBase::element());
+	void parse(void *&ptr) override {
+		memcpy(&result_, ptr, sizeof(result_));
 	}
```
avdgrinten, line 378: Since we do not modify `ptr` anymore, we don't need to take a ref anymore.

avdgrinten: Also, we can make the base `parse()` function pure virtual since now all derived classes implement it.

Dennisbonke (approved):
LGTM

PR #947, opened 2025-05-22
posix: correctly organize class devices in sysfs

no92:
This correctly places devices in sysfs, so that e.g. card0 shows up under `/sys/devices/pci0000:00/0000:00:02.0/drm/card0`, unlike the previous incorrect `/sys/devices/pci0000:00/0000:00:02.0/card0` (notice the missing drm directory). This fixes some libdrm functions that previously just returned errors.

### DIFF
--- a/posix/subsystem/src/drvcore.cpp
+++ b/posix/subsystem/src/drvcore.cpp
@@ -78,7 +78,7 @@ public:
 // Device implementation.
 //-----------------------------------------------------------------------------
 
-Device::Device(std::shared_ptr<Device> parent, std::string name, UnixDevice *unix_device, Subsystem *subsys)
+Device::Device(std::shared_ptr<sysfs::Object> parent, std::string name, UnixDevice *unix_device, Subsystem *subsys)
 : sysfs::Object{parent ? parent : (subsys != nullptr ? subsys->object() : globalDevicesObject), std::move(name)},
 		_unixDevice{unix_device}, _parentDevice{std::move(parent)}, _subsystem{subsys} { }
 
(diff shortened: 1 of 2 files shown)


Dennisbonke (approved):
LGTM

posix/subsystem/src/drvcore.cpp
```diff
@@ -78,7 +78,7 @@ struct UeventAttribute : sysfs::Attribute {
 // Device implementation.
 //-----------------------------------------------------------------------------
 
-Device::Device(std::shared_ptr<Device> parent, std::string name, UnixDevice *unix_device, Subsystem *subsys)
+Device::Device(std::shared_ptr<sysfs::Object> parent, std::string name, UnixDevice *unix_device, Subsystem *subsys)
```
avdgrinten, line 81: Instead of making the parent device a `sysfs::Object`, I'd rather distinguish between parent device + parent directory.

posix/subsystem/src/drvcore.hpp
```diff
@@ -120,6 +125,22 @@ struct BusDriver : sysfs::Object {
...
+
+		auto it = parent->classDirectories().find(object()->name());
+		if(it != parent->classDirectories().end())
+			return it->second;
+
+		auto classDir = std::make_shared<sysfs::Object>(parent, object()->name());
+		classDir->addObject();
+		parent->classDirectories().insert({object()->name(), classDir});
+		return classDir;
+	}
```
avdgrinten, line 143: Does this take into account the fact that (on Linux) class dirs are not created if the parent is a class?

PR #955, opened 2025-06-12
build: use pkgconfig over symlinked subprojects for libasync/libsmarter

no92:
Depends on managarm/bootstrap-managarm#515.


Dennisbonke (approved):
LGTM

PR #957, opened 2025-07-02
SMBIOS support

no92:
This PR is enough to use `dmidecode` on managarm.


kernel/thor/system/smbios/smbios3.cpp
```diff
@@ -0,0 +1,228 @@
...
+};
+
+namespace smbios {
+
+void publish() {
+	if (smbios3) {
+		KernelFiber::run([=] { async::detach_with_allocator(*kernelAlloc, smbios3->run()); });
+	} else {
+		KernelFiber::run([=] { async::detach_with_allocator(*kernelAlloc, NoSmbios{}.run()); });
+	}
```
avdgrinten, line 197: Why is a `NoSmbios` mbus object useful?

kernel/thor/system/smbios/smbios3.cpp
```diff
@@ -0,0 +1,228 @@
...
+	    uintptr_t va = reinterpret_cast<uintptr_t>(ptr);
+
+	    KernelPageSpace::global().mapSingle4k(
+	        va, smbios3Ptr & ~(kPageSize - 1), page_access::read, CachingMode::null
+	    );
+
+	    smbios::Smbios3Header header = {};
+	    memcpy(&header, reinterpret_cast<void *>(va + smbios3Offset), sizeof(header));
+
+	    KernelPageSpace::global().unmapSingle4k(va);
```
avdgrinten: It's probably not guaranteed that the header doesn't cross a page boundary. Also, this can use `PhysicalWindow`.

kernel/thor/system/smbios/smbios3.cpp
```diff
@@ -0,0 +1,228 @@
...
+	    tableData.resize(tableSize);
+	    memcpy(
+	        tableData.data(),
+	        reinterpret_cast<void *>(reinterpret_cast<uintptr_t>(tablePtr) + tableOffset),
+	        tableSize
+	    );
+
+	    for (size_t off = 0; off < mappingSize; off += kPageSize) {
+		    KernelPageSpace::global().unmapSingle4k(reinterpret_cast<VirtualAddr>(tablePtr) + off);
+	    }
```
avdgrinten: This can also use `PhysicalWindow`.

kernel/thor/system/smbios/thor-internal/smbios/smbios.hpp
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+#include <stdint.h>
+
+namespace thor::smbios {
+
+struct [[gnu::packed]] Smbios3Header {
```
avdgrinten: This struct does not need to be packed (but `static_assert`ing its size would be a good idea).

protocols/hw/src/client.cpp
```diff
@@ -816,5 +816,65 @@ async::result<void> Device::enableDma() {
...
+			)
+		);
+
+	HEL_CHECK(offer.error());
+	HEL_CHECK(send_req.error());
+	HEL_CHECK(recv_head.error());
+
+	auto resp = *bragi::parse_head_only<managarm::hw::GetSmbiosHeaderReply>(recv_head);
+	if(resp.error() != managarm::hw::Errors::SUCCESS || !resp.size())
+		co_return {};
```
avdgrinten: The server side sends a buffer even if `resp.size() == 0`. It is a good idea to not make the message flow dependent on `resp.size()`.

kernel/klibc/eir/interface.hpp
```diff
@@ -56,6 +56,7 @@ struct EirInfo {
 	EirFramebuffer frameBuffer;
 
 	uint64_t acpiRsdp;
+	uint64_t smbios3;
```
avdgrinten: New code should use ELF notes to pass data to thor (see `THOR_DEFINE_ELF_NOTE`) instead of expanding `EirInfo`.

kernel/eir/protos/uefi/entry.cpp
```diff
@@ -105,17 +102,31 @@ initgraph::Task findAcpi{
...
+	    // acquire SMBIOS3 table info
+	    efi_guid smbios3_guid = SMBIOS3_TABLE_GUID;
+	    const efi_configuration_table *t = st->configuration_table;
+	    for (size_t i = 0; i < st->number_of_table_entries && t; i++, t++)
+		    if (!memcmp(&smbios3_guid, &t->vendor_guid, sizeof(smbios3_guid))) {
+			    eirSmbios3Ptr = reinterpret_cast<physaddr_t>(t->vendor_table);
+			    infoLogger() << "eir: Got SMBIOS3" << frg::endlog;
+		    }
+    }
+};
```
avdgrinten: It's probably a good idea to make a generic function to find a table by its GUID.

kernel/thor/system/smbios/smbios3.cpp
no92: It's useful for knowing that SMBIOS detection has completed, even if nothing was detected. This allows a driver to filter for an SMBIOS object on mbus and break out of the loop, either by having a table or the info that there is none. The concept is similar to what we already do for ACPI objects iirc.

PR #958, opened 2025-07-03
drivers/rtl8168: add support for MacVer34

ElectrodeYT:
This adds support for the RTL8168evl chipset. Seems to me to be very similar to MacVer35, but with slightly differently ordered init logic. I have not tested whether this order matters, but given that it has its own init function in the linux driver, I assume it likely does.

This works well enough to obtain a DHCP lease on my network and receive UDP broadcast packets.


no92 (approved):
LGTM

PR #959, opened 2025-07-03
drivers/uart: support kernel logging

ElectrodeYT:


drivers/uart/src/main.cpp
```diff
@@ -361,17 +395,17 @@ int main() {
 
 	// Set the baud rate.
 	base.store(uart_register::lineControl, line_control::dlab(true));
-	base.store(uart_register::baudLow, BaudRate::low9600);
 	base.store(uart_register::baudHigh, BaudRate::high9600);
```
no92: Why is this reordering done? If it's necessary according to spec or similar, it might be nice to leave a comment.

ElectrodeYT: I must admit, I honestly can't remember why I did this; probably was trying to debug something not working or something. I can't find any spec that says that it should be done in that order, so I've changed it back.

PR #967, opened 2025-07-23
eventfd tests + UAF fix

no92:


posix/subsystem/src/process.cpp
```diff
@@ -1320,18 +1320,18 @@ async::result<void> Process::terminate(TerminationState state) {
 	if(realTimer)
 		realTimer->cancel();
 
+	if(_procfs_dir) {
+		auto result [[maybe_unused]] = co_await _procfs_dir->getOwner()->unlink(_procfs_dir->getName());
+		assert(result);
+	}
```
avdgrinten, line 1327: Add a comment why we need to destroy this early. It doesn't need to be as detailed as the commit message.

PR #968, opened 2025-07-24
Add stubbed support for VT's

Dennisbonke:
This is a draft as it's not yet working, but this eases following along.


Dennisbonke:
Solved by #1186

PR #976, opened 2025-07-28
testsuites/posix-tests: allow specifying by globs which tests to run

no92:


avdgrinten (approved):
Nit: this would be slightly cleaner if an argument parser was used.

PR #982, opened 2025-07-29
drivers/libblockfs: Support non-native gpt block sizes

Qwinci:
Add support for reading the gpt on devices where the device block size doesn't match what the gpt was created with by probing for the gpt at different locations for the common block sizes and fix some hardcoded usages of 512 byte block size. This is useful for an example to boot a Managarm image created with the typical 512 byte block size using UFS which only supports hardware block sizes greater than 4kb.

### DIFF
--- a/drivers/libblockfs/src/gpt.cpp
+++ b/drivers/libblockfs/src/gpt.cpp
@@ -1,62 +1,109 @@
 
 #include <stdlib.h>
-#include <iostream>
+#include <print>
+#include <array>
 
 #include "gpt.hpp"
 
 namespace blockfs {
 namespace gpt {
 
+namespace {
+	constexpr std::array<size_t, 4> commonSectorSizes{512, 2048, 4096, 8192};
+
+	constexpr size_t gptSectorNumber = 1;
+}
+
 // --------------------------------------------------------
 // Table
 // --------------------------------------------------------
 
 Table::Table(BlockDevice *device)
-: device(device) { }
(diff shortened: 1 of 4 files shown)


drivers/libblockfs/src/gpt.cpp
```diff
@@ -1,62 +1,111 @@
...
 
-async::result<void> Table::parse() {
-	assert(getDevice()->sectorSize == 512);
+async::result<std::pair<void *, DiskHeader *>> Table::probeSectorSize_(size_t sectorSize) {
+	size_t deviceGptSector = (sectorSize * gptSectorNumber) / getDevice()->sectorSize;
+	size_t deviceGptOffset = (sectorSize * gptSectorNumber) % getDevice()->sectorSize;
+
+	size_t deviceSectors = (sectorSize + deviceGptOffset + getDevice()->sectorSize - 1) / getDevice()->sectorSize;
+
+	auto headerBuffer = static_cast<char *>(malloc(deviceSectors * getDevice()->sectorSize));
```
qookei: You could just use `new`/`delete` instead of `malloc`/`free`. Would also be nice to change the usage below.

qookei (commented):
I think the changes look good overall, just a small nit.

drivers/libblockfs/src/gpt.cpp
Qwinci: Done.

PR #985, opened 2025-08-01
posix: allow error returns for procfs attribute reads

no92:
This fixes the inconsistent boot failures that dereferenced a pointer to `0x40`.

### DIFF
--- a/posix/subsystem/src/process.cpp
+++ b/posix/subsystem/src/process.cpp
@@ -1295,6 +1295,9 @@ async::result<Error> Process::exec(std::shared_ptr<Process> process,
 }
 
 void Process::retire(Process *process) {
+	if(process->_procfs_dir)
+		process->_procfs_dir->unlinkSelf();
+
 	assert(process->_parent);
 	process->_parent->_childrenUsage.userTime += process->_generationUsage.userTime;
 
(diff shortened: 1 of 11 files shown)


Dennisbonke (approved):
LGTM

posix/subsystem/src/process.cpp
```diff
@@ -1294,7 +1294,12 @@ async::result<Error> Process::exec(std::shared_ptr<Process> process,
 	co_return Error::success;
 }
 
-void Process::retire(Process *process) {
+async::result<void> Process::retire(Process *process) {
+	if(process->_procfs_dir) {
+		auto result [[maybe_unused]] = co_await process->_procfs_dir->getOwner()->unlink(process->_procfs_dir->getName());
+		assert(result);
+	}
```
avdgrinten: Instead of making this function async, we can add a synchronous `directUnlink()` to `procfs::DirectoryNode`. We already have a `directMkdir()` for this purpose.

posix/subsystem/src/procfs.cpp
```diff
@@ -181,6 +190,13 @@ std::shared_ptr<FsNode> Link::getTarget() {
 	return _target;
 }
 
+void Link::unlinkSelf() {
+	assert(_target->getType() == VfsType::directory);
+
+	auto node = std::static_pointer_cast<DirectoryNode>(_target);
+	node->directUnlink(_name);
+}
```
avdgrinten: This looks wrong. Why does it unlink from the target and not from the owner?

avdgrinten: Also, this doesn't handle errors from `directUnlink()` (I suspect that `directUnlink()` fails here and we don't see that it fails because of that).

PR #986, opened 2025-08-01
Input device sysfs improvements

no92:


drivers/usb/devices/hid/src/main.cpp
```diff
@@ -715,9 +715,38 @@ async::detached HidDevice::run(proto::Device device, int config_num, int intf_nu
...
+		mbus_ng::Properties interfaceDescriptor = {
+			{"drvcore.mbus-parent", mbus_ng::StringItem{std::to_string(parentId_)}},
+			{"usb.interface_drivers", mbus_ng::ArrayItem{{
+				mbus_ng::ArrayItem{{
+					mbus_ng::StringItem{std::format("{}.{}", config_num, intf.num())},
+					mbus_ng::StringItem{"usbhid"},
+				}},
+			}}},
+		};
+		auto interfaceEntity = (co_await mbus_ng::Instance::global().createEntity("usb-interface", interfaceDescriptor)).unwrap();
```
avdgrinten, line 731: Seems that this mbus object does nothing and its `localLane` is immediately closed? Also, why is something generic like "usb-interface" created by the HID driver and not by the USB driver?

drivers/usb/hcds/xhci/src/main.cpp
```diff
@@ -261,6 +261,10 @@ Controller::enumerateDevice(std::shared_ptr<proto::Hub> parentHub, int port, pro
 	arch::dma_object<proto::DeviceDescriptor> descriptor{&_memoryPool};
 	FRG_CO_TRY(co_await device->readDescriptor(descriptor.view_buffer(), 0x0100));
 
+	arch::dma_object<proto::ConfigDescriptor> configDescriptor{&_memoryPool};
+	FRG_CO_TRY(co_await device->readDescriptor(configDescriptor.view_buffer(), 0x0200));
+	FRG_CO_TRY(co_await device->useConfiguration(0, configDescriptor->configValue));
```
avdgrinten, line 266: If we want to do this, we should do it for all HCDs (should be simple and can be tested in qemu). Maybe we should also validate this against what other OSes do if they have no driver for a device.

drivers/usb/devices/hid/src/main.cpp
no92: Opened as issue #992.

PR #987, opened 2025-08-02
Coredumping support

no92:


posix/subsystem/src/coredump.cpp
```diff
@@ -0,0 +1,475 @@
...
+	HelRegisterInfo regInfo;
+	HEL_CHECK(helQueryRegisterInfo(kHelRegsSimd, &regInfo));
+	return regInfo.setSize;
+}();
+
+bool dumpArea(VmContext::AreaAccessor &area) {
+	return !area.backingFile() || (area.isWritable() || area.isExecutable());
+}
+
+struct prstatus {
```
Dennisbonke: I believe there is a similar struct in mlibc for this purpose in sys/procfs.h? Wouldn't surprise me if the other structs defined here should just go into there and we use them from mlibc.

Dennisbonke (commented):
Did a quick glance, found something of interest.

drivers/libblockfs/src/libblockfs.cpp
```diff
@@ -506,7 +506,7 @@ async::result<protocols::fs::TraverseLinksResult> traverseLinks(std::shared_ptr<
 
 		if (component == "..") {
 			if (parent == self)
-				co_return std::make_tuple(nodes, protocols::fs::FileType::unknown, 0);
+				co_return std::make_tuple(nodes, protocols::fs::FileType::directory, processedComponents);
```
avdgrinten, line 509: Is `nodes` guaranteed to be empty in this case? If think that should be the case and we can `assert()` it.

posix/subsystem/src/observations.cpp
```diff
@@ -15,7 +15,10 @@ void dumpRegisters(std::shared_ptr<Process> proc) {
 	auto thread = proc->threadDescriptor();
 
 	uintptr_t pcrs[2];
-	HEL_CHECK(helLoadRegisters(thread.getHandle(), kHelRegsProgram, pcrs));
+	auto registerLoad = helLoadRegisters(thread.getHandle(), kHelRegsProgram, pcrs);
+	if(registerLoad == kHelErrNoDescriptor)
+		return;
+	HEL_CHECK(registerLoad);
```
avdgrinten: Why can this happen? This looks wrong (i.e., this looks as if posix is confused about whether the descriptor is still alive or not).

posix/subsystem/src/coredump.cpp
```diff
@@ -0,0 +1,475 @@
...
+#include "process.hpp"
+
+namespace {
+
+async::result<frg::expected<Error, smarter::shared_ptr<File, FileHandle>>> createFile(Process *proc, std::string name) {
+	PathResolver resolver;
+	resolver.setup(proc->fsContext()->getRoot(), proc->fsContext()->getWorkingDirectory(), name, proc);
+	auto resolveResult = co_await resolver.resolve(
+		resolvePrefix | resolveNoTrailingSlash);
+	assert(resolveResult);
```
avdgrinten: Missing error handling.

posix/subsystem/src/coredump.cpp
avdgrinten: Aside from Dennis's comment, this struct also has the wrong naming convention (lower case instead of CamelCase).

posix/subsystem/src/coredump.cpp
```diff
@@ -0,0 +1,475 @@
...
+	int pr_ppid;
+	int pr_pgrp;
+	int pr_sid;
+	timeval pr_utime;
+	timeval pr_stime;
+	timeval pr_cutime;
+	timeval pr_cstime;
+} prstatus;
+
+struct prpsinfo {
```
avdgrinten: Naming convention.

posix/subsystem/src/coredump.cpp
```diff
@@ -0,0 +1,475 @@
...
+	int pr_pid;
+	int pr_ppid;
+	int pr_pgrp;
+	int pr_sid;
+	char pr_fname[16];
+	char pr_psargs[80];
+};
+
+#if defined(__x86_64__)
+struct regs {
```
avdgrinten: Naming convention.

posix/subsystem/src/coredump.cpp
```diff
@@ -0,0 +1,475 @@
...
+	unsigned long fs_base;
+	unsigned long gs_base;
+	unsigned long ds;
+	unsigned long es;
+	unsigned long fs;
+	unsigned long gs;
+	int pr_fpvalid = 0;
+};
+
+struct fpregs {
```
avdgrinten: Naming convention.

posix/subsystem/src/coredump.cpp
```diff
@@ -0,0 +1,475 @@
...
+}
+
+async::result<void> Process::coredump(TerminationState state) {
+	if(!dumpable_)
+		co_return;
+
+	std::println("posix: writing coredump for process {}", pid());
+	auto fileResult = co_await createFile(this, std::format("core.{}", pid()));
+	if(!fileResult)
+		co_return;
```
avdgrinten, line 93: Should we log on errors or propagate errors?

posix/subsystem/src/coredump.cpp
```diff
@@ -0,0 +1,475 @@
...
+	noteSize += sizeof(Elf64_Nhdr) + 8 + ntFileSize(this, noteFiles);
+	noteSize = ntSizeAlign(noteSize);
+	// NT_PRFPREG
+	noteSize += sizeof(Elf64_Nhdr) + 8 + sizeof(fpregs);
+
+	size_t memoryDumpOffset = sizeof(Elf64_Ehdr) + (phdrCount * sizeof(Elf64_Phdr)) + noteSize;
+	memoryDumpOffset = (memoryDumpOffset + 0x1000 - 1) & ~0xFFF;
+
+	auto resizeResult = co_await file->truncate(memoryDumpOffset + memoryDumpSize);
+	assert(resizeResult);
```
avdgrinten: Missing error handling.

posix/subsystem/src/coredump.cpp
```diff
@@ -0,0 +1,475 @@
...
+
+	auto memory = co_await file->accessMemory();
+	assert(memory);
```
avdgrinten: Missing error handling.

posix/subsystem/src/coredump.cpp
```diff
@@ -0,0 +1,475 @@
...
+
+
+	for(auto area : *vmContext()) {
+		if(!dumpArea(area))
+			continue;
+
+		auto loadMemory = co_await helix_ng::readMemory(vmContext()->getSpace(),
+			area.baseAddress(), area.size(),
+			reinterpret_cast<void *>(reinterpret_cast<uintptr_t>(memoryDump) + fileOffset));
+		assert(!loadMemory.error());
```
avdgrinten: Missing error handling.

posix/subsystem/src/coredump.cpp
no92: Neither, according to `core(5)` not being able to create the file due to e.g. missing permissions just does not produce a coredump.

posix/subsystem/src/coredump.cpp
no92: This failing seems like a logic error, though.

drivers/libblockfs/src/libblockfs.cpp
no92: I don't think `nodes` needs to be necessarily empty here. On Linux, trying to traverse further up than root will just resolve to `/`, even if you later try to descend down again.

posix/subsystem/src/coredump.cpp
avdgrinten: Well, this reads arbitrary memory which may return `kHelErrFault` etc.

PR #991, opened 2025-08-03
drivers/freebsd-e1000: use pkgconfig instead of meson subprojects

no92:
Depends on managarm/bootstrap-managarm#543.


Dennisbonke (approved):
LGTM

PR #994, opened 2025-08-04
posix: correctly handle connect(2) for unix sockets

### LINKED ISSUE #842: posix: Unix datagram sockets incorrectly share connect semantics with stream sockets
For `SOCK_DGRAM`, `connect()` does not establish a connect (instead it sets the default remote address).

no92:
Fixes #842.


Dennisbonke (approved):
LGTM

PR #995, opened 2025-08-04
posix, thor: Implement RISC-V signal support

48cf:


kernel/thor/generic/hel.cpp
```diff
@@ -3684,8 +3709,8 @@ HelError helQueryRegisterInfo(int set, HelRegisterInfo *info) {
 #elif defined (__aarch64__)
 			outInfo.setSize = 35 * sizeof(uintptr_t);
 #elif defined (__riscv) && __riscv_xlen == 64
-			// TODO: Implement this.
-			return kHelErrUnsupportedOperation;
+			// PC + 30 GP registers (X0 is constant zero, no need to save it).
```
robotman2412: This seems incorrect; RISC-V has 31 integer GPRs (if you do not count x0); shouln't this size be 32 * uintptr_t?

kernel/thor/generic/hel.cpp
```diff
@@ -31,7 +31,13 @@
 
 using namespace thor;
 
+
 namespace {
+#if defined(__riscv) && __riscv_xlen == 64
+	// 32 FP registers (32 * 8 bytes) + FCSR (8 bytes).
+	constexpr size_t riscvFpStateSize = 0x108;
+#endif
```
avdgrinten: Maybe this should be moved to `Executor` (on RISC-V only). That way, it is harder to miss when we update the FPU code.

posix/subsystem/src/process.cpp
```diff
@@ -768,7 +765,11 @@ async::result<void> SignalContext::raiseContext(SignalItem *item, Process *proce
 	size_t totalFrameSize = simdStateSize + sizeof(SignalFrame);
 
 	// Store the current register stack on the stack.
+#if defined(__riscv) && __riscv_xlen == 64
+	assert(alignof(SignalFrame) == 16);
+#else
 	assert(alignof(SignalFrame) == 8);
+#endif
 	auto frame = alignFrame(totalFrameSize);
```
avdgrinten: Maybe the better assertion here would be `assert(frame & (alignof(SignalFrame) - 1) == 0)` on all archs.

PR #1004, opened 2025-08-07
posix: publish dev node for DRM devices in sysfs

no92:


avdgrinten:
LGTM but we may want to make this more general (= not duplicate it for all busses/classes etc).

PR #1011, opened 2025-08-10
Even more Raspberry Pi 4 and ARM fixes.

qookei:
TBD:
- [ ] Linker scripts cleanup
- [x] I-cache sync for exec. pages
- [ ] Cache flushes for DMA in XHCI

Obsoletes #829.

### DIFF
--- a/kernel/thor/generic/thor-internal/arch-generic/cursor.hpp
+++ b/kernel/thor/generic/thor-internal/arch-generic/cursor.hpp
@@ -13,8 +13,7 @@ namespace thor {
 
 template <typename T>
 concept CursorPolicy = requires (uint64_t pte, uint64_t *ptePtr,
-		PhysicalAddr pa, PageFlags flags, CachingMode cachingMode,
-		PageSpace *space, void (*fn)(void *),void *ctxt) {
+		PhysicalAddr pa, PageFlags flags, CachingMode cachingMode) {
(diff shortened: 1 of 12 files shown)


kernel/thor/generic/thor-internal/arch-generic/cursor.hpp
```diff
@@ -137,6 +139,7 @@ struct PageCursor {
 		ptEnt = Policy::pteBuild(pa, flags, cachingMode);
 
 		__atomic_store_n(currentPtePtr_(), ptEnt, __ATOMIC_RELAXED);
+		Policy::pteWriteBarrier();
```
avdgrinten, line 146: Shouldn't we do the barrier in the higher level functions that map multiple pages? So for example in `mapPresentPagesByCursor` and similar (and also in the kernel mapping functions)?

That way, we'd do fewer barriers if multiple pages are mapped at a time.

avdgrinten (approved):

kernel/thor/generic/thor-internal/arch-generic/cursor.hpp
qookei: Linux appears to do a barrier per write (although it also avoids barriers for user pages and instead lets the user task potentially fault again). Added a TODO about investigating this optimization.

PR #1012, opened 2025-08-13
posix: rework syscall cancellation to avoid passing around OneshotEvents

no92:


kernel/thor/generic/service.cpp
```diff
@@ -57,6 +57,17 @@ namespace stdio {
...
+			auto preamble = bragi::read_preamble(reqBuffer);
+			if (preamble.error()) {
+				infoLogger() << "thor: failed to decode preamble" << frg::endlog;
+				co_return;
+			}
+
+			if (preamble.id() != managarm::fs::CntRequest<KernelAlloc>::message_id) {
+				co_await DismissSender{conversation};
+				co_return;
+			}
```
avdgrinten: Log for bad requests to add some visibility that they happen.

posix/subsystem/src/process.cpp
```diff
@@ -956,16 +956,35 @@ Process::~Process() {
...
+	auto cancelId = __atomic_load_n(&cancelEventPtr->cancellationId, __ATOMIC_ACQUIRE);
+	if (cancelId != 0) {
+		HelHandle handle;
+		HEL_CHECK(helTransferDescriptor(cancelEventPtr->lane, _fileContext->getUniverse().getHandle(), kHelTransferDescriptorIn, &handle));
+		int fd = cancelEventPtr->fd;
+
+		if (fd == -1) {
+			cancelEventRegistry_.cancel(helix_ng::CredentialsView{credentials_}, cancelId);
+		} else {
+			assert(fd >= 0);
```
avdgrinten: This can be triggered from user code.

PR #1014, opened 2025-08-15
Implement ThreadGroups, make poll() cancellable

no92:


posix/subsystem/src/process.cpp
```diff
@@ -898,7 +898,9 @@ Generation::~Generation() {
 
 // PID 1 is reserved for the init process, therefore we start at 2.
 ProcessId nextPid = 2;
+ProcessId nextTid = 1;
```
avdgrinten: TIDs and PIDs use the same namespace. TID allocation needs to use `PidHull`.

posix/subsystem/src/requests.cpp
```diff
@@ -3033,20 +3033,36 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+					}),
+					async::lambda([cancellation = ct.value()](auto c) -> async::result<void> {
+						co_await async::suspend_indefinitely(c, cancellation);
+					}),
+					async::lambda([&](auto c) -> async::result<void> {
+						k = co_await epoll::wait(epfile.get(), events, 16, c);
+					})
+				);
 			}
+			self->cancelEventRegistry().removeEvent(self->credentials(), req.cancellation_id());
```
avdgrinten: `registerEvent` + `removeEvent` is easy to mess up. We should add a `scopedRegistration()` method that returns a RAII object that automatically does the unregistration on destruction.

posix/subsystem/src/process.hpp
```diff
@@ -439,26 +444,16 @@ struct Process : std::enable_shared_from_this<Process> {
...
-	PidHull *getHull() {
-		return _hull.get();
-	}
-
-	int pid() {
-		assert(_hull);
-		return _hull->getPid();
-	}
+	Process *getParent();
+	PidHull *getHull();
```
avdgrinten: `getHull()` should probably refer to the TID and callers should be changed to use `threadGroup()->getHull()` if the need the PID hull.

posix/subsystem/src/process.hpp
```diff
@@ -669,6 +665,89 @@ struct Process : std::enable_shared_from_this<Process> {
 		return {credentials_};
 	}
 
+private:
+	pid_t tid_;
```
avdgrinten: `tid` needs to be stored as a `std::shared_ptr<PidHull>` (see other comment).

posix/subsystem/src/process.cpp
```diff
@@ -1437,24 +1507,87 @@ Process::wait(int pid, WaitFlags flags, async::cancellation_token ct) {
...
+		auto oldGroup = process->tgPointer_.get();
+		oldGroup->threads_.erase(oldGroup->threads_.iterator_to(process));
+	}
+
+	process->tgPointer_ = shared_from_this();
+
+	if (!procfsLink_) {
+		auto procfs_root = std::static_pointer_cast<procfs::DirectoryNode>(getProcfs()->getTarget());
+		procfsLink_ = procfs_root->createProcDirectory(process);
+	}
```
avdgrinten: Why is this done in this function?

posix/subsystem/src/process.cpp
```diff
@@ -1437,24 +1507,87 @@ Process::wait(int pid, WaitFlags flags, async::cancellation_token ct) {
...
+
+		it = _children.erase(it);
+	}
+}
+
+void ThreadGroup::reassociateProcess(Process *process) {
+	if(process->tgPointer_ && process->tgPointer_.get() != this) {
+		auto oldGroup = process->tgPointer_.get();
+		oldGroup->threads_.erase(oldGroup->threads_.iterator_to(process));
+	}
```
avdgrinten: AFAICT re-association is not possible, is it? Rename to `associateProcess()` and assert that the thread group is not set yet.

posix/subsystem/src/observations.cpp
```diff
@@ -212,16 +212,29 @@ async::result<void> observeThread(std::shared_ptr<Process> self,
...
 
-			gprs[kHelRegError] = kHelErrNone;
-			gprs[kHelRegOut0] = child->pid();
+			if (cloneResult) {
+				newThread = cloneResult.value()->threadDescriptor().getHandle();
+
+				gprs[kHelRegError] = kHelErrNone;
+				gprs[kHelRegOut0] = cloneResult.value()->tid();
+			} else {
+				gprs[kHelRegError] = cloneResult.error() | toHelError;
```
avdgrinten: Instead of converting to `HelError`, the supercall can use a separate register to report POSIX errors.

posix/subsystem/src/process.cpp
```diff
@@ -946,16 +949,44 @@ std::shared_ptr<Process> Process::findProcess(ProcessId pid) {
...
-Process::Process(std::shared_ptr<PidHull> hull, Process *parent)
-: _parent{parent}, _hull{std::move(hull)},
-		_clientPosixLane{kHelNullHandle}, _clientFileTable{nullptr},
-		notifyType_{NotifyType::null} { }
+Process::Process(ThreadGroup *threadGroup, std::shared_ptr<PidHull> tidHull)
+: tgPointer_{threadGroup} {
+	if (!tidHull)
+		hull_ = std::make_shared<PidHull>(nextPid++);
+	else
+		hull_ = std::move(tidHull);
```
avdgrinten: This looks weird. Why is this allocated on demand?

posix/subsystem/src/process.cpp
```diff
@@ -899,6 +901,7 @@ Generation::~Generation() {
 // PID 1 is reserved for the init process, therefore we start at 2.
 ProcessId nextPid = 2;
 std::map<ProcessId, PidHull *> globalPidMap;
+std::map<ProcessId, PidHull *> globalTidMap;
```
avdgrinten: This should not be needed `globalPidMap` maps PID to `PidHull` and from `PidHull` you should be able to both get to the TG and to the `Process`.

posix/subsystem/src/process.cpp
```diff
@@ -1159,15 +1196,76 @@ std::shared_ptr<Process> Process::fork(std::shared_ptr<Process> original) {
...
+	CLONE_NEWNS | CLONE_SYSVSEM | CLONE_SETTLS | CLONE_PARENT_SETTID | CLONE_CHILD_CLEARTID |
+	CLONE_DETACHED | CLONE_UNTRACED | CLONE_CHILD_SETTID | CLONE_NEWCGROUP | CLONE_NEWUTS |
+	CLONE_NEWIPC | CLONE_NEWUSER | CLONE_NEWPID | CLONE_NEWNET | CLONE_IO);
+
+std::expected<std::shared_ptr<Process>, Error>
+Process::clone(std::shared_ptr<Process> original, void *ip, void *sp, posix::superCloneArgs *args) {
+	if (args->flags & unsupportedCloneFlags) {
+		std::println("posix: unexpected clone flags {:#x}", args->flags & unsupportedCloneFlags);
+		return std::unexpected{Error::illegalArguments};
+	}
```
avdgrinten: This should check for supported flags, not for unsupported ones.

posix/subsystem/src/process.cpp
```diff
@@ -1200,16 +1298,21 @@ std::shared_ptr<Process> Process::clone(std::shared_ptr<Process> original, void
...
 	process->_clientAuxEnd = original->_clientAuxEnd;
-	process->_uid = original->_uid;
-	process->_euid = original->_euid;
-	process->_gid = original->_gid;
-	process->_egid = original->_egid;
-	original->_children.push_back(process);
-	process->_hull->initializeProcess(process.get());
+
+	if (!(args->flags & CLONE_THREAD)) {
+		process->getPidHull()->initializeProcess(process.get());
```
avdgrinten: This looks weird since the PID of the TG now points to `process`. Is that correct? Do we ever want to call `initializeProcess()` on `getPidHull()` (instead of `getTidHull()`)?

posix/subsystem/src/process.cpp
```diff
@@ -1437,24 +1483,164 @@ Process::wait(int pid, WaitFlags flags, async::cancellation_token ct) {
...
+std::shared_ptr<ThreadGroup> ThreadGroup::init(std::shared_ptr<PidHull> hull) {
+	initThreadGroup = std::make_shared<ThreadGroup>(std::move(hull), nullptr);
+	return initThreadGroup;
+}
+
+ThreadGroup *ThreadGroup::create(std::shared_ptr<PidHull> hull, ThreadGroup *parent) {
+	auto tg = std::make_shared<ThreadGroup>(std::move(hull), parent);
+	parent->_children.push_back(tg);
+	return tg.get();
+}
```
avdgrinten, line 1509: Shouldn't this return a `shared_ptr`?

no92: I have this return a pointer as the ownership of the `ThreadGroup` lies with the parent `ThreadGroup`, where it is removed when it gets reaped.

posix/subsystem/src/process.cpp
```diff
@@ -1384,75 +1573,67 @@ async::result<void> Process::terminate(TerminationState state) {
...
-				continue;
-			if(std::holds_alternative<TerminationByExit>(it->_state) && !(flags & (waitExited)))
-				continue;
-			if(std::holds_alternative<TerminationBySignal>(it->_state) && !(flags & (waitExited)))
-				continue;
+	tg->parent_->_children.erase(std::ranges::find_if(tg->parent_->_children,
+	[pid = tg->pid()](auto e) {
+		return e->pid() == pid;
+	}));
+}
```
48cf: The formatting here looks weird.

PR #1016, opened 2025-08-25
thor: Disable I/O and memory decode only when reading BAR sizes

48cf:
This partially reverts e69de93e3c9a11dbe7b5ef6cc2e72b283a13ed29.


avdgrinten (approved):

avdgrinten:
It may be useful to add a comment on why we re-enable decode. For example:
`// Restore the original command byte value such that devices used by the kernel keep working (e.g., framebuffers or UARTs).`

PR #1022, opened 2025-08-27
Broadcom GENETv5 NIC driver

qookei:


servers/netserver/src/phy/generic.cpp
```diff
@@ -37,6 +37,26 @@ constexpr uint16_t anegComplete = 1 << 5; // Auto-Negotiation Complete
...
+	uint64_t startNs;
+	HEL_CHECK(helGetClock(&startNs));
+
+	while (bmcr & bmcr::reset) {
+		uint64_t currNs;
+		HEL_CHECK(helGetClock(&currNs));
+
+		if (currNs - startNs > 500'000'000) { // 500ms
+			std::println("generic-phy: Waiting for PHY reset timed out after 500ms");
+			break;
```
48cf: Shouldn't this return a timeout error instead of continuing with a log?

drivers/nic/bcmgenet/src/main.cpp
```diff
@@ -0,0 +1,610 @@
...
+		std::println("{} Unexpected link speed >1Gbps, assuming 1Gbps ({})", this, (int)phy_->speed());
+		speed = 2;
+	}
+
+	auto umacCmd = space_.load(reg::umac::cmd);
+	space_.store(reg::umac::cmd, umacCmd
+			/ umac::cmd::speed(speed));
+}
+
+void BcmGenetNic::setupBufs_() {
```
48cf, line 245: I understand that allocating many small buffers and keeping them alive in an `std::vector` was done to work around the physical allocator bug. Now that it was fixed maybe these could be allocated as one contiguous buffer and sliced as needed? Though FWIW it might be worth changing in the k1x-emac driver to do it this way as well to reduce the amount of contiguous physical memory allocations. Not necessarily an issue but I still wanted to bring it up.

48cf (approved):
LGTM

servers/netserver/src/phy/generic.cpp
qookei: Originally this was motivated by what NetBSD does, but after checking Linux, it seems to return an error if it times out, so let's do that instead.

drivers/nic/bcmgenet/src/main.cpp
qookei: I think smaller physical allocations are preferable, since larger ones could be unsatisfiable due to fragmentation, especially since the buddy allocator only hands out chunks with alignment == size.

avdgrinten (approved):

PR #1024, opened 2025-08-29
Misc aarch64 + pci fixes

Qwinci:


no92 (approved):
LGTM

kernel/eir/arch/arm/platform/virt/link.x
```diff
@@ -7,6 +7,7 @@ SECTIONS {
 	.text : ALIGN(0x1000) {
 		*(.text.init)
 		*(.text*)
+		*(.got .got.* .igot.*)
```
qookei: Why not do what `kernel/thor/link.x` does and put this in it's own section next to `.data`? (also this shouldn't be in `.text`?)

Qwinci: I moved it to its own section, though it doesn't really make a difference given that the result is a flat binary with no permission information anyway.

qookei (approved):

PR #1025, opened 2025-08-31
Clock/regulator support

Qwinci:
Add hw requests for getting a mbus entity by phandle, enable/disable/set frequency for clocks and enable/disable/set voltage for regulators.

### DIFF
--- a/protocols/hw/include/protocols/hw/client.hpp
+++ b/protocols/hw/include/protocols/hw/client.hpp
@@ -4,6 +4,7 @@
 #include <dtb.hpp>
 #include <helix/ipc.hpp>
 #include <frg/expected.hpp>
+#include <protocols/mbus/client.hpp>
 #include <memory>
 #include <optional>
 #include <vector>
@@ -216,5 +217,7 @@ struct Device {
 private:
 	helix::UniqueLane _lane;
 };
+
+async::result<mbus_ng::Entity> getEntityByPhandle(uint32_t phandle);
 
 } } // namespace protocols::hw
--- a/protocols/hw/meson.build
+++ b/protocols/hw/meson.build
@@ -4,7 +4,7 @@ incs = [ 'include', '../../kernel/common' ]
 src = [ 'src/client.cpp', hw_bragi ]
 
 if not provide_deps
-	deps = [ helix_dep, libarch, frigg ]
+	deps = [ helix_dep, mbus_proto_dep, core_dep, libarch, frigg ]
 
 	libhw = shared_library('hw_protocol', src,
 		dependencies : deps,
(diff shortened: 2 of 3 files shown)


protocols/hw/include/protocols/hw/client.hpp
```diff
@@ -157,11 +164,20 @@ struct Device {
...
 	async::result<helix::UniqueDescriptor> accessDtRegister(uint32_t index);
 	async::result<helix::UniqueDescriptor> installDtIrq(uint32_t index);
 
+	async::result<frg::expected<Error>> enableClock();
+	async::result<frg::expected<Error>> disableClock();
+	async::result<frg::expected<Error>> setClockFrequency(uint64_t frequency);
+
+	async::result<frg::expected<Error>> enableRegulator();
+	async::result<frg::expected<Error>> disableRegulator();
+	async::result<frg::expected<Error>> setRegulatorVoltage(uint64_t microvolts);
```
avdgrinten: How is the programming model supposed to work here? I assume that users need to call `setClockFrequency()` _before_ `enableClock()` (and likewise for regulators)? Should the kernel return an error if the user tries to switch the frequency of an enabled clock / change the voltage of an enabled regulator?

avdgrinten: We should add comments to indicate the programming model.

protocols/hw/include/protocols/hw/client.hpp
```diff
@@ -157,6 +157,7 @@ struct Device {
 
 	async::result<DtInfo> getDtInfo();
 	async::result<std::string> getDtPath();
+	async::result<std::optional<int64_t>> getDtEntityByPhandle(uint32_t phandle);
```
avdgrinten: Isn't mbus enumeration enough to solve this? We could add a phandle property to the mbus object and search for that.

Qwinci: hmm, I guess it is. Maybe it could be abstracted into a freestanding helper function in either hw or a new library?

protocols/hw/include/protocols/hw/client.hpp
Qwinci: > How is the programming model supposed to work here? I assume that users need to call `setClockFrequency()` _before_ `enableClock()` (and likewise for regulators)? Should the kernel return an error if the user tries to switch the frequency of an enabled clock / change the voltage of an enabled regulator?

It probably depends on the specific hardware in question, I think it might make sense to leave that up to the driver (so it would disable the clock if it is necessary for a freq change). Other related things that might be nice on the kernel side is to keep track of possible parent/child clocks + enable/disable refcounts for actual hw enable/disable in the generic interface so the generic code can adjust parent clock if necessary based on a flag the driver sets and only do actual hw clock enable/disable if needed.

> We should add comments to indicate the programming model.

Ill add them, though it might need to be changed (probably makes sense to try to write a driver using the API now to see if there are any immediate issues with it).

PR #1028, opened 2025-09-08
posix: Turn some asserts into logs

Dennisbonke:


posix/subsystem/src/process.cpp
```diff
@@ -1398,6 +1398,18 @@ async::result<void> Process::terminate(TerminationState state) {
...
+		std::cout << "posix: Invalid arguments for flags passed to process::wait, waitExited must be set" << std::endl;
+		co_return Error::illegalArguments;
+	}
+	if(flags & ~(waitNonBlocking | waitExited | waitLeaveZombie)) {
+		std::cout << "posix: Invalid arguments for flags passed to process::wait, unknown flags are set" << std::endl;
+		co_return Error::illegalArguments;
+	}
 	assert(pid == -1 || pid > 0);
 	assert(flags & waitExited);
 	assert(!(flags & ~(waitNonBlocking | waitExited | waitLeaveZombie)));
```
avdgrinten: We should also remove the asserts.

posix/subsystem/src/requests.cpp
```diff
@@ -488,6 +488,8 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 				resp.set_mode(mode);
 			} else if (waitResult.error() == Error::interrupted) {
 				resp.set_error(managarm::posix::Errors::INTERRUPTED);
+			} else if (waitResult.error() == Error::illegalArguments) {
+				resp.set_error(managarm::posix::Errors::ILLEGAL_ARGUMENTS);
```
avdgrinten: Can we convert the whole if chain to the error conversion helper?

posix/subsystem/src/process.cpp
Dennisbonke: Done.

posix/subsystem/src/requests.cpp
Dennisbonke: Done.

posix/subsystem/src/process.cpp
```diff
@@ -1398,9 +1398,18 @@ async::result<void> Process::terminate(TerminationState state) {
 
 async::result<frg::expected<Error, Process::WaitResult>>
 Process::wait(int pid, WaitFlags flags, async::cancellation_token ct) {
-	assert(pid == -1 || pid > 0);
-	assert(flags & waitExited);
-	assert(!(flags & ~(waitNonBlocking | waitExited | waitLeaveZombie)));
+	if(pid <= 0 && pid != -1) {
+		std::cout << "posix: Invalid arguments for pid passed to process::wait" << std::endl;
```
no92: This is not correct I'm afraid, as for waitpid the following behavior is supported when `pid < -1`: `meaning wait for any child process whose process group ID is equal to the absolute value of pid.`

posix/subsystem/src/process.cpp
```diff
@@ -1398,9 +1398,18 @@ async::result<void> Process::terminate(TerminationState state) {
...
+		co_return Error::illegalArguments;
+	}
+	if(!(flags & waitExited)) {
+		std::cout << "posix: Invalid arguments for flags passed to process::wait, waitExited must be set" << std::endl;
```
no92: `s/Invalid/Unsupported` because you can wait for `WCONTINUED` without `WEXITED` with waitid.

posix/subsystem/src/process.cpp
```diff
@@ -1398,9 +1398,18 @@ async::result<void> Process::terminate(TerminationState state) {
...
+		co_return Error::illegalArguments;
+	}
+	if(flags & ~(waitNonBlocking | waitExited | waitLeaveZombie)) {
+		std::cout << "posix: Invalid arguments for flags passed to process::wait, unknown flags are set" << std::endl;
```
no92: Same as above.

posix/subsystem/src/process.cpp
Dennisbonke: True, shall I change it to unsupported then? Same for the other 2?

no92 (approved):
LGTM

PR #1031, opened 2025-09-11
eir: Support booting in EL2 if VHE is supported

avdgrinten:
This enables boot (via the UEFI boot protocol) in EL2 if we have VHE.

### DIFF
--- a/kernel/eir/arch/arm/arch.cpp
+++ b/kernel/eir/arch/arm/arch.cpp
@@ -4,6 +4,8 @@
 #include <eir-internal/generic.hpp>
 #include <eir-internal/memory-layout.hpp>
 
+extern "C" void eirExcVectors();
+
 extern "C" [[noreturn]] void eirEnterKernel(uint64_t entry, uint64_t stack);
 
 extern "C" void
@@ -264,6 +266,15 @@ void initProcessorEarly() {
 	uint64_t currentel;
 	asm volatile("mrs %0, currentel" : "=r"(currentel));
 	eir::infoLogger() << "Starting Eir in EL " << (currentel >> 2) << frg::endlog;
+
+	// Install exception handlers (in case the boot protocol did not do that already).
+	auto vbar = reinterpret_cast<void *>(eirExcVectors);
+	if ((currentel >> 2) == 1) {
+		asm volatile("msr vbar_el1, %0" : : "r"(vbar) : "memory");
+	} else {
+		assert((currentel >> 2) == 2);
+		asm volatile("msr vbar_el2, %0" : : "r"(vbar) : "memory");
+	}
 }
 
 void initProcessorPaging() {


kernel/eir/arch/arm/arch.cpp
```diff
@@ -145,11 +147,80 @@ address_t getSingle4kPage(address_t address) {
...
+		if ((currentel >> 2) == 2)
+			panicLogger() << "eir: We are in EL2 but VHE is unsupported" << frg::endlog;
+	}
+
+	// Install exception handlers (in case the boot protocol did not do that already).
+	auto vbar = reinterpret_cast<void *>(eirExcVectors);
+	asm volatile("msr vbar_el1, %0" : : "r"(vbar) : "memory");
+
+	// Disable paging.
+	// We reprogram MAIR and HCR which could conflict with the current paging mode.
```
avdgrinten: As mentioned by @qookei: this is problematic since it effectively also disables caching and we may still have dirty cache lines. The better solution is probably to do this later in the initgraph when we have our own page tables available (then it would be enough to only flush Eir's pages to avoid issues).

avdgrinten:
This can be re-reviewed.

kernel/eir/arch/arm/arch.cpp
```diff
@@ -10,6 +10,41 @@ namespace eir {
...
+		uint64_t hcr_el2;
+		asm volatile("mrs %0, hcr_el2" : "=r"(hcr_el2) : : "memory");
+		infoLogger() << "eir: Entering VHE mode" << frg::endlog;
+		asm volatile(
+		    // clang-format off
+		    // Set the E2H bit.
+		    "msr hcr_el2, %0" "\n"
+		    "\t" "isb" "\n"
+		    // Flush the TLB since the E2H bit may be cached.
+		    "\t" "tlbi alle2" "\n"
```
avdgrinten: Actually, this hangs the Orion O6. I think we need to disable paging first.

kernel/eir/arch/arm/arch.cpp
```diff
@@ -4,8 +4,177 @@
...
 
+extern "C" void eirExcVectors();
+
+extern "C" [[noreturn]] void eirEnterKernel(uint64_t entry, uint64_t stack);
+
 namespace eir {
 
+namespace {
+
+void disableMmuEl1(uintptr_t flushStart, uintptr_t flushEnd, size_t dcLineSize) {
```
Qwinci: Maybe this would be cleaner as a naked function with a single asm block using a raw string and then directly using the arguments from the registers?

kernel/eir/arch/arm/arch.cpp
```diff
@@ -4,8 +4,177 @@
...
+	    :
+	    : "r"(flushStart), // %0
+	      "r"(flushEnd),   // %1
+	      "r"(dcLineSize), // %2
+	      "r"(sctlr)       // %3
+	    : "x0", "memory"
+	);
+}
+
+void disableMmuEl2(uintptr_t flushStart, uintptr_t flushEnd, size_t dcLineSize) {
```
Qwinci: Same here, maybe they could even be merged with an added `el` parameter given the only difference looks to be the sctlr register?

kernel/eir/arch/arm/arch.cpp
```diff
@@ -4,8 +4,177 @@
...
+	);
+}
+
+void enterE2h() {
+	infoLogger() << "eir: Entering VHE mode" << frg::endlog;
+	uint64_t hcr;
+	asm volatile("mrs %0, hcr_el2" : "=r"(hcr) : : "memory");
+	hcr |= UINT64_C(1) << 34;
+
+	asm volatile(
```
Qwinci: Maybe a raw string here too?

kernel/eir/arch/arm/arch.cpp
```diff
@@ -4,8 +4,177 @@
...
+		     "msr hcr_el2, %0" "\n"
+		"\t" "isb" "\n"
+		// Flush the TLB since the E2H bit may be cached.
+		"\t" "tlbi alle2" "\n"
+		"\t" "tlbi alle1" "\n"
+		"\t" "dsb ish" "\n"
+		"\t" "isb"
+	    // clang-format on
+	    :
+	    : "r"(hcr | (UINT64_C(1) << 34))
```
Qwinci: The `| (UINT64_C(1) << 34)` part is redundant as its already done in the above c++ code.

kernel/eir/arch/arm/arch.cpp
```diff
@@ -4,8 +4,177 @@
...
+	);
+}
+
+// Must only be called in either EL1 or in EL2 with E2H=1.
+void enterKernelPaging() {
+	uint64_t aa64mmfr0;
+	asm volatile("mrs %0, id_aa64mmfr0_el1" : "=r"(aa64mmfr0));
+
+	// Setup system registers for paging (MAIR and TCR).
+	if (aa64mmfr0 & (uint64_t(0xF) << 28))
```
Qwinci: This should be `(aa64mmfr0 & (uint64_t(0xF) << 28)) == 0xF` as a value of 1 indicates support for 52-bit VA's and not that 4kb granule is unsupported.
```suggestion
	if ((aa64mmfr0 & (uint64_t(0xF) << 28)) == 0xF)
```

kernel/eir/arch/arm/arch.cpp
```diff
@@ -203,7 +353,59 @@ void initProcessorPaging() {
...
+	asm volatile("mrs %0, currentel" : "=r"(currentel));
+
+	if (!physOffset) {
+		// Running from identity mapping. Paging may or may not be enabled.
+		// Reconfigure paging.
+		infoLogger() << "eir: Will reprogram MMU before jumping to kernel" << frg::endlog;
+
+		uint64_t ctr;
+		asm("mrs %0, ctr_el0" : "=r"(ctr));
+		auto dcLineSize = ((ctr >> 16) & 0b1111) << 4;
```
Qwinci: Shouldn't this be
```suggestion
		auto dcLineSize = 4 << ((ctr >> 16) & 0b1111);
```
as it stores `log2(size in 32-bit words)`?

kernel/eir/arch/arm/arch.cpp
```diff
@@ -4,8 +4,177 @@
...
+	               (2 << 12) |            // TTBR0 Inner shareable
+	               (2 << 28) |            // TTBR1 Inner shareable
+	               (uint64_t(pa) << 32) | // 48-bit intermediate address
+	               (uint64_t(2) << 30);   // TTBR1 4K granule
+
+	// TODO: If paging is already enabled, we should not overwrite MAIR and TCR
+	//       with potentially conflicting values here.
+	//       Instead, ensure that the current values are sane and error out if they are not.
+	//       This does not apply if paging is off.
+	asm volatile(
```
Qwinci, line 80: Maybe raw strings for the asms in this function too?

kernel/eir/arch/arm/eir-internal/arch/stack.hpp
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+namespace eir {
+
+inline void runFnPtrOnStack(void *ctx, void (*fn)(void *), void *sp) {
+	register void *x0 asm("x0") = ctx;
+	asm volatile(
```
Qwinci, line 7: Maybe raw string here too?

kernel/eir/arch/riscv/eir-internal/arch/stack.hpp
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+namespace eir {
+
+inline void runFnPtrOnStack(void *ctx, void (*fn)(void *), void *sp) {
+	register void *a0 asm("a0") = ctx;
+	asm volatile(
```
Qwinci, line 7: Same

kernel/eir/arch/x86/eir-internal/arch/stack.hpp
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+namespace eir {
+
+inline void runFnPtrOnStack(void *ctx, void (*fn)(void *), void *sp) {
+	register void *rdi asm("rdi") = ctx;
+	asm volatile(
```
Qwinci, line 7: Same

kernel/eir/arch/arm/arch.cpp
avdgrinten: How do raw strings help here and elsewhere?

Qwinci: Wouldn't need to add quotes/`\t`/`\n` escapes e.g.
```cpp
asm volatile(R"(
    mrs x0, hcr_el2
    // Set the E2H bit.
    orr x0, x0, 1 << 34
    msr hcr_el2, x0
    isb
    // Flush the TLB since the E2H bit may be cached.
    tlbi alle2
    tlbi alle1
    dsb ish
    isb
)" : : : "x0");
```

avdgrinten: That's true but at the cost that the generated asm is not indented correctly. Not sure what's the better trade off.

kernel/eir/arch/arm/arch.cpp
avdgrinten: Good catch.

kernel/eir/arch/arm/arch.cpp
avdgrinten: Yes, true, good catch.

PR #1032, opened 2025-09-13
eir: Handle boot framebuffer consistently on all boot protocols

avdgrinten:
This moves the boot framebuffer handling to its own file and it fixes a few small issues (in MB2 and Limine protocols).


marv7000 (approved):
LGTM

PR #1034, opened 2025-09-16
posix: Add NULL check for foreground process groups

marv7000:
Fix crash caused by spamming letters in Weston terminal


no92 (approved):
LGTM

PR #1038, opened 2025-09-19
eir: Determine logger from device tree

marv7000:


kernel/common/dtb.hpp
```diff
@@ -532,3 +534,29 @@ inline void DeviceTree::walkTree(T &&walker) {
 	rn.walkChildren(static_cast<T &>(walker));
 	walker.pop();
 }
+
+inline frg::optional<DeviceTreeNode> DeviceTree::findNode(frg::string_view path) {
+	// We only accept absolute paths.
+	if (path.size() < 1 || path[0] != '/')
+		return frg::null_opt;
```
avdgrinten: I think we should simply only strip the initial `/` if it exists. That way we can also handle relative paths (but they would still behave like absolute paths).

kernel/eir/system/uart/uart.cpp
```diff
@@ -59,4 +59,15 @@ void initFromAcpi(AnyUart &uart, unsigned int subtype, const acpi_gas &base) {
...
 
+void initFromDtb(AnyUart &uart, const DeviceTree &tree, const DeviceTreeNode &chosen) {
+	auto compatible = chosen.findProperty("compatible")->asString();
+
+#ifdef __aarch64__
+	if (*compatible == "arm,pl011") {
+		auto addr = chosen.findProperty("reg")->asU64(0);
+		uart.emplace<PL011>(addr, 0);
+	}
+#endif
```
avdgrinten: This `#ifdef` can be removed.

PR #1039, opened 2025-09-21
eir: Let Eir's UART subsystem pass boot UART tag

avdgrinten:
The move to `eir-linux` temporarily dropped the code to set the boot UART tag. Re-introduce code to do this but make it more generic this time (such that it works for DT and ACPI provided UARTs).

### DIFF
--- a/kernel/eir/system/uart/uart.cpp
+++ b/kernel/eir/system/uart/uart.cpp
@@ -1,31 +1,93 @@
+#include <eir-internal/arch.hpp>
+#include <eir-internal/main.hpp>
+#include <eir-internal/memory-layout.hpp>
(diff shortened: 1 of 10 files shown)


kernel/eir/system/uart/uart.cpp
```diff
@@ -1,7 +1,73 @@
+#include <eir-internal/arch.hpp>
+#include <eir-internal/main.hpp>
+#include <eir-internal/memory-layout.hpp>
 #include <eir-internal/uart/uart.hpp>
 
+// FIXME
+namespace eir {
+extern BootUartConfig bootUartConfig;
+}
```
avdgrinten: Just noticed that I forgot about this `FIXME`. Will amend this in a minute.

PR #1054, opened 2025-09-28
meta: enable more clang-tidy lints

no92:


drivers/libblockfs/src/libblockfs.cpp
```diff
@@ -507,8 +507,10 @@ async::result<protocols::fs::TraverseLinksResult> traverseLinks(std::shared_ptr<
 		if (component == "..") {
 			if (parent == self)
 				co_return std::make_tuple(nodes, protocols::fs::FileType::directory, processedComponents);
-
-			parent = self->fs.accessInode(FRG_CO_TRY(co_await parent->findEntry(".."))->inode);
+			auto entry = FRG_CO_TRY(co_await parent->findEntry(".."));
+			if (!entry)
+				co_return std::make_tuple(nodes, protocols::fs::FileType::directory, processedComponents);
```
avdgrinten: Shouldn't directories always have a `..` entry?

kernel/thor/generic/kernlet.cpp
```diff
@@ -88,7 +88,7 @@ void BoundKernlet::setupMemoryViewBinding(size_t index, void *p) {
 	if(logBinding)
 		infoLogger() << "thor: Binding memory view " << p
 				<< " to instance offset " << defn.offset << frg::endlog;
-	memcpy(_instance + defn.offset, &p, sizeof(void *));
+	memcpy(_instance + defn.offset, static_cast<void *>(&p), sizeof(void *));
```
avdgrinten, line 91: What does this fix? What lint is violated here?

posix/subsystem/src/process.cpp
```diff
@@ -375,7 +375,9 @@ std::shared_ptr<FileContext> FileContext::clone(std::shared_ptr<FileContext> ori
 
 	for(auto entry : original->_fileTable) {
 		//std::cout << "Clone FD " << entry.first << std::endl;
-		context->attachFile(entry.first, entry.second.file, entry.second.closeOnExec);
+
+		// the new number of fd in the file context is always <= the number of fd in the original
+		[[maybe_unused]] auto discard = context->attachFile(entry.first, entry.second.file, entry.second.closeOnExec);
```
avdgrinten: We should assert that the return value indicates success.

posix/subsystem/src/subsystem/usb/attributes.cpp
```diff
@@ -231,9 +231,8 @@ async::result<frg::expected<Error, std::string>> EpTypeAttribute::show(sysfs::Ob
 		case 1: co_return "Isochronous";
 		case 2: co_return "Bulk";
 		case 3: co_return "Interrupt";
+		default: __builtin_unreachable();
```
avdgrinten: `abort()` is probably better than `__builtin_unreachable()` for situations like this.

kernel/thor/generic/kernlet.cpp
no92: clangd complains about
> Multilevel pointer conversion from 'uacpi_handle *' (aka 'void **') to 'void *', please use explicit cast

drivers/libblockfs/src/libblockfs.cpp
no92: Replaced by an assert.

PR #1062, opened 2025-10-03
eir: Remove platform specific early init for raspi4

avdgrinten:
This generalizes the raspi4 early init by:
- Moving the Raspberry Pi specific mode setting to an initgraph task that matches the raspi4's `compatible` string.
- Makes the generic DT UART code work on the Raspberry Pi 4. This requires doing `ranges` translation in Eir.

Note that this PR does not yet remove the Raspberry Pi assembly entry point. I'll do that separately.


kernel/eir/arch/arm/platform/raspi4/raspi4.cpp
```diff
@@ -232,125 +233,105 @@ size_t getCmdline(void *dest)
...
+						    fb_width = *width;
+				    } else if (key == "bcm2708_fb.fbheight") {
+					    if (auto height = value.to_number<unsigned int>(); height)
+						    fb_height = *height;
+				    }
+			    }
+
+			    l = s;
+		    }
+	    }
```
48cf, line 305: Kind of a nitpick, but maybe update this to use `frg::parse_arguments`?

48cf (approved):

PR #1064, opened 2025-10-04
eir: Let raspi4 use eir-linux

### LINKED ISSUE #1042: eir: Merge Raspberry Pi 4 platform into eir-linux
Going forward, we want to have a single aarch64 image.

avdgrinten:
Resolve #1042. While the previous PR made the raspi4 code work inside eir-linux, we can now move the remaining low level code and drop the raspi4 platform's assembly entry point.


avdgrinten:
The code in `dropToEl1()` is a translation of the previous raspi4 code. We may want to set additional flags in `sctlr_el1` in the future but for now this should be fine.

Qwinci (approved):

PR #1066, opened 2025-10-06
Preparatory improvements for ppoll

no92:


posix/subsystem/src/epoll.cpp
```diff
@@ -106,11 +105,14 @@ struct OpenFile : File {
 		// This is the correct behavior for edge-triggered items.
 		// Level-triggered items stay pending until the event disappears.
 		auto result = resultOrError.value();
+		if ((std::get<1>(result) & ~(item->eventMask | EPOLLERR | EPOLLHUP)) != 0)
+			std::println("\e[1;31mposix.epoll {}: Item {} returned with invalid result {:#x} when valid mask is {:#x}\e[0m",
+				item->epoll->structName(), item->file->structName(), std::get<1>(result), (item->eventMask | EPOLLERR | EPOLLHUP));
```
avdgrinten: This message is a bit vague. Maybe: `"returned with invalid result {:#x}"` -> `"returned result {:#x} that is not contained in mask {:#x"}"`?

PR #1067, opened 2025-10-06
ppoll implementation + related fixes

no92:
Depends on a mlibc PR.

### DIFF
--- a/testsuites/posix-tests/src/sigaltstack.cpp
+++ b/testsuites/posix-tests/src/sigaltstack.cpp
@@ -1,7 +1,7 @@
-#include <cassert>
+#include <assert.h>
+#include <errno.h>
 #include <setjmp.h>
 #include <signal.h>
-#include <stdio.h>
 
 #include "testsuite.hpp"
 
@@ -52,3 +52,41 @@ DEFINE_TEST(sigaltstack, ([] {
 }))
 
 #endif
+
+DEFINE_TEST(sigaltstack_eperm, ([] {
+	if (setjmp(env)) {
+		int ret = sigaltstack(&old_ss, nullptr);
+		assert(ret == 0);
+
+		operator delete(ss.ss_sp);
+
+		return;
+	}
+
+	ss.ss_sp = operator new(SIGSTKSZ);
+	ss.ss_size = SIGSTKSZ;
+	ss.ss_flags = 0;
+
+	int ret = sigaltstack(&ss, &old_ss);
+	assert(ret == 0);
+
+	struct sigaction sa = {};
+	sigemptyset(&sa.sa_mask);
+	sa.sa_flags = SA_SIGINFO | SA_ONSTACK;
+	sa.sa_sigaction = [] (int, siginfo_t *, void *) {
+		stack_t current_ss = {};
+		current_ss.ss_sp = operator new(SIGSTKSZ);
+		current_ss.ss_size = SIGSTKSZ;
+		current_ss.ss_flags = 0;
+		int ret = sigaltstack(&current_ss, nullptr);
+		assert(ret == -1);
+		assert(errno == EPERM);
+		operator delete(current_ss.ss_sp);
+		longjmp(env, 1);
+	};
+
+	ret = sigaction(SIGUSR1, &sa, nullptr);
+	assert(ret == 0);
+
+	raise(SIGUSR1);
+}))


posix/subsystem/src/requests.cpp
```diff
@@ -2599,13 +2599,29 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+
+			std::set<int> defaultIgnoredSignals = {SIGCHLD, SIGURG, SIGWINCH};
+
 			SignalHandler saved_handler;
 			if(req.mode()) {
 				SignalHandler handler;
 				if(req.sig_handler() == uintptr_t(SIG_DFL)) {
 					handler.disposition = SignalDisposition::none;
+					if (defaultIgnoredSignals.contains(req.sig_number()))
+						co_await removePendingSignal(req.sig_number());
```
avdgrinten, line 2622: We should add a comment that this is demanded by POSIX.

posix/subsystem/src/requests.cpp
```diff
@@ -2599,13 +2599,29 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
 				}else if(req.sig_handler() == uintptr_t(SIG_IGN)) {
 					handler.disposition = SignalDisposition::ignore;
+					co_await removePendingSignal(req.sig_number());
```
avdgrinten, line 2626: We should add a comment that this is demanded by POSIX.

core/drm/src/core.cpp
```diff
@@ -174,20 +174,19 @@ drm_core::File::accessMemory(void *object) {
...
 	// Wait until we surpass the input sequence.
 	while(sequence == self->_eventSequence)
-		co_await self->_eventBell.async_wait();
+		if (!co_await self->_eventBell.async_wait(c))
+			break;
 
 	co_return protocols::fs::PollWaitResult{self->_eventSequence,
-			self->_eventSequence > 0 ? EPOLLIN : 0};
+			(self->_eventSequence > 0 ? EPOLLIN : 0) & mask};
 }
```
avdgrinten: Is this a correct implementation of the mask? Shouldn't you keep waiting if the result of the `&` is zero?

posix/subsystem/src/requests.cpp
```diff
@@ -3022,38 +3030,77 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+							co_await async::suspend_indefinitely(c, cancelEvent);
+							if (async::cancellation_token{cancelEvent}.is_cancellation_requested())
+								interrupted = true;
+						}),
+						async::lambda([&](auto c) -> async::result<void> {
+							if (req.has_signal_seq()) {
+								if (self->enteredSignalSeq() == req.signal_seq()) {
+									auto [newSeq, active] = co_await self->threadGroup()->signalContext()->pollSignal(req.signal_seq(), ~self->signalMask(), c);
+									if (active)
+										interrupted = true;
```
avdgrinten: Why is this needed (also in the other case below)?

posix/subsystem/src/process.cpp
```diff
@@ -1474,6 +1478,12 @@ async::result<Error> Process::exec(std::shared_ptr<Process> process,
 	process->_threadDescriptor = std::move(execResult.thread);
 	process->_vmContext = std::move(exec_vm_context);
 	process->threadGroup()->_signalContext->resetHandlers();
+	process->setAltStackEnabled(false);
+	process->setAltStackSp(0, 0);
+
+	if (process->inSignalHandler())
+		process->threadGroup()->signalContext()->restoreContext(process->threadDescriptor(), process.get());
```
avdgrinten: What does this do?

posix/subsystem/src/process.hpp
```diff
@@ -556,6 +556,14 @@ struct Process : std::enable_shared_from_this<Process> {
 		_altStackEnabled = en;
 	}
 
+	void setInSignalHandler(bool inHandler) {
+		inSignalHandler_ = inHandler;
+	}
+
+	bool inSignalHandler() {
+		return inSignalHandler_;
+	}
```
avdgrinten: Tracking signal handlers like this is problematic. Signal handlers can be nested. Also, processes can `sigjmp` out of signal handlers (and that's commonly done in practice). I don't think it is possible to reliably determine if we are in a signal handler or not.

posix/subsystem/src/requests.cpp
no92: We set `interrupted = true` as a flag so that we can later determine if the wait request was cancelled (due to signal delivery or a cancellation delivered over IPC).

We cannot avoid this by checking the return value of `epoll::wait()` alone, as timeout expirations do not return EINTR.

I'll push a fixup with some explanatory code comments soon.

PR #1075, opened 2025-10-08
thor: Refactor IRQ handling on aarch64

48cf:
This PR refactors aarch64 IRQ handling to not be GIC specific. This work is needed for the upcoming Apple interrupt controller driver.

### DIFF
--- a/kernel/thor/generic/irq.cpp
+++ b/kernel/thor/generic/irq.cpp
@@ -418,7 +418,7 @@ void IrqPin::_doService() {
 		if(anyAck) {
 			if(logService)
 				infoLogger() << "thor: IRQ pin " << name()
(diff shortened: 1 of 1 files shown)


kernel/thor/arch/arm/trap.cpp
```diff
@@ -0,0 +1,311 @@
...
+		);
+	}
+
+	return false;
+}
+
+constexpr bool logSGIs = false;
+constexpr bool logSpurious = false;
+
+void handleGicIrq(IrqImageAccessor image, ClaimedExternalIrq irq) {
```
qookei: `handleGicIrq` could probably be moved into `gic.cpp`?

PR #1078, opened 2025-10-10
ci-boot: Add utility

avdgrinten:
`ci-boot.py` takes commands from `vm-util.py` on `/dev/ttyS0` as newline-delimited JSON. This allows us to launch `kernel-tests` / `posix-tests` etc. on CI.

### DIFF
--- a/utils/ci-boot/ci-boot.target
+++ b/utils/ci-boot/ci-boot.target
@@ -1,0 +1,7 @@
+[Unit]
+Description=CI Boot
+Documentation=https://docs.managarm.org/
+Requires=basic.target
+Conflicts=rescue.service rescue.target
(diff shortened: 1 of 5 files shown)


utils/ci-boot/ci-boot.target
```diff
@@ -0,0 +1,7 @@
+[Unit]
+Description=CI Boot
+Documentation=https://docs.managarm.org/
+Requires=basic.target
```
Dennisbonke, line 4: We might want to depend on the sysinit target too? Or does basic already depend on that?

avdgrinten: Yes, `basic.target` depends on `sysinit.target`. I copied the dependencies of the new target from `multi-user.target`.

PR #1108, opened 2025-11-02
posix: Improve setpgid and setsid error handling

no92:
Improvments to posix-tests or to fix `os-test` tests.

### DIFF
--- a/posix/subsystem/src/requests.cpp
+++ b/posix/subsystem/src/requests.cpp
@@ -2720,8 +2720,9 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
 
 			managarm::posix::SvrResponse resp;
 
+			// POSIX: if the calling process is already a group leader, EPERM.
 			if(self->pgPointer()->getSession()->getSessionId() == self->pid()) {
-				co_await sendErrorResponse(managarm::posix::Errors::ACCESS_DENIED);
+				co_await sendErrorResponse(managarm::posix::Errors::INSUFFICIENT_PERMISSION);
 				continue;
 			}
 


testsuites/posix-tests/src/path-resolution.cpp
```diff
@@ -66,4 +68,6 @@ DEFINE_TEST(mkdir_trailing_dot, ([] {
 		close(fd);
 
 	rmdir("a");
+
+	setrlimit(RLIMIT_NOFILE, &oldlimit);
```
avdgrinten: This can use `scope_exit`.

posix/subsystem/src/requests.cpp
```diff
@@ -3815,29 +3815,46 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+			} else if (req->pid() > 0) {
 				target = Process::findProcess(req->pid());
-				if(!target) {
+				if (!target) {
+					co_await sendErrorResponse(managarm::posix::Errors::NO_SUCH_RESOURCE);
+					continue;
+				}
+
+				auto isSelf = req->pid() == self->pid();
+				auto isChild = (!isSelf && target->getParent()->pid() == self->pid());
```
avdgrinten: `getParent()` can be nullptr if this is called by init.

posix/subsystem/src/requests.cpp
```diff
@@ -3815,29 +3815,46 @@ async::result<void> serveRequests(std::shared_ptr<Process> self,
...
+			if (target->pid() == target->pgPointer()->getSession()->getSessionId()) {
 				co_await sendErrorResponse(managarm::posix::Errors::INSUFFICIENT_PERMISSION);
 				continue;
 			}
 
-			if(target->getParent()->pid() == self->pid()) {
-				if(target->didExecute()) {
+			// POSIX: if `pid` matches the process ID of a child and the child has successfully
+			// executed one of the `exec*` function, return EACCES.
+			if (target->getParent()->pid() == self->pid()) {
```
avdgrinten: Likewise, `getParent()` can be nullptr. Also, maybe this case should be folded into the `if` above where we retrieve the child?

PR #1112, opened 2025-11-04
netserver: Support TCP listening sockets

avdgrinten:
Code authored by @Qwinci (all credit goes to him). I factored out the TCP server code from FIN handling code that we also need to upstream.

### DIFF
--- a/servers/netserver/src/ip/tcp4.hpp
+++ b/servers/netserver/src/ip/tcp4.hpp
@@ -2,7 +2,7 @@
 
 #include <helix/ipc.hpp>
 #include <smarter.hpp>
-#include <map>
+#include <vector>
 
 class Ip4Packet;
 
@@ -19,10 +19,10 @@ struct Tcp4Socket;
 
 struct Tcp4 {
 	void feedDatagram(smarter::shared_ptr<const Ip4Packet>);
-	bool tryBind(smarter::shared_ptr<Tcp4Socket> socket, TcpEndpoint ipAddress);
+	bool tryBind(smarter::shared_ptr<Tcp4Socket> socket, bool unique, TcpEndpoint ipAddress);
 	bool unbind(TcpEndpoint remote);
 	void serveSocket(int flags, helix::UniqueLane lane);
 
 private:
-	std::map<TcpEndpoint, smarter::shared_ptr<Tcp4Socket>> binds;
+	std::multimap<TcpEndpoint, smarter::shared_ptr<Tcp4Socket>> binds;
 };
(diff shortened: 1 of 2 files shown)


protocols/fs/fs.bragi
```diff
@@ -95,6 +95,7 @@ enum CntReqType {
 	PT_GET_FILE_FLAGS = 30,
 	PT_SET_FILE_FLAGS = 31,
 	PT_PEERNAME = 42,
+	PT_ACCEPT = 51,
```
avdgrinten: I still want to convert this to a distinct message type instead of overloading the old CntRequest.

servers/netserver/src/ip/tcp4.hpp
```diff
@@ -19,10 +19,10 @@ struct Tcp4Socket;
...
 struct Tcp4 {
 	void feedDatagram(smarter::shared_ptr<const Ip4Packet>);
-	bool tryBind(smarter::shared_ptr<Tcp4Socket> socket, TcpEndpoint ipAddress);
+	bool tryBind(smarter::shared_ptr<Tcp4Socket> socket, bool unique, TcpEndpoint ipAddress);
 	bool unbind(TcpEndpoint remote);
 	void serveSocket(int flags, helix::UniqueLane lane);
 
 private:
-	std::map<TcpEndpoint, smarter::shared_ptr<Tcp4Socket>> binds;
+	std::multimap<TcpEndpoint, smarter::shared_ptr<Tcp4Socket>> binds;
```
avdgrinten, line 27: Note that I refactored the original code to use a `std::multimap` instead of a `std::vector` but we probably want to revise the binding logic in a separate step anyway (e.g., to support `SO_REUSEADDR` and to turn the multimap of (local port, local IP) pairs into a map of (local port, local IP, remote port, remote IP) tuples.

posix/subsystem/src/extern_socket.cpp
```diff
@@ -24,6 +24,65 @@ struct Socket : File {
...
+		req.set_req_type(managarm::fs::CntReqType::PT_LISTEN);
+
+		auto ser = req.SerializeAsString();
+		uint8_t buffer[128];
+
+		auto [offer, send_req, recv_resp] =
+			co_await helix_ng::exchangeMsgs(
+				getPassthroughLane(),
+				helix_ng::offer(
+					helix_ng::sendBuffer(ser.data(), ser.size()),
```
no92: This could use `sendBragiHeadOnly`.

posix/subsystem/src/extern_socket.cpp
```diff
@@ -24,6 +24,65 @@ struct Socket : File {
...
+					helix_ng::recvBuffer(buffer, 128)
```
no92: This could use `recvInline`.

PR #1113, opened 2025-11-06
netserver: allow non-blocking UDP sockets

ElectrodeYT:


servers/netserver/src/ip/udp4.cpp
```diff
@@ -492,6 +516,7 @@ struct Udp4Socket {
 	uint64_t _inSeq;
 
 	bool ipPacketInfo_ = false;
+	bool nonBlock_;
```
avdgrinten: This should probably get a default value to avoid pitfalls when refactoring. Other than that, LGTM.

ElectrodeYT: The TCP one doesn't have a default value either, and it's explicitly inited by the constructor, so I think it's personally fine, but I can add it if you want

servers/netserver/src/ip/udp4.cpp
```diff
@@ -441,6 +441,28 @@ struct Udp4Socket {
...
+		}
+		if(flags & O_NONBLOCK)
+			self->nonBlock_ = true;
+		else
+			self->nonBlock_ = false;
+		co_return;
+	}
+
+
+	static async::result<int> getFileFlags(void *object) {
```
no92, line 459: This should probably also OR in `O_RDWR`.

no92 (commented):
One issue

servers/netserver/src/ip/udp4.cpp
ElectrodeYT: Fixed

servers/netserver/src/ip/udp4.cpp
ElectrodeYT: I added it

PR #1130, opened 2025-11-15
OpenSSH work

no92:


posix/subsystem/src/fs.hpp
```diff
@@ -178,7 +178,7 @@ struct FsNode {
 	//! Opens the file (regular files only).
 	// TODO: Move this to the link instead of the inode?
 	virtual async::result<frg::expected<Error, smarter::shared_ptr<File, FileHandle>>>
-	open(std::shared_ptr<MountView> mount, std::shared_ptr<FsLink> link,
+	open(std::shared_ptr<Process> process, std::shared_ptr<MountView> mount, std::shared_ptr<FsLink> link,
 			SemanticFlags semantic_flags);
```
avdgrinten, line 182: This function can take a `Process *`. The caller needs to guarantee that the process survives the caller (by holding a `shared_ptr` across the call).

posix/subsystem/src/process.hpp
```diff
@@ -967,6 +967,7 @@ struct ControllingTerminalState {
 
 	TerminalSession *getSession() { return associatedSession_; }
 
+	DeviceId devId;
```
avdgrinten: This works but I'm not sure if this is a good approach. I'd prefer something like a `weak_ptr` to the `UnixDevice` but it seems that PTYs are not `UnixDevice`s yet.

PR #1131, opened 2025-11-15
posix/tmp_fs, libblockfs: fix seek EOF behaivour, posix/tmp_fs: implement semanticAppend

ElectrodeYT:


posix/subsystem/src/tmp_fs.cpp
```diff
@@ -632,9 +632,8 @@ MemoryFile::seek(off_t delta, VfsSeek whence) {
 	}else if(whence == VfsSeek::relative){
 		_offset += delta;
 	}else if(whence == VfsSeek::eof) {
-		assert(whence == VfsSeek::eof);
 		auto node = static_cast<MemoryNode *>(associatedLink()->getTarget().get());
-		_offset += delta + node->_fileSize;
+		_offset = delta + node->_fileSize;
 	}
```
avdgrinten, line 639: Technically unrelated but we should probably add an `else` that return an error here.

posix/subsystem/src/file.hpp
```diff
@@ -562,6 +562,8 @@ struct File : private smarter::crtp_counter<File, DisposeFileHandle> {
 	DefaultOps _defaultOps;
 
 	bool _isOpen;
+
+protected:
 	bool _append;
```
avdgrinten: It's weird that we have this in `File` but there are no users. We should move it into `MemoryFile` instead.

ElectrodeYT: It's set by other constructors too although it isn't used (and if removed from `File` and as a argument from the places that set it does compile); the correct thing to do might be to go through where it is set and seeing if we are missing handling for things

avdgrinten: At least for `extern_fs`, this flag can't be handled here so keeping it in the general `File` class may be confusing. Since going through all other users if probably out of scope of this PR, just moving it from `File` to `MemoryFile` probably makes most sense. We can open issues for the file that that do not correctly handle it.

ElectrodeYT: Done, I replaced `_append` with checking for semanticAppend in the existing flags

PR #1139, opened 2025-11-24
meta: use CLI11 via pkg-config instead of a meson subproject

no92:
Depends on managarm/bootstrap-managarm#611.


avdgrinten:
We can still use the subproject as a fallback.

PR #1144, opened 2025-11-25
posix: avoid using messy Linux timerfd headers

no92:


posix/subsystem/src/timerfd.cpp
```diff
@@ -24,9 +20,7 @@
 #include "protocols/fs/common.hpp"
 #include "timerfd.hpp"
 
-// avoid flock redefinition
-#define HAVE_ARCH_STRUCT_FLOCK
-#include <linux/timerfd.h>
+#define TFD_IOC_SET_TICKS _IOW('T', 0, uint64_t)
```
avdgrinten, line 24:
```suggestion
// Avoid <linux/timerfd.h> inclusion since that header is not compatible with libc.
#define TFD_IOC_SET_TICKS _IOW('T', 0, uint64_t)
```

PR #1148, opened 2025-12-01
Handle futex wait timeouts

no92:


kernel/thor/generic/thor-internal/timer.hpp
```diff
@@ -104,7 +104,7 @@ struct PrecisionTimerEngine final {
 	struct SleepOperation;
 
 	struct [[nodiscard]] SleepSender {
-		using value_type = void;
+		using value_type = bool;
```
avdgrinten, line 108: We should document what the return value means.

PR #1151, opened 2025-12-03
Implement futex waking with count

no92:


hel/include/hel.h
```diff
@@ -1118,7 +1118,9 @@ HEL_C_LINKAGE HelError helFutexWait(int *pointer, int expected, int64_t deadline
 //! Wakes up all waiters of a futex.
 //! @param[in] pointer
 //!     Pointer that identifies the futex.
-HEL_C_LINKAGE HelError helFutexWake(int *pointer);
+//! @param[in] count
+//!     Maximum number of waiters to wake.
+HEL_C_LINKAGE HelError helFutexWake(int *pointer, uint32_t count);
```
avdgrinten: It's more in line with the other syscalls to take an `unsigned int` here.

PR #1154, opened 2025-12-08
Split CoW mappings fix

no92:


posix/subsystem/src/process.cpp
```diff
@@ -221,6 +224,7 @@ async::result<void *> VmContext::remapFile(void *oldPointer,
 	area.copyView = std::move(it->second.copyView);
 	area.file = std::move(it->second.file);
 	area.offset = it->second.offset;
+	area.effectiveOffset = it->second.offset;
```
avdgrinten, line 228: It probably makes sense add a comment here: `TODO: This needs to be revised when we implement remapFile() for anonymous mappings.`.

avdgrinten (approved):

PR #1167, opened 2025-12-13
libblockfs: Handle multi-block access in ext2 bitmaps

### LINKED ISSUE #1094: EXT2 does not support multi-page blocks
Under certain conditions one can trigger [this assert](https://github.com/managarm/managarm/blob/0ff6e2cfa392321e3027cc510f18f89a610b4e6a/drivers/libblockfs/src/ext2fs.cpp#L563-L564). This has at least been observed during a crash of WebKitGTK while writing the coredump for this crash.

avdgrinten:
Fix #1094.

While doing the fix I also noticed that `handleInodeTable()` (which did not have a TODO) also didn't handle multi-block access correctly.

### DIFF
--- a/drivers/libblockfs/src/ext2/ext2fs.cpp
+++ b/drivers/libblockfs/src/ext2/ext2fs.cpp
@@ -12,6 +12,7 @@
 #include <async/result.hpp>
 #include <core/align.hpp>
 #include <core/clock.hpp>
+#include <core/logging.hpp>
 #include <helix/ipc.hpp>
 #include <helix/memory.hpp>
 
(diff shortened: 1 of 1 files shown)


drivers/libblockfs/src/ext2/ext2fs.cpp
```diff
@@ -761,29 +761,40 @@ async::detached FileSystem::manageInodeTable(helix::UniqueDescriptor memory) {
...
+			// TODO: Use shifts instead of division.
+			auto bg_idx = (manage.offset() + progress) / sizePerGroup;
+			auto bg_offset = (manage.offset() + progress) % sizePerGroup;
+			auto block = bgdt[bg_idx].inodeTable;
+			assert(block);
 
-		assert(bg_offset % device->sectorSize == 0);
-		assert(manage.length() % device->sectorSize == 0);
+			// Do not cross block group boundaries.
+			auto chunk = std::min(manage.length() - progress, sizePerGroup - bg_offset);
```
avdgrinten, line 782: We also need to ensure that `chunk` is page aligned. Otherwise, the `helUpdateMemory()` below may (try to) publish a partially initialized page.

PR #1193, opened 2026-01-15
supercall ostrace logging

no92:
Depends on managarm/bragi#11.

### DIFF
--- a/posix/subsystem/src/observations.cpp
+++ b/posix/subsystem/src/observations.cpp
@@ -9,7 +9,9 @@
 #include <protocols/posix/data.hpp>
 #include <protocols/posix/supercalls.hpp>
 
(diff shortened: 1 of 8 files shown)


posix/subsystem/src/observations.cpp
```diff
@@ -153,6 +180,10 @@ async::result<void> observeThread(std::shared_ptr<Process> self,
 			gprs[kHelRegOut0] = 0;
 			HEL_CHECK(helStoreRegisters(new_thread, kHelRegsGeneral, &gprs));
 
+			managarm::supercall::SupercallFork request;
+			request.set_pid(child->pid());
+			logSupercall(request);
```
avdgrinten: Should requests and responses be logged separately?

PR #1196, opened 2026-01-17
libblockfs: pass access mode through for open calls on nodes

no92:


protocols/fs/fs.bragi
```diff
@@ -885,3 +882,15 @@ head(128):
 	int32 fsid1;
 	uint64 flags;
 }
+
+message NodeOpenRequest 47 {
+head(128):
+	byte append;
+	byte read;
+	byte write;
+}
```
avdgrinten, line 891: We may want to expose posix-subsystem's semanticFlags here (by moving them into the fs protocol).

avdgrinten (approved):

PR #1202, opened 2026-01-21
eir, thor: Rework Eir to map per-CPU region and enable profiling on all CPUs

avdgrinten:
The per-CPU region mapping change makes it easy to enable profiling on remote CPUs since it makes sure that the per-CPU work queue is already up when we want to enable profiling.

### DIFF
--- a/kernel/thor/generic/profile.cpp
+++ b/kernel/thor/generic/profile.cpp
@@ -48,8 +48,8 @@ void initializeProfile() {
 	globalProfileRing.initialize(reinterpret_cast<uintptr_t>(profileMemory), 1 << 20);
 
 	// Dump the per-CPU profiling data to the global ring buffer.
-	// TODO: Start one such fiber per CPU.
-	KernelFiber::run([=] {
+	auto fiberMain = [] {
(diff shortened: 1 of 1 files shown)


kernel/thor/arch/arm/cpu.cpp
```diff
@@ -211,13 +211,14 @@ void prepareCpuDataFor(CpuData *context, int cpu) {
 }
 
 void setupBootCpuContext() {
-	auto context = &cpuData.getFor(0);
+	for (size_t c = 0; c < getCpuCount(); ++c)
```
Dennisbonke, line 214: Isn't it weird to do this for all
CPU's in `setupBootCpuContext`? Note that I don't have a better name or location but just something that struck me as odd. Same goes for the other architectures.

Qwinci (approved):
LGTM

kernel/thor/arch/arm/cpu.cpp
avdgrinten: That's a good point. I'll make a note to refactor this but merge this PR as-is. We can probably even move this to generic code instead of arch-specific code.

PR #1209, opened 2026-01-29
thor: Make coroutines WQ-affine

avdgrinten:
This PR changes `coroutine<>` and coroutines created by `enable_detached_coroutine` to always resume on the `WorkQueue` that they belong to.
- This is achieved by using `await_transform()` to augment each `co_await`: instead of libasync's `sender_awaiter`, we now use `WorkQueueAffineAwaiter` which handles this logic.
- Since coroutines now do not change work queues, we can get rid of most `co_await wq->schedule()` calls (they are still needed in sender/receiver logic + uses of libasync algorithms though).
- For `coroutine<>`, the `WorkQueue` pointer is determined by querying it from the receiver (via `async::execution::get_env()`). For `enable_detached_coroutine` there is no receiver and the `WorkQueue *` is passed to `enable_detached_coroutine`.
- The `WorkQueue` changes optimize submission from the `WorkQueue`'s context. This mitigates the overhead of posting to the queue once per `co_await`.

### DIFF
--- a/kernel/thor/generic/thor-internal/coroutine.hpp
+++ b/kernel/thor/generic/thor-internal/coroutine.hpp
@@ -113,17 +113,6 @@ protected:
 	~coroutine_continuation() = default;
 };
 
-// "Control flow path" that the coroutine takes. This state is used to distinguish inline
-// completion from asynchronous completion. In contrast to other state machines, the states
-// are not significant only their own; only transitions matter:
-// On past_suspend -> past_start transitions, we continue inline.
-// On past_start -> past_suspend transitions, we call resume().
-enum class coroutine_cfp {
-	indeterminate,
-	past_start, // We are past start_inline().
-	past_suspend // We are past final_suspend.
-};
-
 template<typename T, typename R>
 struct coroutine_operation;
 
@@ -191,13 +180,7 @@ struct coroutine {
 				}
 
 				void await_suspend(std::coroutine_handle<void>) noexcept {
-					auto cfp = promise_->cfp_.exchange(coroutine_cfp::past_suspend,
-							std::memory_order_release);
-					if(cfp == coroutine_cfp::past_start) {
-						// We do not need to synchronize with the thread that started the
-						// coroutine here, as that thread is already done on its part.
-						promise_->cont_->resume();
-					}
+					promise_->cont_->resume();
 				}
 
 				void await_resume() noexcept {
@@ -218,7 +201,6 @@ struct coroutine {
 	private:
 		thor::WorkQueue *wq_{nullptr};
 		coroutine_continuation<T> *cont_ = nullptr;
-		std::atomic<coroutine_cfp> cfp_{coroutine_cfp::indeterminate};
 	};
 
 	coroutine()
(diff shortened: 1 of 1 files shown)


kernel/thor/generic/load-balancing.cpp
```diff
@@ -32,7 +32,7 @@ LoadBalancer::LoadBalancer()
 void LoadBalancer::setOnline(CpuData *cpu) {
 	auto *node = &lbNode.get(cpu);
 	node->cpu = cpu;
-	async::detach_with_allocator(*kernelAlloc, loadBalancer->run_(cpu));
+	spawnOnWorkQueue(*kernelAlloc, WorkQueue::generalQueue().lock(), loadBalancer->run_(cpu));
```
qookei: Shouldn't this be `cpu->generalWorkQueue.lock()`? To run load balancing on the respective CPU's work queues, as was done in `run_` before.

kernel/thor/generic/memory-view.cpp
```diff
@@ -373,7 +369,6 @@ struct ZeroMemory final : MemoryView {
 
 	coroutine<frg::expected<Error>> copyFrom(uintptr_t, void *buffer, size_t size,
 			WorkQueue *wq) override {
-		co_await wq->enter();
```
qookei, line 376: Given coroutines do not resume on a WQ during initial suspend (#1214), doesn't this mean the copy can happen outside the work queue if the caller also didn't suspend yet?

kernel/thor/generic/thor-internal/coroutine.hpp
```diff
@@ -354,10 +413,20 @@ async::sender_awaiter<coroutine<T>, T> operator co_await(coroutine<T> s) {
 	return {std::move(s)};
 }
 
+namespace detail {
+	template<typename T>
+	constexpr T &lastArg(T &t) { return t; }
+
+	template<typename Head, typename... Tail>
+	constexpr auto &lastArg(Head &, Tail &... tail) { return lastArg(tail...); }
```
qookei, line 421: Is our clang new enough for pack indexing? Then this would just be something along the lines of:
```cpp
template <typename ...Ts>
constexpr auto &lastArg(Ts &...ts) { return ts...[sizeof...(ts)-1]; }
```

kernel/thor/generic/memory-view.cpp
avdgrinten: Yes, but that is not a correctness issue since we're just copying to kernel memory here. This will automatically be fixed when we implement #1214.

kernel/thor/generic/load-balancing.cpp
avdgrinten: Good catch, will fix.

kernel/thor/generic/thor-internal/coroutine.hpp
avdgrinten: Good point, I didn't know that pack indexing is already implemented in Clang. However, it requires us to bump to C++26 which I'll do in a separate PR.

PR #1216, opened 2026-02-04
thor: Refactor memory management to use touchRange() and lockRange()

### LINKED ISSUE #1210: thor: FrontalMemory::fetchRange() assumes that memory is not evicted immediately
If memory is evicted very shortly after initialization, the assertion
```
Assertion failed: pit->loadState == ManagedSpace::kStatePresent
In function fetchRange at ../../../src/managarm/kernel/thor/generic/memory-view.cpp:1207
```
on the exit path of `FrontalMemory::fetchRange()` can fail (e.g., because the page is already in `kStateWantWriteback`).

avdgrinten:
This removes `fetchRange()` (which returned the memory page that was faulted in) in favor of the simpler `touchRange()` (which doesn't return it). Returning the memory page was only ever useful for pinned pages anyway (as non-pinned pages can be evicted immediately).

We also remove `asyncLockRange()` in favor of the synchronous `lockRange()` (which marks a range of memory as pinned without actually faulting in all pages).

Fix #1210.

### DIFF
--- a/kernel/thor/generic/hel.cpp
+++ b/kernel/thor/generic/hel.cpp
@@ -1633,7 +1633,7 @@ HelError doSubmitLockMemoryView(HelHandle handle, smarter::shared_ptr<IpcQueue> 
 			uintptr_t context,
 			enable_detached_coroutine edc) -> void {
 		MemoryViewLockHandle lockHandle{memory, offset, size};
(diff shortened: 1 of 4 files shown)


48cf (approved):

Dennisbonke (approved):
LGTM

PR #1221, opened 2026-02-05
posix: Let chroot() and chdir() fail on non-directories

### LINKED ISSUE #1219: posix: Does not disallow me from cd'ing into files
If I do the following commands, I do not get the expected result of being denied to cd into the file. Ext2 correctly complains about readEntries on a file that doesn't implement it, tmpfs just crashes.
- `cd ~`
- `touch foo`
- `cd foo`

Image

avdgrinten:
Fix #1219.


Dennisbonke (approved):
LGTM

Issue #344, opened 2021-06-07
thor: Fibers should be preemptible


avdgrinten:
This would reduce the latency of user space IRQ handlers (especially if no kernlets are in use).

PR #1227, opened 2026-02-07
thor: Make fibers preemptible

avdgrinten:
This is a small change that implements `KernelFiber::handlePreemption()`.


avdgrinten:
Fixes #344

PR #1228, opened 2026-02-08
Aarch64 smp fixes

Qwinci:
- Add TLB invalidation to the aarch64 smp trampoline
- Factor the max memory region count to `eir/interface.hpp` so it stays in sync between eir/thor
- Fix an incorrect cpu count check inside the smp code where `getCpuCount()` was compared with `cpuConfigNote->totalCpus` (which is returned by `getCpuCount()`)
- Boot the cpus starting with `cpu@` from the children of `cpus` instead of relying on specific compatible properties which may not exist on all boards (eir also uses `cpu@` to count the cpus)
- Don't use the all excluding self gic sgi mode as it may not be properly implemented (Linux uses the normal targeted mode instead)

### DIFF
--- a/kernel/thor/arch/arm/smp.cpp
+++ b/kernel/thor/arch/arm/smp.cpp
@@ -312,10 +312,9 @@ static initgraph::Task initAPs{&globalInitEngine, "arm.init-aps",
 			});
 		}
 
-		if (getCpuCount() != cpuConfigNote->totalCpus)
-			panicLogger() << "thor: Booted " << getCpuCount()
-					<< " CPUs but Eir detected " << cpuConfigNote->totalCpus
-					<< frg::endlog;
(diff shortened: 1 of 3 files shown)


kernel/thor/generic/thor-internal/physical.hpp
```diff
@@ -144,7 +144,7 @@ class PhysicalChunkAllocator {
 		BuddyAccessor buddyAccessor;
 	};
 
-	Region _allRegions[8];
```
qookei, line 147: Maybe this number could be put somewhere in `eir/interface.hpp` maybe, so that we avoid mismatches in the future?

Qwinci: Good point, I moved the eir `numRegions` variable renamed as `eirMaxMemoryRegions` to `eir/interface.hpp` and changed thor to use that.

Issue #1141, opened 2025-11-25
libblockfs: Make traverseLinks() filesystem-independent


avdgrinten:
`traverseLinks()` can be implemented on top of `getLink()`. See also: https://github.com/managarm/managarm/pull/936/files#r2118172343

avdgrinten:
Done in 8e523989c2285eadd04d408571d28890d4e9b3dd

PR #1241, opened 2026-02-12
drivers/libblockfs: add basic read-only btrfs support

no92:


drivers/libblockfs/src/btrfs/btrfs.hpp
```diff
@@ -0,0 +1,140 @@
...
+constexpr bool verboseLogging = false;
+constexpr bool debugTreeWalking = false;
+
+}
+
+#define STUBBED                                                                                    \
+	do {                                                                                           \
+		std::println("libblockfs: {} is stubbed", __FUNCTION__);                                   \
+		assert(!"stubbed");                                                                        \
+	} while (0)
```
avdgrinten: We should avoid crashing the driver for stubbed functionality. This is used in several functions that can return errors. In these cases we should return an error. In other cases, it is probably preferable to log + continue.

drivers/libblockfs/src/btrfs/btrfs.hpp
```diff
@@ -0,0 +1,140 @@
...
+
+struct OpenFile;
+
+struct BtreePtrLayer {
+	LogicalAddress logical;
+	key key;
+	std::vector<std::byte> buffer;
+};
+
+using BtreePtr = std::vector<BtreePtrLayer>;
```
avdgrinten, line 68: What exactly does this represent? Doesn't it make more sense to cache the raw bytes while using some kind of iterator / view class that allows high level access?

drivers/libblockfs/src/btrfs/ops.cpp
```diff
@@ -0,0 +1,326 @@
...
+}
+
+async::result<protocols::fs::Error> chmod(std::shared_ptr<void> object, int mode) {
+	(void)object;
+	(void)mode;
+	STUBBED;
+}
+
+async::result<protocols::fs::TraverseLinksResult>
+traverseLinks(std::shared_ptr<void> object, std::deque<std::string> components) {
```
avdgrinten: Is there anything that stops us from moving this function to generic code (#1141)?

drivers/libblockfs/src/btrfs/btrfs.hpp
no92: This is not really possible in all cases, e.g. createRegular returns a shared_ptr to an Inode.

drivers/libblockfs/src/btrfs/btrfs.hpp
avdgrinten: AFAICT this is still entirely undocumented.

drivers/libblockfs/src/btrfs/spec.hpp
```diff
@@ -0,0 +1,244 @@
...
+#include <generator>
+#include <helix/memory.hpp>
+#include <stdint.h>
+#include <utility>
+
+#include "types.hpp"
+
+namespace blockfs::btrfs {
+
+struct [[gnu::packed]] timespec {
```
avdgrinten: Maybe `BtrfsTimespec`?

drivers/libblockfs/src/btrfs/spec.hpp
```diff
@@ -0,0 +1,244 @@
...
+	int64_t sec;
+	int32_t nsec;
+};
+
+struct [[gnu::packed]] UUID {
```
avdgrinten: `Uuid`.

drivers/libblockfs/src/btrfs/spec.hpp
```diff
@@ -0,0 +1,244 @@
...
+	BLOCK_GROUP_ITEM = 0xC0,
+	FREE_SPACE_INFO = 0xC6,
+	FREE_SPACE_EXTENT = 0xC7,
+	DEV_EXTENT_ITEM = 0xCC,
+	DEV_ITEM = 0xD8,
+	CHUNK_ITEM = 0xE4,
+	DEV_STATS_ITEM = 0xF9,
+};
+
+struct [[gnu::packed]] key {
```
avdgrinten: `Key`.

drivers/libblockfs/src/btrfs/spec.hpp
```diff
@@ -0,0 +1,244 @@
...
+	// Return a key with zero offset.
+	// This can be useful for comparing keys if you only care about objectid and type.
+	key noOffset() const {
+		return {objectid, type};
+	}
+};
+
+static_assert(sizeof(key) == 0x11, "Bad key size");
+
+struct [[gnu::packed]] key_ptr {
```
avdgrinten: `KeyPtr` (and similarly for all other types below).

drivers/libblockfs/src/btrfs/spec.hpp
```diff
@@ -0,0 +1,244 @@
...
+	uint64_t generation;
+	uint64_t owner;
+	uint32_t nritems;
+	uint8_t level;
+
+	static struct block_header *fromMapping(helix::Mapping &mapping) {
+		return reinterpret_cast<block_header *>(mapping.get());
+	}
+
+	std::span<key_ptr> key_ptrs() const {
```
avdgrinten: `keyPtrs()`.

protocols/fs/fs.bragi
```diff
@@ -242,6 +242,8 @@ head(128):
 		tag(87) int64 gid;
 
 		tag(88) uint64 cancellation_id;
+
+		tag(89) string fs_type;
```
avdgrinten: Instead of adding more tags, we should convert the request.

drivers/libblockfs/src/btrfs/btrfs.cpp
```diff
@@ -0,0 +1,706 @@
...
+			    info.size,
+			    info.stripe.device_id,
+			    static_cast<uint64_t>(info.stripe.offset)
+			);
+		}
+	}
+
+	// Allocate a page cache for the tree structures.
+	HEL_CHECK(helCreateManagedMemory(
+	    superblock_.total_bytes, kHelManagedReadahead, &treeBackingMemory, &treeFrontalMemory
```
avdgrinten: We probably don't want readahead here (since we'll mostly be doing random access on this?).

drivers/libblockfs/src/btrfs/btrfs.cpp
```diff
@@ -0,0 +1,706 @@
...
+	stats.numInodes = stats.numBlocks;
+	stats.inodesFree = stats.blocksFree;
+	stats.inodesFreeUser = stats.inodesFree;
+	stats.maxNameLength = BTRFS_NAME_LEN;
+
+	uint32_t fsid[4];
+	memcpy(fsid, &superblock_.fs_uuid, sizeof(superblock_.fs_uuid));
+
+	stats.fsid[0] = (std::byteswap(fsid[0]) ^ std::byteswap(fsid[2]))
+	                ^ 0 /*(this->accessRoot()->number >> 32)*/;
```
avdgrinten: This looks weird.

no92:
Fixed the comments.

avdgrinten (approved):

PR #1242, opened 2026-02-13
XHCI hub support debugging

qookei:


qookei:
Current status:
- both USB 1 and USB 2 hubs work in isolation
- plugging in the USB 2 hub into the USB 1 hub works (the USB 2 hub is essentially downgraded to USB 1, but at least this shows that chaining hubs works in general)
- ~~plugging in the USB 1 hub into the USB 2 hub doesn't work (fails on the Address Device command for the USB 1 hub itself, likely some slot context is wrong)~~ this appears to work, maybe it's just the hubs I have being flaky...

avdgrinten (approved):

PR #1244, opened 2026-02-15
thor: Introduce thread condition logic

avdgrinten:
Right now, `Thread::raiseSignals()` handles various asynchronous actions that a thread can take, namely entering interrupted state if `_pendingSignals` is set, killing the thread if `_pendingKill` is set and CPU migration if we are not on `_lbCb->getAssignedCpu()`. It would be nicer to unify this in a single bitmask.

This PR adds such a `Thread::_pendingConditions` bitmask and converts the transition to the interrupted state to the new bitmask.

### DIFF
--- a/kernel/thor/generic/thread.cpp
+++ b/kernel/thor/generic/thread.cpp
@@ -85,6 +85,10 @@ bool Thread::blockCurrent(bool interruptible) {
 	if(thisThread->_unblockLatch.exchange(false, std::memory_order_acquire))
 		return true;
 
+	// Avoid taking _mutex if we are interrupted anyway.
+	if (interruptible && thisThread->_pendingConditions.load(std::memory_order_relaxed))
+		return false;
+
 	StatelessIrqLock irqLock;
 	auto lock = frg::guard(&thisThread->_mutex);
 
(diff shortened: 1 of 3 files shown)


kernel/thor/generic/thread.cpp
```diff
@@ -293,6 +299,29 @@ void Thread::interruptCurrent(Interrupt interrupt, SyscallImageAccessor image, I
...
+void Thread::handleConditions(SyscallImageAccessor image) {
+	assert(image.iplState()->current < ipl::schedule);
+	auto thisThread = getCurrentThread();
+
+	auto pending = thisThread->_pendingConditions.load(std::memory_order_relaxed);
+	while (pending) {
+		// Dequeue the highest pending bit.
+		auto c = Condition{1} << frg::floor_log2(pending);
+		pending = thisThread->_pendingConditions.fetch_and(~c);
+		if (!(pending & c))
```
48cf: Since bits in _pendingConditions can only be cleared by the thread, this check seems unnecessary - maybe this should be an assert?

avdgrinten: Good catch.

PR #1259, opened 2026-02-20
thor: Map IPC queue memory into kernel virtual memory

avdgrinten:
This gives us a contiguous mapping instead of reading/writing elements in page-sized chunks, yielding a small performance improvement (~7% in the 1-byte exchange messages micro-benchmark).

### DIFF
--- a/kernel/thor/generic/hel.cpp
+++ b/kernel/thor/generic/hel.cpp
@@ -1,3 +1,4 @@
+#include <span>
 #include <string.h>
 #include <cstddef>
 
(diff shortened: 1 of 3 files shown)


kernel/thor/generic/thor-internal/memory-view.hpp
```diff
@@ -477,6 +477,43 @@ struct ImmediateMemory final : MemoryView {
...
+	ImmediateWindow &operator=(ImmediateWindow other) {
+		swap(*this, other);
+		return *this;
+	}
+
+	~ImmediateWindow();
+
+	template<typename T>
+	T *access(uintptr_t offset) {
+		return reinterpret_cast<T *>(reinterpret_cast<std::byte *>(_base) + offset);
```
qookei: Not strictly necessary in this case, but `std::launder` here?

qookei (approved):

PR #1264, opened 2026-02-22
core/virtio: Fix usage of incorrectly sized loadConfig

Qwinci:
This caused e.g. the disk size reported by the virtio block device to incorrectly be zero for a 4gb image.


avdgrinten (approved):

Dennisbonke (approved):
LGTM

PR #1265, opened 2026-02-23
Fixes for more XHCI correctness issues

qookei:


drivers/usb/hcds/xhci/src/ring.cpp
```diff
@@ -295,25 +295,21 @@ void ProducerRing::updateLink() {
...
 
-	// Status stage
-	FRG_CO_TRY(co_await nextEvent_());
+	// Short packets are not allowed for control transfers.
+	// In case of a short packet completion, two events are posted:
+	// one for data stage and one for status stage.
+	if (ev.completionCode == 13) {
+		FRG_CO_TRY(co_await nextEvent_());
+		co_return proto::UsbError::shortPacket;
+	}
```
avdgrinten: The comment seems to be contradictory (?). I think short packets for the data stage are allowed.

drivers/usb/hcds/xhci/src/main.cpp
```diff
@@ -347,6 +347,12 @@ void Controller::processEvent(Event ev) {
 async::result<frg::expected<proto::UsbError, uint32_t>>
 Controller::enableSlot(uint8_t slotType) {
 	auto event = co_await submitCommand(Command::enableSlot(slotType));
+
+	if (event.completionCode != CompletionCode::success)
+		std::cout
+			<< this << "enableSlot(" << (int)slotType << ") failed: "
+			<< event.completionCodeName() << std::endl;
```
avdgrinten: Minor but the prints have no space between `this` and the message. Also, is it intended that we print the `this` address instead of the controller's MMIO address (as we do in other logs)?

Aside from that, we could also (but don't have to) use `std::println()`.

avdgrinten (approved):

drivers/usb/hcds/xhci/src/main.cpp
qookei: `Controller *` and `Controller &` have `operator<<` overloads for `std::ostream` that format to `xhci pci.<address>: ` (incl. the space at the end). As for `std::println` I think I'll add a commit to replace `std::cout` with it.

qookei: Replaced the `std::cout` and `printf` usage with `std::println`.

drivers/usb/hcds/xhci/src/ring.cpp
qookei: Short packets are now allowed on control transfers.

PR #1269, opened 2026-02-24
posix: Implement sigqueue

no92:


posix/subsystem/src/observations.cpp
```diff
@@ -389,7 +391,10 @@ async::result<void> observeThread(std::shared_ptr<Process> self,
 			HEL_CHECK(helStoreRegisters(thread.getHandle(), kHelRegsGeneral, &gprs));
 			UserSignal info;
 			info.pid = self->pid();
-			info.uid = 0;
+			info.uid = self->threadGroup()->uid();
+			info.code = sicode;
+			info.val = sival;
```
avdgrinten, line 406: Managarm should filter the sicode such that mlibc cannot claim to be the kernel.

PR #1274, opened 2026-02-25
thor: Add writebackFence() and invalidateRange()

avdgrinten:
Adds:
- `MemoryView::writebackFence()` which waits for inflight writeback operations to complete. This is exposed in userspace as `helix_ng::writebackFence()`.
- `MemoryView::invalidateRange()` which forcefully invalidates pages of a managed memory space. This is exposed as `helix_ng::invalidateMemory()`.

### DIFF
--- a/testsuites/kernel-tests/meson.build
+++ b/testsuites/kernel-tests/meson.build
@@ -2,8 +2,9 @@ executable('kernel-tests',
 	[
 		'src/main.cpp',
 		'src/faults.cpp',
-		'src/mapping.cpp'
+		'src/mapping.cpp',
+		'src/memory.cpp',
 	],
-	dependencies: [ hel_dep ],
+	dependencies: [ helix_dep ],
 	install : true
 )
--- a/testsuites/kernel-tests/src/memory.cpp
+++ b/testsuites/kernel-tests/src/memory.cpp
@@ -1,0 +1,106 @@
+#include <cassert>
+#include <cstddef>
+
+#include <async/algorithm.hpp>
+#include <async/result.hpp>
+#include <helix/ipc.hpp>
+
+#include "testsuite.hpp"
+
+namespace {
+
+// Processes an exact number of notifications.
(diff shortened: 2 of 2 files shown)


kernel/thor/generic/memory-view.cpp
```diff
@@ -1095,6 +1099,63 @@ Error BackingMemory::updateRange(ManageRequest type, size_t offset, size_t lengt
 	return Error::success;
 }
 
+coroutine<frg::expected<Error>> BackingMemory::writebackFence(uintptr_t offset, size_t size) {
+	assert(!(offset % kPageSize));
+	assert(!(size % kPageSize));
+
```
avdgrinten, line 1133: This should use error returns.

kernel/thor/generic/memory-view.cpp
```diff
@@ -1115,20 +1069,27 @@ Error BackingMemory::updateRange(ManageRequest type, size_t offset, size_t lengt
...
+					ref_rc(pit->monitor.get());
+					pendingMonitors.push_back(pit->monitor.get());
+					pit->monitor = {};
 				}else{
 					assert(pit->loadState == ManagedSpace::kStateAnotherWriteback);
 					pit->loadState = ManagedSpace::kStateWantWriteback;
 					_managed->_writebackList.push_back(&pit->cachePage);
+					ref_rc(pit->monitor.get());
+					pendingMonitors.push_back(pit->monitor.get());
+					pit->monitor = {};
```
48cf: Is there any reason this is done in two steps? (assigning a default-constructed value and then the new value)

48cf (approved):
LGTM but 6ccae8d2dfc8746067999912bbcfa1b2a979e4cd's message should probably be `helix: Add invalidateMemory() wiring`

kernel/thor/generic/memory-view.cpp
avdgrinten: Not really, we can get rid of the assignment to default.

PR #1276, opened 2026-02-26
thor: Handle EC queries outside of interrupt context

### LINKED ISSUE #1248: ec.cpp allocates from an irq which triggers an assertion
The code here https://github.com/managarm/managarm/blob/a2692a7bb0e73ee37969168569f3864afa029925/kernel/thor/system/acpi/ec.cpp#L195 should not allocate since it runs in an irq context. Ideally it should be scheduled to run in a regular fiber context later on

avdgrinten:
Fix #1248

### DIFF
--- a/kernel/thor/system/acpi/ec.cpp
+++ b/kernel/thor/system/acpi/ec.cpp
@@ -1,11 +1,12 @@
+#include <async/sequenced-event.hpp>
+#include <frg/optional.hpp>
 #include <frg/scope_exit.hpp>
 #include <stdlib.h>
 #include <thor-internal/acpi/acpi.hpp>
 #include <thor-internal/arch/ints.hpp>
+#include <thor-internal/coroutine.hpp>
 #include <thor-internal/debug.hpp>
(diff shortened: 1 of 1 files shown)


kernel/thor/system/acpi/ec.cpp
```diff
@@ -313,6 +278,38 @@ static void installEcHandlers() {
...
+				}
+				if (!idx) {
+					infoLogger() << "thor: EC indicates no outstanding events" << frg::endlog;
+					return;
+				}
+
+				static const char *hexChars = "0123456789ABCDEF";
+				char methodName[5] = {'_', 'Q', hexChars[(idx >> 4) & 0xF], hexChars[idx & 0xF]};
+
+				infoLogger() << "thor: evaluating EC query " << methodName << frg::endlog;
```
Dennisbonke: Should we gate this behind a debug logging flag? This feels like it could be really spammy on real hardware.

avdgrinten: Fixed.

PR #1281, opened 2026-03-01
thor, helix: Clean up IpcQueue code

avdgrinten:
After the introduction of SQs, some of the IpcQueue code was left in a somewhat messy state. This PR does not add any major new functionality but it does clean up some loose ends:
- The 1st commit generalizes `waitProgressFutex()` in userspace to be able to wait for multiple notification bits at the same time. Before, this was done in an ad-hoc way to wait for the `kHelUserNotifyProgress | kHelUserNotifyAlert` bits in mlibc.
- The 2nd commit cleans up code paths that wait on new chunks to be appended to use the `kHel{Kernel,User}NotifySupply{Cq,Sq}Chunks` bits as intended (previous code would clear but not re-check them).
- The 3rd commit replaces `kHelDriveWaitCqProgress` by a more general `kHelDriveWait` flag that takes a mask of notifications to ignore (instead of hard-coding ignored notifications in the kernel).
- The 4th commit adds a `kHelProgessFull` bit that is set on chunks that will not be written to again but that are not retired yet (because we're still waiting for their `next` pointers to be set). This can be used by userspace to detect stalls (i.e., situations in which not enough chunks are being supplied).
- The final commit adds missing bounds checking to the kernel side.

### DIFF
--- a/hel/include/hel.h
+++ b/hel/include/hel.h
@@ -479,6 +479,8 @@ struct HelQueueParameters {
 static const int kHelUserNotifyCqProgress = (1 << 0);
 //! Set in userNotify after kernel has supplied new SQ chunks.
 static const int kHelUserNotifySupplySqChunks = (1 << 1);
+//! Set in userNotify if the queue encounters a contract violation.
+static const int kHelUserNotifyError = (1 << 14);
 //! Set in userNotify when the queue is alerted.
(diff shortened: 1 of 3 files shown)


avdgrinten:
Circularly depends on https://github.com/managarm/mlibc/pull/1686 due to the signature change of `helDriveQueue` + the `kHelProgressFull` bit.

PR #1287, opened 2026-03-04
Systemd upgrade items for Managarm

Dennisbonke:
Various fixes all around, depends on managarm/mlibc#1688 for statx

### DIFF
--- a/posix/subsystem/src/process.cpp
+++ b/posix/subsystem/src/process.cpp
@@ -509,9 +509,22 @@ void CompileSignalInfo::operator() (const UserSignal &info) const {
 	si->si_value = info.val;
 }
 
+void CompileSignalFdInfo::operator() (const UserSignal &info) const {
+	si->ssi_code = info.code;
+	si->ssi_pid = info.pid;
+	si->ssi_uid = info.uid;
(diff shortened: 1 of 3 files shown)


posix/subsystem/src/requests/filesystem.cpp
```diff
@@ -1013,6 +1017,8 @@ async::result<void> handleFstatAt(RequestContext& ctx) {
 		resp.set_ctime_nanos(stats.ctimeNanos);
 		resp.set_mount_id(target_mount ? target_mount->mountId() : 0);
 		resp.set_stat_dev(target_link->getTarget()->superblock()->deviceNumber());
+		resp.set_statx_attr(attr);
+		resp.set_statx_attr_mask(STATX_ATTR_MOUNT_ROOT);
```
no92: For maintainability it would be nice to break this out into a `constexpr int statxAttrMask = STATX_ATTR_MOUNT_ROOT;`, and to use that to verify that `attr` only contains bits set in that mask (`assert((attr & ~statxAttrMask) == 0);` or similar)

no92 (approved):

PR #1289, opened 2026-03-09
posix: return EBADF when writing to read-only files

no92:


posix/subsystem/src/tmp_fs.cpp
```diff
@@ -668,6 +668,9 @@ async::result<frg::expected<Error, size_t>>
 MemoryFile::writeAll(Process *, const void *buffer, size_t length) {
 	auto node = static_cast<MemoryNode *>(associatedLink()->getTarget().get());
 
+	if (!(flags_ & semanticWrite))
+		co_return Error::badExecutable;
```
avdgrinten: This looks wrong.

no92: Good catch, fixed.

PR #1298, opened 2026-03-11
posix: supply dirent's d_off and d_ino for ReadEntriesRequest

no92:


protocols/fs/include/protocols/fs/common.hpp
```diff
@@ -149,7 +150,13 @@ inline Error operator|(managarm::fs::Errors e, ToFsProtoError) {
 
 using ReadResult = std::expected<size_t, Error>;
 
-using ReadEntriesResult = std::optional<std::string>;
+struct ReadEntryData {
+	std::string name;
+	ino_t inode;
+	long offset;
+};
```
avdgrinten: Rename this back to `ReadEntriesResult`.

protocols/fs/include/protocols/fs/common.hpp
```diff
@@ -149,7 +150,13 @@ inline Error operator|(managarm::fs::Errors e, ToFsProtoError) {
...
+
+using ReadEntriesResult = std::expected<ReadEntryData, managarm::fs::Errors>;
```
avdgrinten: We should probably just drop the alias.

PR #1299, opened 2026-03-11
posix: rewrite FIFOs to use a ringbuffer

no92:
This aligns with expectations that the buffer space would at some point get exhausted. This also prepares us for Linux's F_SETPIPE_SZ fcntl operation. Using a ringbuffer is correct, as pipes and FIFOs do not have to preserve message boundaries.

### DIFF
--- a/posix/subsystem/src/fifo.cpp
+++ b/posix/subsystem/src/fifo.cpp
@@ -3,17 +3,15 @@
 #include <string.h>
 #include <sys/epoll.h>
 #include <iostream>
-#include <deque>
 #include <map>
-#include <numeric>
 #include <print>
 
 #include <async/recurring-event.hpp>
 #include <bragi/helpers-std.hpp>
+#include <frg/ringbuffer.hpp>
 #include <helix/ipc.hpp>
 #include "fifo.hpp"
 #include "fs.bragi.hpp"
-#include "protocols/fs/common.hpp"
 
 #include <sys/ioctl.h>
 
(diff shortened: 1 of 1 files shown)


posix/subsystem/src/fifo.cpp
```diff
@@ -94,63 +89,61 @@ struct OpenFile : File {
...
+				return !_channel->ring.size();
+			}, ce)))
 				co_return std::unexpected{Error::interrupted};
-			}
-		}
 
-		if(_channel->packetQueue.empty()) {
-			assert(!_channel->writerCount);
-			co_return std::unexpected{Error::eof};
+			chunk = _channel->ring.dequeue({static_cast<uint8_t *>(data), maxLength});
```
avdgrinten: We can avoid two `dequeue()` call sites by turning this into a `while(true)` loop with the `dequeue()` at the top.

posix/subsystem/src/fifo.cpp
```diff
@@ -94,63 +89,61 @@ struct OpenFile : File {
...
+		while (!chunk) {
+			if (nonBlock_)
+				co_return Error::wouldBlock;
+
+			co_await _channel->statusBell.async_wait_if([&]() {
+				return !_channel->ring.available_space();
+			}); // TODO: EINTR
+
+			chunk = _channel->ring.enqueue({static_cast<const uint8_t *>(data), maxLength});
+		}
```
avdgrinten: Same.

PR #1304, opened 2026-03-15
thor: Let memory reclaim be driven by working sets

avdgrinten:
This PR changes how we approach memory reclaiming. Previously, we first decided which pages we want to reclaim and then unmapped them. However, since we want to do anonymous swap in the future, this approach is quite cumbersome as we do not know where a CoW page is mapped (as a single CoW page may be linked in many `CopyOnWriteMemory` views).

This PR changes the approach as follows:
- We scan page table access bits to determine which pages are frequently used. We store the age of a page (i.e., the number of times it was scanned with access bit clear since the last time it was scanned with access bit set) in software-available bits in the PTE.
- When a page's age becomes too high, the page is unmapped.
- Unmapped pages are added to the LRU list in `MemoryReclaimer` and may be reclaimed (using the same mechanism as before).
- There can be a race between a new mapping of a page appearing and reclaim. This is handled by adding a `fenceEphemeral` eviction mode that waits for all temporary references to a page to disappear. Likewise, even in the current code, there was a race between a page being cleaned and writeback. This is handled by adding a new `fenceDirty` eviction mode that waits for shootdown to finish after marking a page as dirty.

### DIFF
--- a/kernel/thor/generic/address-space.cpp
+++ b/kernel/thor/generic/address-space.cpp
@@ -12,6 +12,12 @@ namespace thor {
 
 namespace {
 	constexpr bool logCleanup = false;
+	constexpr bool logRss = false;
+
+	// Used in working set limit computation.
+	// TODO: Do not make this global and add a hierarchical API that allows userspace
+	//       to control weights or similar for working set size computation.
+	constinit std::atomic<size_t> numVirtualSpaces{0};
 
 	uint32_t compilePageFlags(MappingFlags mappingFlags) {
 		uint32_t pageFlags = 0;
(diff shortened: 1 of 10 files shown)


kernel/thor/generic/thor-internal/address-space.hpp
```diff
@@ -501,7 +514,7 @@ struct VirtualSpace {
 	retrievePhysical(VirtualAddr address);
 
 	size_t rss() {
-		return _ops->getRss();
+		return rss_.load(std::memory_order_relaxed);
```
48cf: This should probably be clamped to 0 if it ever becomes negative.

kernel/thor/generic/address-space.cpp
```diff
@@ -268,6 +272,65 @@ void VirtualSpace::retire() {
...
+			}
+			if(!mapping)
+				break;
+
+			nextAddress = mapping->address + mapping->length;
+
+			if(mapping->state.load(std::memory_order_relaxed) != MappingState::active)
+				continue;
+
+			co_await mapping->exposeRcu.barrier();
```
avdgrinten: This is unnecessary.

PR #1322, opened 2026-04-04
Introduce request dispatcher infrastructure

avdgrinten:
The first commit introduces the new `dispatchRequest()` and `dispatchTail()` functions to help with Bragi message dispatching. The remaining commits convert some of the existing server code to use `dispatchRequest()`.

A non-trivial use case of `dispatchTail()` can be found in kernletcc code. A non-trivial handler (that also does Bragi logging) can be found in the DRM code.

### DIFF
--- a/core/include/core/dispatch.hpp
+++ b/core/include/core/dispatch.hpp
@@ -1,0 +1,162 @@
+#pragma once
+
+#include <expected>
+#include <vector>
+
+#include <async/result.hpp>
+#include <bragi/helpers-all.hpp>
+#include <bragi/helpers-std.hpp>
+#include <helix/ipc.hpp>
+
+enum class DispatchError {
+	none, ///< No error
+	shutdown, ///< The IPC lane was shut down (e.g. due to a client exit)
+	ipcError, ///< An IPC error occured
+	malformedMessage, ///< The received message could not be decoded
+};
+
+// Helper function for dispatchRequests().
+template <typename Message, typename Visitor, typename ...Args>
+async::result<DispatchError>
+doDispatch(helix_ng::RecvInlineResult &recvHead, helix::BorrowedDescriptor conversation, bragi::preamble preamble, Visitor &&visitor, Args &&...args) {
+	// Eagerly reset() after parsing to free up space in IPC queue.
+	auto maybeMessage = bragi::parse_head_only<Message>(recvHead);
+	recvHead.reset();
+	if (!maybeMessage)
+		co_return DispatchError::malformedMessage;
+
+	auto res = co_await visitor(
+		std::move(*maybeMessage), conversation, preamble,
+		std::forward<Args>(args)...);
+	if (!res)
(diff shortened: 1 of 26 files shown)


core/include/core/dispatch.hpp
```diff
@@ -0,0 +1,162 @@
...
+
+	if (preamble.error())
+		co_return std::unexpected(DispatchError::malformedMessage);
+
+	DispatchError err = DispatchError::none;
+
+	bool found = (
+		(
+			preamble.id() == bragi::message_id<Messages>
+			? (err = co_await doDispatch<Messages>(recvHead, std::move(conversation), preamble,
```
no92, line 86: Doesn't this turn into a use-after-move of `conversation` if there are duplicate `Messages`? Is the uniqueness of `Messages` enforced by the compiler already, or should we do that?

core/include/core/dispatch.hpp
```diff
@@ -0,0 +1,162 @@
...
+#include <expected>
+#include <vector>
+
+#include <async/result.hpp>
+#include <bragi/helpers-all.hpp>
+#include <bragi/helpers-std.hpp>
+#include <helix/ipc.hpp>
+
+enum class DispatchError {
+	none, ///< No error
```
no92, line 12: Isn't this unnecessary if DispatchError is always used in a `std::expected`?

core/include/core/dispatch.hpp
avdgrinten: I think Bragi enforces distinct IDs (?).

core/include/core/dispatch.hpp
```diff
@@ -0,0 +1,162 @@
...
+	shutdown, ///< The IPC lane was shut down (e.g. due to a client exit)
+	ipcError, ///< An IPC error occured
+	malformedMessage, ///< The received message could not be decoded
+};
+
+// Helper function for dispatchRequests().
+template <typename Message, typename Conversation, typename Visitor, typename ...Args>
+async::result<DispatchError>
+doDispatch(helix_ng::RecvInlineResult &recvHead, Conversation &&conversation, bragi::preamble preamble, Visitor &&visitor, Args &&...args) {
```
avdgrinten: We should actually change the `conversion` to `BorrowedLane` here such that we can still `dismiss` if we do not understand the request.

core/include/core/dispatch.hpp
avdgrinten: Discussed offline: passing the same message type is not harmful as either both will not match or `||` will short-circuit. Testing for duplicate message types would probably be expensive.

core/include/core/dispatch.hpp
avdgrinten: We do return this type from `doDispatch()`. Also, it probably makes sense to have a "no error" variant.

no92 (approved):

PR #1328, opened 2026-04-13
eir, thor: Add command line option to limit CPU count at boot

marv7000:


avdgrinten:
CI fails because the DTB code in Eir does not set `totalCpus`.

avdgrinten:
Also, RISC-V and Aarch64 AP booting code needs to be adjusted in the same way as x86 AP booting code.

avdgrinten (approved):

PR #1336, opened 2026-04-30
Several DRM fixes and improvements for KDE

Dennisbonke:


core/drm/src/core.cpp
```diff
@@ -124,8 +124,8 @@ drm_core::File::importBufferObject(helix_ng::Credentials creds) {
 void drm_core::File::postEvent(drm_core::Event event) {
 	HEL_CHECK(helGetClock(&event.timestamp));
 
+	++_eventSequence;
```
no92: This change should be dropped.

core/drm/src/ioctl.cpp
```diff
@@ -1209,7 +1220,9 @@ struct drm_core::File::HandleIoctl {
 			}
 
 			if(req.drm_flags() & DRM_MODE_PAGE_FLIP_EVENT) {
-				self->pageFlipEvent(std::move(config), self, req.drm_cookie(), std::move(crtc_ids));
+				if (logDrmRequests)
+					std::println("\tSending page flip event...");
+				self->pageFlipEvent(std::move(config), self, req.drm_cookie(), std::vector<uint32_t>{crtc_ids.begin(), crtc_ids.end()});
```
no92: It might be nicer to change the `pageFlipEvent` function to take a `std::ranges::input_range` or `std::ranges::forward_range` instead of a vector.

core/drm/src/core.cpp
```diff
@@ -248,6 +248,17 @@ drm_core::PrimeFile::seekEof(void *object, int64_t offset) {
 	co_return static_cast<ssize_t>(self->offset);
 }
 
+async::result<frg::expected<protocols::fs::Error, protocols::fs::PollStatusResult>> drm_core::PrimeFile::pollStatus(void *object) {
+	co_return protocols::fs::PollStatusResult{1, EPOLLIN | EPOLLOUT};
```
no92, line 254: We should leave a TODO for when we implement fence fds.

core/drm/src/core.cpp
```diff
@@ -248,6 +248,17 @@ drm_core::PrimeFile::seekEof(void *object, int64_t offset) {
...
+}
+
+async::result<frg::expected<protocols::fs::Error, protocols::fs::PollWaitResult>> drm_core::PrimeFile::pollWait(void *object, uint64_t sequence, int mask, async::cancellation_token cancellation) {
+	if(sequence == 0)
```
no92: We should leave a TODO for when we implement fence fds.

core/drm/src/ioctl.cpp
avdgrinten: That would require the function to be moved to a header though which doesn't make a lot of sense.

no92:
Drafted because weston hangs with these changes.

core/drm/src/core.cpp
```diff
@@ -248,6 +248,23 @@ drm_core::PrimeFile::seekEof(void *object, int64_t offset) {
...
+}
+
+async::result<frg::expected<protocols::fs::Error, protocols::fs::PollWaitResult>>
+drm_core::PrimeFile::pollWait(
+    void *, uint64_t sequence, int, async::cancellation_token cancellation
+) {
+	// TODO: properly support this once we implement fences
+	if (sequence == 0)
+		co_return protocols::fs::PollWaitResult{1, EPOLLIN | EPOLLOUT};
+	async::suspend_indefinitely(cancellation);
```
avdgrinten: Missing `co_await`.

core/drm/include/core/drm/core.hpp
```diff
@@ -103,7 +104,7 @@ struct File {
 	async::detached pageFlipEvent(std::unique_ptr<drm_core::Configuration> config,
 		drm_core::File *self, uint64_t cookie, uint32_t crtc_id);
 	async::detached pageFlipEvent(std::unique_ptr<drm_core::Configuration> config,
-		drm_core::File *self, uint64_t cookie, std::vector<uint32_t> crtc_id);
+		drm_core::File *self, uint64_t cookie, std::ranges::input_range auto const &crtc_id);
```
avdgrinten: This is making the function generic for no good reason.

PR #1337, opened 2026-05-01
Implement SIGSTOP+SIGCONT

no92:


no92:
`pause()` makes the observation loop wait for a signal, but the stopping of the threads is currently done inside of the interrupt observation, which is never reached.

posix/subsystem/src/observations.cpp
```diff
@@ -695,6 +695,35 @@ async::result<void> observeThread(std::shared_ptr<Process> self,
...
+
+				// Wait for all other threads to signal their stop.
+				co_await self->threadGroup()->threadStoppedEvent.async_wait_if([&] {
+					return !self->threadGroup()->allStopped();
+				});
+				// Send waitpid/SIGCHLD notifications.
+				self->threadGroup()->doSignalStop();
+
+				// Await SIGCONT
+				co_await self->threadGroup()->processSigcontEvent.async_wait();
```
avdgrinten: We should use `async_wait_if()` (and loop).

posix/subsystem/src/observations.cpp
```diff
@@ -695,6 +695,35 @@ async::result<void> observeThread(std::shared_ptr<Process> self,
...
+				co_await self->threadGroup()->threadContinuedEvent.async_wait_if([&] {
+					return !self->threadGroup()->allContinued();
+				});
+				self->threadGroup()->doSignalContinue();
+			} else if (self->stoppageState == Process::StoppageState::stopRequested) {
+				self->stoppageState = Process::StoppageState::stopped;
+
+				self->threadGroup()->threadStoppedEvent.raise();
+				// Await SIGCONT
+				co_await self->threadGroup()->processSigcontEvent.async_wait();
```
avdgrinten: Likewise: `async_wait_if()`.

posix/subsystem/src/process.cpp
```diff
@@ -622,11 +622,58 @@ void SignalQueue::issueSignal(int sn, SignalInfo info, uint64_t seq) {
...
+				if (auto item = proc->tryFetchSignal(1 << (SIGCONT - 1)); item)
+					delete item;
+				else
+					break;
+			}
+		}
+
+		auto leaderThread = threads_.front();
+		leaderThread->stoppageState = Process::StoppageState::initiateStop;
+		HEL_CHECK(helInterruptThread(leaderThread->threadDescriptor().getHandle()));
```
avdgrinten, line 657: Why do we need the `initiateStop` state? Can't we put the logic that waits for all threads to enter `stopped` state here?

posix/subsystem/src/process.cpp
```diff
@@ -622,11 +622,58 @@ void SignalQueue::issueSignal(int sn, SignalInfo info, uint64_t seq) {
...
+
+bool isStopSignal(int signal) {
+	return std::ranges::contains(stopSignals, signal);
+}
+
+} // namespace
+
 void ThreadGroup::issueThreadGroupSignal(int sn, SignalInfo info) {
-	signalQueue_.issueSignal(sn, info, ++currentSignalSeq_);
+	if (isStopSignal(sn)) {
```
avdgrinten, line 636: AFAICT `issueThreadGroupSignal()` is the wrong place to handle stopping: while `SIGSTOP` is never queued, `SIGTSTP`, `SIGTTIN` and `SIGTTOU` are queued. Also, whether the process is stopped or not depends on the signal action. Only the default action stops the process (?).

posix/subsystem/src/process.cpp
```diff
@@ -1795,9 +1856,61 @@ ThreadGroup *ThreadGroup::create(std::shared_ptr<PidHull> hull, ThreadGroup *par
 	return tg.get();
 }
 
+void ThreadGroup::doSignalStop() {
+	_state = StoppedBySignal{SIGSTOP};
+	getParent()->_notifyQueue.push_back(this);
+	getParent()->_notifyBell.raise();
+	notifyType_ = NotifyType::stopped;
+	notifyTypeChange_.raise();
```
avdgrinten, line 1878: We probably also need to discard old `stopped` notifications.

posix/subsystem/src/process.cpp
```diff
@@ -1795,9 +1856,61 @@ ThreadGroup *ThreadGroup::create(std::shared_ptr<PidHull> hull, ThreadGroup *par
...
+		});
+	}
+}
+
+void ThreadGroup::doSignalContinue() {
+	_state = ContinuedBySignal{};
+	getParent()->_notifyQueue.push_back(this);
+	getParent()->_notifyBell.raise();
+	notifyType_ = NotifyType::continued;
+	notifyTypeChange_.raise();
```
avdgrinten, line 1897: We probably also need to discard old `stopped` notifications.

no92:
Depends on managarm/mlibc#1774.

posix/subsystem/src/process.cpp
no92: Awaiting for all threads of the process to stop should probably be done from the observation loop (where it would block the affected process) instead of here, where it could block the issuing process.

Dennisbonke (approved):
LGTM

posix/subsystem/src/observations.cpp
```diff
@@ -562,18 +571,34 @@ async::result<void> observeThread(std::shared_ptr<Process> self,
...
 
-				if (!std::get<1>(check))
-					co_await self->pollSignal(std::get<0>(check), ~self->signalMask());
+				if (!std::get<1>(check)) {
+					co_await async::race_and_cancel(
+						async::lambda([&] (auto c) -> async::result<void> {
+							co_await self->pollSignal(std::get<0>(check), ~self->signalMask(), c);
+						}),
+						async::lambda([tg = self->threadGroup()] (auto c) -> async::result<void> {
+							co_await tg->interruptObservationEvent.async_wait(c);
```
avdgrinten, line 588: `async_wait()` is prone to races and should not be used in new code.

posix/subsystem/src/process.cpp
```diff
@@ -1795,9 +1869,63 @@ ThreadGroup *ThreadGroup::create(std::shared_ptr<PidHull> hull, ThreadGroup *par
...
+	    [](auto state) { return state == Process::StoppageState::none; },
+	    &Process::stoppageState
+	);
+}
+
 async::result<void> ThreadGroup::terminateGroup(TerminationState state) {
 	// Only the first terminate() call goes through.
-	if (!std::holds_alternative<std::monostate>(_state))
+	if (std::holds_alternative<TerminationByExit>(_state) && std::holds_alternative<TerminationBySignal>(_state))
 		co_return;
```
avdgrinten, line 1929: Should this be `||`? The condition is always false.

posix/subsystem/src/process.cpp
```diff
@@ -1742,11 +1813,14 @@ Process::wait(int pid, WaitFlags flags, async::cancellation_token ct) {
...
 			};
 
-			if(!(flags & waitLeaveZombie)) {
+			if (!(flags & waitLeaveZombie)) {
 				// erasing from the queue invalidates the iterator, so we take a pointer before
 				auto tg = *it;
 				threadGroup()->_notifyQueue.erase(it);
-				ThreadGroup::retire(tg);
+				if (std::holds_alternative<TerminationByExit>((*it)->_state)
+				    || std::holds_alternative<TerminationBySignal>((*it)->_state)) {
```
avdgrinten, line 1821: `it` -> `tg`.

posix/subsystem/src/process.cpp
```diff
@@ -622,12 +622,72 @@ void SignalQueue::issueSignal(int sn, SignalInfo info, uint64_t seq) {
...
+			thread->stoppageState = Process::StoppageState::stopRequested;
+			HEL_CHECK(helInterruptThread(thread->threadDescriptor().getHandle()));
+		}
+
+		interruptObservationEvent.raise();
+	} else if (sn == SIGCONT) {
+		for (auto thread : threads_ | std::views::drop(1)) {
+			thread->stoppageState = Process::StoppageState::continueRequested;
+			HEL_CHECK(helInterruptThread(thread->threadDescriptor().getHandle()));
+		}
```
avdgrinten, line 669: This looks problematic to me. We `drop(1)` but it's not clear that the first process in the list did not change. Also, we overwrite the `stoppageState` even though it may not be `sopped` yet.

avdgrinten:
One fix for the pause bug is to just let `SIGSTOP` be queued like all other signals. Then we can deal with issue that `SIGSTOP` / `SIGKILL` need to act immediately in a separate PR later on. The current logic in `pause()` seems to be non-trivial to fix to me.

PR #1340, opened 2026-05-01
Various KDE improvements

Dennisbonke:
This PR implements various missing items for KDE and improves our VT emulation. Should not depend on any mlibc PR.

### DIFF
--- a/drivers/libevbackend/src/libevbackend.cpp
+++ b/drivers/libevbackend/src/libevbackend.cpp
@@ -366,12 +366,34 @@ File::ioctl(void *object, uint32_t, helix_ng::RecvInlineResult msg,
 	}
 }
 
+async::result<int> File::getFileFlags(void *object) {
+	auto self = static_cast<File *>(object);
+	int flags = O_RDWR;
+	if(self->_nonBlock)
+		flags |= O_NONBLOCK;
+	co_return flags;
+}
+
+async::result<void> File::setFileFlags(void *object, int flags) {
+	auto self = static_cast<File *>(object);
+	if (flags & ~O_NONBLOCK) {
+		std::cout << "libevbackend: setFileFlags on evdev event file called with unknown flags" << std::endl;
+	}
+	if (flags & O_NONBLOCK)
+		self->_nonBlock = true;
+	else
+		self->_nonBlock = false;
+	co_return;
+}
+
 
 constexpr auto fileOperations = protocols::fs::FileOperations{
 	.read = &File::read,
 	.ioctl = &File::ioctl,
 	.pollWait = &File::pollWait,
-	.pollStatus = &File::pollStatus
+	.pollStatus = &File::pollStatus,
+	.getFileFlags = &File::getFileFlags,
+	.setFileFlags = &File::setFileFlags
 };
 
 helix::UniqueLane File::serve(smarter::shared_ptr<File> file) {
--- a/posix/subsystem/src/devices/ttyn.cpp
+++ b/posix/subsystem/src/devices/ttyn.cpp
@@ -4,10 +4,23 @@
 
 #include "../common.hpp"
 #include "ttyn.hpp"
+#include "../process.hpp"
 #include <bragi/helpers-std.hpp>
 #include <core/dispatch.hpp>
+#include <core/tty.hpp>
+
+#include <linux/vt.h>
+
+#include <termios.h>
 
 #include <bitset>
+
+int activeVt = 1;
+int nextVt = 2;
+
+namespace {
+	constexpr bool logVTRequests = false;
+}
 
 namespace {
 
@@ -48,7 +61,7 @@ private:
 	struct HandleIoctl {
 		async::result<std::expected<void, DispatchError>> operator()(
 				managarm::fs::GenericIoctlRequest &&req, helix::BorrowedDescriptor conversation,
-				bragi::preamble, TTYNFile *) {
+				bragi::preamble, TTYNFile *self) {
 			if(req.command() == TIOCSCTTY) {
 				auto [extractCreds] = co_await helix_ng::exchangeMsgs(
 					conversation,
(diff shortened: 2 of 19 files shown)


drivers/libevbackend/src/libevbackend.cpp
```diff
@@ -366,12 +366,35 @@ File::ioctl(void *object, uint32_t, helix_ng::RecvInlineResult msg,
...
+		flags |= O_NONBLOCK;
+	co_return flags;
+}
+
+async::result<void> File::setFileFlags(void *object, int flags) {
+	auto self = static_cast<File *>(object);
+	if (flags & ~O_NONBLOCK) {
+		std::cout << "posix: setFileFlags on evdev event file called with unknown flags" << std::endl;
+		co_return;
+	}
```
avdgrinten, line 381: Since we can't error out here, don't we want to drop the `co_return` such that the `O_NONBLOCK` handling below still runs? Also in the message `posix` -> `libevbackend`.

posix/subsystem/src/process.cpp
```diff
@@ -610,8 +610,11 @@ SignalHandler SignalContext::changeHandler(int sn, SignalHandler handler) {
 }
 
 void SignalQueue::issueSignal(int sn, SignalInfo info, uint64_t seq) {
-	assert(sn > 0);
+	assert(sn >= 0);
 	assert(sn - 1 < 64);
+	// Signal number 0 means only do the permission checks and see if the process exists, but don't actually raise a signal.
+	if(sn == 0)
+		return;
```
avdgrinten: Requesting review of this part from @no92

protocols/fs/fs.bragi
```diff
@@ -531,6 +531,13 @@ head(128):
 		tag(42) int64 pgid;
 
 		tag(46) uint64 ticks;
+
+		// VT related items
+		tag(47) uint8 vt_mode;
+		tag(48) uint8 vt_waitv;
+		tag(49) uint16 vt_relsig;
+		tag(50) uint16 vt_acqsig;
+		tag(51) uint16 vt_frsig;
```
avdgrinten: We should not add any more tags.

posix/subsystem/src/tmp_fs.cpp
```diff
@@ -884,6 +884,8 @@ DirectoryNode::mkdir(Process *proc, std::string name, mode_t mode) {
 
 	auto node = std::make_shared<DirectoryNode>(static_cast<Superblock *>(superblock()), mode & ~umask, uid, gid);
 	auto the_node = node.get();
+	co_await the_node->chmod(mode);
+	co_await the_node->chown(uid, gid);
```
avdgrinten: This should not be needed anymore as the uid/gid/mode is passed to the constructor call above (?).

posix/subsystem/src/un-socket.cpp
```diff
@@ -591,6 +591,8 @@ struct OpenFile : File {
 			}
 			assert(nodeResult);
 			auto node = nodeResult.value();
+			co_await node->getTarget()->chmod(0666);
+			co_await node->getTarget()->chown(process->threadGroup()->uid(), process->threadGroup()->gid());
```
avdgrinten: We should pass this to `mksocket()` instead.

posix/subsystem/src/un-socket.cpp
```diff
@@ -646,11 +646,11 @@ struct OpenFile : File {
 			server->_statusBell.raise();
 
 			co_await async::race_and_cancel(
-				[&](async::cancellation_token c) { return raceSendTimeout(c); },
-				[&](async::cancellation_token c) -> async::result<void> {
+				async::lambda([&](async::cancellation_token c) { return raceSendTimeout(c); }),
```
avdgrinten, line 647: This doesn't need `async::lambda` as it's not a coroutine (the case below does need it though).

posix/subsystem/src/devices/ttyn.cpp
```diff
@@ -65,8 +78,177 @@ struct TTYNFile final : File {
 					helix_ng::sendBuffer(ser.data(), ser.size())
 				);
 				HEL_CHECK(sendResp.error());
+			}else if(req.command() == VT_OPENQRY) {
```
avdgrinten, line 79: I did not review the ioctls. CC @no92

posix/subsystem/src/process.cpp
```diff
@@ -610,8 +610,11 @@ SignalHandler SignalContext::changeHandler(int sn, SignalHandler handler) {
 }
 
 void SignalQueue::issueSignal(int sn, SignalInfo info, uint64_t seq) {
-	assert(sn > 0);
```
no92, line 613: Removing this assert is incorrect, the permission checking behavior of `sn == 0` should be handled in the kill(2) handler instead.

posix/subsystem/src/devices/ttyn.cpp
```diff
@@ -65,18 +78,195 @@ struct TTYNFile final : File {
...
+				resp.set_vt_mode(self->_mode);
+				resp.set_vt_waitv(self->_waitv);
+				resp.set_vt_relsig(self->_relsig);
+				resp.set_vt_acqsig(self->_acqsig);
+				resp.set_vt_frsig(self->_frsig);
+
+				auto ser = resp.SerializeAsString();
+				auto [sendResp] = co_await helix_ng::exchangeMsgs(
+					conversation,
+					helix_ng::sendBuffer(ser.data(), ser.size())
```
no92: We should use the `helix_ng::sendBragi*` helpers instead.

protocols/fs/fs.bragi
```diff
@@ -641,6 +641,15 @@ head(128):
 		tag(73) uint64 size;
 
 		tag(74) uint32 status;
+
+		// VT related items
```
no92: It would be better to have symmetrical `Vt*Response` messages instead of adding members here.

posix/subsystem/src/devices/ttyn.cpp
```diff
@@ -59,24 +72,187 @@ struct TTYNFile final : File {
...
+				HEL_CHECK(send_resp.error());
+			}else if(req.command() == VT_OPENQRY) {
+				auto [extractCreds] = co_await helix_ng::exchangeMsgs(
+					conversation,
+					helix_ng::extractCredentials()
+				);
+				HEL_CHECK(extractCreds.error());
+
+				if(nextVt == MAX_NR_CONSOLES)
+					nextVt = MAX_NR_CONSOLES; // TODO: Return an error here instead.
```
no92: Linux sets the value to `-1` if there's no availabe VT.

posix/subsystem/src/devices/ttyn.cpp
```diff
@@ -59,24 +72,187 @@ struct TTYNFile final : File {
...
+					req.vt_mode(), req.vt_waitv(), req.vt_relsig(), req.vt_acqsig(), req.vt_frsig());
+			}
+
+			self->_mode = req.vt_mode();
+			self->_waitv = req.vt_waitv();
+			self->_relsig = req.vt_relsig();
+			self->_acqsig = req.vt_acqsig();
+			self->_frsig = req.vt_frsig();
+			
+			managarm::fs::GenericIoctlReply resp;
```
no92: `VtSetModeRequest` might want to use a `VtSetModeResponse` instead, but this is not essential for now. Same for the other ioctls.

Dennisbonke:
Addressed all comments raised.

no92 (approved):

PR #1346, opened 2026-05-04
thor: Refactor overflow builtins, fix a few missing checks

avdgrinten:


kernel/thor/generic/address-space.cpp
```diff
@@ -387,7 +387,10 @@ VirtualSpace::map(smarter::borrowed_ptr<MemorySlice> slice,
 	assert(length);
 	assert(!(length % kPageSize));
 
-	if(offset + length > slice->length())
```
no92, line 390: This is missing the include of `#include <frg/safe_int.hpp>`.

PR #1360, opened 2026-05-09
thor: Refactor Universe methods to perform locking internally

avdgrinten:
We will need to change the locking for OOM safety changes in the future, so it makes sense to refactor the API (which exposed the locking externally) first.


no92 (approved):
Hard to review but looks reasonable to me.

PR #1363, opened 2026-05-11
Initial ext4 support

Qwinci:
Add initial ext4 support to libblockfs. Right now it only supports read/write without journaling/hash tree directories/other stuff like that which makes it unsuitable for actual use right now if the filesystem needs to be accessed from other OS's.

### DIFF
--- a/drivers/libblockfs/src/ext2/ext2fs.cpp
+++ b/drivers/libblockfs/src/ext2/ext2fs.cpp
@@ -17,8 +17,11 @@
 #include <helix/memory.hpp>
 
 #include <array>
+#include <bit>
 
 #include "ext2fs.hpp"
+#include "extents.hpp"
+#include "../checksums.hpp"
 
 namespace blockfs {
 namespace ext2fs {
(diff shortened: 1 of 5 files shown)


drivers/libblockfs/src/libblockfs.cpp
```diff
@@ -448,13 +448,13 @@ async::detached serveDevice(helix::UniqueLane lane, std::unique_ptr<raw::RawFs>
 
 async::detached runDevice(BlockDevice *device) {
 	if (!tracingInitialized) {
-		co_await ostContext.create();
 		tracingInitialized = true;
+		co_await ostContext.create();
 	}
```
avdgrinten: This is a bit counterintuitive. Maybe we should just use an `async::mutex` around this block instead of swapping the order.

drivers/libblockfs/src/ext2/ext2fs.cpp
```diff
@@ -28,14 +30,287 @@ namespace {
...
+
+			uint32_t value = crc32.finalize();
+			desc->inodeBitmapCsumLow = value & 0xffff;
+
+			if(fs.bgdt.descriptorSize() >= offsetof(DiskGroupDesc, inodeBitmapCsumHigh) + 2)
+				desc->inodeBitmapCsumHigh = value >> 16;
+		}
+	}
+
+	struct Md4Context {
```
avdgrinten: This appears to be unused?

drivers/libblockfs/src/ext2/ext2fs.cpp
```diff
@@ -28,14 +30,287 @@ namespace {
...
+			crc16 = (crc16 >> 1) ^ (crc16 & 1 ? 0xA001 : 0);
+
+			for(uint32_t j = 0; j < 256; j += 2 * i)
+				arr[i + j] = crc16 ^ arr[j];
+		}
+
+		return arr;
+	}();
+
+	struct Crc32c {
```
avdgrinten: We should move the checksum algorithms to their own header. If we need them elsewhere (and not just in libblockfs), we could also add them to frigg.

drivers/libblockfs/src/ext2/ext2fs.hpp
```diff
@@ -145,6 +295,61 @@ enum {
 	EXT2_S_IFDIR = 0x4000
 };
 
+enum {
+	EXT4_INDEX_FL = 0x1000,
+	EXT4_EXTENTS_FL = 0x80000,
+	EXT4_INLINE_DATA_FL = 0x10000000
+};
+
+struct HashTreeEntry {
```
avdgrinten: Since the hash tree handling is not added yet, should these structs be broken out to a separate commit?

drivers/libblockfs/src/ext2/extents.hpp
```diff
@@ -0,0 +1,137 @@
...
+	stop
+};
+
+struct ExtentWalker {
+	ExtentWalker(FileSystem *fs, Inode *inode, bool lookup) : fs{fs}, inode{inode}, lookup{lookup} { }
+
+	static constexpr uint16_t maxExtentDepthAfterInode = 4;
+
+	template<typename Enter, typename Leave>
+	async::result<bool> walk(uint64_t blockIndex, Enter enter, Leave leave) {
```
avdgrinten, line 37: It would be good to add a comment to document what `blockIndex` is and when enter/leave are called.

avdgrinten (commented):
Some high level comments but I didn't not check the extent logic in detail yet.

drivers/libblockfs/src/ext2/extents.hpp
```diff
@@ -0,0 +1,137 @@
...
+
+				prevIndex = i;
+			}
+
+			if(lookup) {
+				if(indices[prevIndex].block > blockIndex)
+					co_return false;
+			}
+
+			assert(pathCount < 4);
```
avdgrinten: Is this guaranteed by ext4?

drivers/libblockfs/src/ext2/extents.hpp
```diff
@@ -0,0 +1,137 @@
...
+			auto ptr = blockBuffer.data() + (pathCount - 1) * bufferBytesPerBlock;
+			co_await fs->device->readSectors(nextBlock * fs->sectorsPerBlock,
+					ptr,
+					fs->sectorsPerBlock);
+			currentBlock = nextBlock;
+			hdr = reinterpret_cast<ExtentHeader *>(ptr);
+		}
+
+		assert(hdr->magic == EXT4_EXTENT_MAGIC);
+		// assert(hdr->entries);
```
avdgrinten: Should this be asserted or not?

Qwinci:
The comments should be addressed now.

PR #1368, opened 2026-05-13
thor: Move allocations out of IRQ disabled regions and spinlocks

avdgrinten:
This is a first step towards avoiding panics on OOM. We move allocations out of code paths that have IRQs disabled such that we can retry critical allocation in the future. For cases where it's impossible to move allocations out of locks, we introduce an `AdaptiveMutex` class that is similar to a spinlock but that allows blocking while the lock is taken (which will cause waiters on the lock to also block).

Circularly dependent on https://github.com/managarm/libasync/pull/44

### DIFF
--- a/kernel/thor/arch/x86/paging.cpp
+++ b/kernel/thor/arch/x86/paging.cpp
@@ -103,14 +103,16 @@ KernelPageSpace &KernelPageSpace::global() {
 static constexpr int maxPcidCount = 8;
 
 void initializeAsidContext(CpuData *cpuData) {
-	auto irqLock = frg::guard(&irqMutex());
-
 	// If PCIDs are not supported, create only one binding.
 	auto pcidCount = getCpuData()->havePcids ? maxPcidCount : 1;
 
 	asidData.get(cpuData).initialize(pcidCount);
 	asidData.get(cpuData)->globalBinding.initialize(globalBindingId);
-	asidData.get(cpuData)->globalBinding.initialBind(*kernelSpacePtr);
+
+	{
+		auto irqLock = frg::guard(&irqMutex());
+		asidData.get(cpuData)->globalBinding.initialBind(*kernelSpacePtr);
+	}
 }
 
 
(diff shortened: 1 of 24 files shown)


heatd:
Since I stumbled upon this, some random comments:

> This is a first step towards avoiding panics on OOM. We move allocations out of code paths that have IRQs disabled such that we can retry critical allocation in the future.

It's really /alright/ to have a limited amount of allocations under locks. I would recommend you have some amount of  reserved memory for these. Linux calls it the "min watermark". Other systems have similar IIRC. Note that you very usually end up having codepaths where only a limited form of reclaim can be attempted (say, while you're holding MM or FS-related locks), even if you're able to block.

> For cases where it's impossible to move allocations out of locks, we introduce an `AdaptiveMutex` class that is similar to a spinlock but that allows blocking while the lock is taken (which will cause waiters on the lock to also block).

Note that spinning on any mutex is pretty imperative for performance. Sleeping on a mutex is not good. This Is Known. So you want the adaptive mutex to be the default (but only spin as long as you don't need to be scheduled out, and while the lock holder is running). So I would not recommend you introduce a separate construct for this.

Hope this gives you some ideas :)

avdgrinten:
> Note that spinning on any mutex is pretty imperative for performance. Sleeping on a mutex is not good. This Is Known. So you want the adaptive mutex to be the default (but only spin as long as you don't need to be scheduled out, and while the lock holder is running). So I would not recommend you introduce a separate construct for this.

That's a good point. Making the new mutex implementation (with some adjustments) the default might be a good idea.

kernel/thor/arch/x86/paging.cpp
```diff
@@ -103,14 +103,16 @@ KernelPageSpace &KernelPageSpace::global() {
 static constexpr int maxPcidCount = 8;
 
 void initializeAsidContext(CpuData *cpuData) {
-	auto irqLock = frg::guard(&irqMutex());
```
qookei, line 106: Same change needed on aarch64 and riscv64?

qookei (commented):
I think the changes look good in general, just some nits.

The commit message for `thor: Move allocation out of IRQ disabled region in x86 IRQ slot allocation` seems inconsistent with the changes? The allocation itself is still protected by an `IrqSpinlock` but the part after allocating isn't.

PR #1369, opened 2026-05-15
virtio: Fix virtio-net modern transport

### LINKED ISSUE #337: net-virtio: Driver does not work without disable-modern=on
In contrast to other virtio drivers that we can run in both legacy and modern modes, virtio-net only works with the legacy transport. This bug needs further investigation.

avdgrinten:
Fix #337


no92 (approved):

Dennisbonke (approved):
LGTM

PR #1379, opened 2026-05-24
eir: Fix missing initgraph edges and inconsistencies

avdgrinten:
Fix initgraph tasks that are either not strictly enough or too strictly ordered. Also take into account that some systems can expose both a DT and ACPI tables.


marv7000 (approved):
LGTM

PR #1381, opened 2026-05-25
eir, thor: Remove EirInfo struct in favor of ELF notes

avdgrinten:
This PR completes the transition from a fixed data structure to separate ELF notes.


marv7000 (approved):
LGTM

PR #1386, opened 2026-05-27
drivers: Add a HDA driver

Qwinci:
Add a HDA driver using uHDA. Depends on https://github.com/managarm/bootstrap-managarm/pull/711

### DIFF
--- a/drivers/libsound/src/card.cpp
+++ b/drivers/libsound/src/card.cpp
@@ -1,0 +1,268 @@
+#include "sound.hpp"
+
+#include <protocols/mbus/client.hpp>
+#include <protocols/fs/server.hpp>
+
+#include <async/recurring-event.hpp>
+
+#include <bragi/helpers-all.hpp>
+#include <bragi/helpers-std.hpp>
+
+#include <linux/types.h>
+#include <sound/asound.h>
+
+#include <fcntl.h>
+#include <unistd.h>
+#include <string.h>
+#include <sys/epoll.h>
+
+#include <print>
+
+namespace {
+
+constexpr bool logRequests = true;
+
+}
+
+namespace alsa {
+
+struct CardFile {
+	static async::result<void>
+	ioctl(void *object, uint32_t id, helix_ng::RecvInlineResult msg,
+			helix::UniqueLane conversation);
+
+	static helix::UniqueLane serve(smarter::shared_ptr<CardFile> file);
+
+	CardFile(sound::Card *card) : card{card} { }
+
+	sound::Card *card;
+};
+
+namespace {
+
+constexpr char cardId[] = "PCH";
+constexpr char driverName[] = "HDA-Intel";
+constexpr char cardName[] = "HDA Intel PCH";
+
+}
+
+async::result<void> CardFile::ioctl(void *object, uint32_t id, helix_ng::RecvInlineResult msg,
+		helix::UniqueLane conversation) {
+	auto self = static_cast<CardFile *>(object);
+
+	if (id != managarm::fs::GenericIoctlRequest::message_id) {
+		std::println(std::cout, "libsound: unknown control ioctl() message with ID {}", id);
+		auto [dismiss] = co_await helix_ng::exchangeMsgs(
+			conversation, helix_ng::dismiss());
+		HEL_CHECK(dismiss.error());
+		co_return;
+	}
+
+	auto req = bragi::parse_head_only<managarm::fs::GenericIoctlRequest>(msg);
+	assert(req);
+	if (req->command() == SNDRV_CTL_IOCTL_CARD_INFO) {
+		if constexpr (logRequests) {
+			std::println(std::cout, "libsound: query card info");
+		}
+
+		managarm::fs::GenericIoctlReply resp;
+
+		resp.set_error(managarm::fs::Errors::SUCCESS);
+		resp.set_result(0);
+
+		snd_ctl_card_info info{};
+		// TODO: get the card index somehow
+		info.card = 0;
+		memcpy(info.id, cardId, sizeof(cardId));
+		memcpy(info.driver, driverName, sizeof(driverName));
+		memcpy(info.name, cardName, sizeof(cardName));
+		memcpy(info.longname, cardName, sizeof(cardName));
+
+		auto [send_resp, send_data] = co_await helix_ng::exchangeMsgs(
+			conversation,
+			helix_ng::sendBragiHeadOnly(resp, frg::stl_allocator{}),
+			helix_ng::sendBuffer(&info, sizeof(info))
+		);
+		HEL_CHECK(send_resp.error());
+		HEL_CHECK(send_data.error());
+	} else if (req->command() == SNDRV_CTL_IOCTL_PVERSION) {
+		if constexpr (logRequests) {
+			std::println(std::cout, "libsound: query version");
+		}
+
+		managarm::fs::GenericIoctlReply resp;
+
+		resp.set_error(managarm::fs::Errors::SUCCESS);
+		resp.set_result(0);
+		resp.set_snd_pversion(SNDRV_PROTOCOL_VERSION(2, 0, 9));
+
+		auto [send_resp] = co_await helix_ng::exchangeMsgs(
+			conversation,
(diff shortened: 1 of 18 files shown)


drivers/libsound/src/card.cpp
```diff
@@ -0,0 +1,267 @@
...
+			conversation,
+			helix_ng::sendBragiHeadOnly(resp, frg::stl_allocator{})
+		);
+		HEL_CHECK(send_resp.error());
+	} else if (req->command() == SNDRV_CTL_IOCTL_PCM_INFO) {
+		if constexpr (logRequests) {
+			std::println(std::cout, "libsound: query pcm info");
+		}
+
+		snd_pcm_info info;
```
no92:
```suggestion
		snd_pcm_info info{};
```

drivers/libsound/src/card.cpp
```diff
@@ -0,0 +1,267 @@
...
+			helix_ng::recvBuffer(&info, sizeof(info))
+		);
+		HEL_CHECK(recv_buffer.error());
+
+		// device + stream + subdevice are filled
+		// TODO: get the card index somehow
+		info.card = 0;
+		memcpy(info.id, "uHDA", sizeof("uHDA"));
+		memcpy(info.name, "uHDA", sizeof("uHDA"));
+		memcpy(info.subname, "subdevice #0", sizeof("subdevice #0"));
```
no92: `strlcpy` seems more appropriate.

drivers/libsound/src/device.cpp
```diff
@@ -0,0 +1,942 @@
...
+	snd_pcm_hw_params hwParams{};
+	snd_pcm_sw_params swParams{};
+	size_t frameSize{};
+	uint32_t lastRemaining{};
+	uint64_t eventSequence{};
+	async::recurring_event eventBell;
+};
+
+struct Params {
+	snd_mask &mask(uint32_t what) {
```
no92: `const`?

drivers/libsound/src/device.cpp
```diff
@@ -0,0 +1,942 @@
...
+		return values.masks[what - SNDRV_PCM_HW_PARAM_FIRST_MASK];
+	}
+
+	snd_interval &interval(uint32_t what) {
```
no92: `const`?

drivers/libsound/src/device.cpp
```diff
@@ -0,0 +1,942 @@
...
+	bool nonBlock;
+	snd_pcm_hw_params hwParams{};
+	snd_pcm_sw_params swParams{};
+	size_t frameSize{};
+	uint32_t lastRemaining{};
+	uint64_t eventSequence{};
+	async::recurring_event eventBell;
+};
+
+struct Params {
```
no92, line 104: The member functions seem like they could all be marked `constexpr`.

drivers/libsound/src/device.cpp
```diff
@@ -0,0 +1,942 @@
...
+	{SNDRV_PCM_HW_PARAM_BUFFER_BYTES, &bufferBytesRule}
+};
+
+} // namespace rules
+
+async::result<void> DeviceFile::ioctl(void *object, uint32_t id, helix_ng::RecvInlineResult msg,
+		helix::UniqueLane conversation) {
+	auto self = static_cast<DeviceFile *>(object);
+
+	if (id != managarm::fs::GenericIoctlRequest::message_id) {
```
no92, line 443: Would be nice to add separate messages for this.

drivers/libsound/src/device.cpp
```diff
@@ -0,0 +1,942 @@
...
+	(void) cancellation;
+
+	auto self = static_cast<DeviceFile *>(object);
+
+	if(sequence > self->eventSequence)
+		co_return protocols::fs::Error::illegalArguments;
+
+	// Wait until we surpass the input sequence.
+	while(sequence == self->eventSequence)
+		co_await self->eventBell.async_wait();
```
no92: `async_wait_if`?

drivers/libsound/src/device.cpp
```diff
@@ -0,0 +1,942 @@
...
+
+	co_return protocols::fs::PollWaitResult{self->eventSequence,
+			self->eventSequence > 0 ? EPOLLOUT : 0};
```
no92: The logic between `pollWait` and `pollStatus` looks different - is this intentional?

drivers/libsound/src/device.cpp
```diff
@@ -0,0 +1,942 @@
...
+					req.flags() & managarm::fs::OpenFlags::OF_NONBLOCK);
+			auto remote_lane = DeviceFile::serve(file);
+
+			managarm::fs::SvrResponse resp;
+			resp.set_error(managarm::fs::Errors::SUCCESS);
+
+			auto ser = resp.SerializeAsString();
+			auto [send_resp, push_pt] = co_await helix_ng::exchangeMsgs(
+				conversation,
+				helix_ng::sendBuffer(ser.data(), ser.size()),
```
no92: `sendBragiHead{Tail,Only}`

drivers/sound/hda/src/controller.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+	async::result<void> run();
+
+	protocols::hw::Device device;
+	UhdaController *uhda;
+	const UhdaCodec *const *codecs;
+	size_t codecCount;
+};
+
+struct Stream : sound::Stream {
```
no92: `final`

drivers/sound/hda/src/kernel_api.cpp
```diff
@@ -0,0 +1,200 @@
...
+namespace {
+
+std::unordered_map<uintptr_t, void *> physicalMappings;
+
+}
+
+UhdaStatus uhda_kernel_allocate_physical(size_t size, uintptr_t *res) {
+	HelHandle memory;
+	void *window;
+	HEL_CHECK(helAllocateMemory(size, kHelAllocContinuous, nullptr, &memory));
```
no92, line 155: Is the `kHelAllocContinuous` necessary outside of the `helix::ptrToPhysical` use below?

drivers/sound/hda/src/kernel_api.cpp
```diff
@@ -0,0 +1,200 @@
...
+	*spinlock = new std::mutex{};
+	return UHDA_STATUS_SUCCESS;
+}
+
+void uhda_kernel_free_spinlock(void *spinlock) {
+	delete static_cast<std::mutex *>(spinlock);
+}
+
+UhdaIrqState uhda_kernel_lock_spinlock(void *spinlock) {
+	static_cast<std::mutex *>(spinlock)->lock();
```
no92: Why not something like `frg::ticket_spinlock` or `pthread_spin_lock`?

meson.build
```diff
@@ -174,6 +174,8 @@ if build_kernel
 endif
 
 if build_drivers
+	uhda = subproject('uhda')
```
no92: It might be nice to have uHDA be discoverable via pkg-config instead. libarch or libsmarter are small examples for how this is wired up in meson.

drivers/libsound/src/device.cpp
Qwinci: The returned references are used to modify the values in some cases and I don't think it really makes sense to add separate const overloads for these.

drivers/libsound/src/device.cpp
Qwinci: The whole poll logic (at least the wait part) isn't fully implemented, for an example eventSequence isn't updated anywhere. The condition was likely incorrect, I updated it to match `pollStatus`.

meson.build
Qwinci: Done.

drivers/sound/hda/src/kernel_api.cpp
avdgrinten: Without it, the memory wouldn't be physically contiguous, right? The contract here seems to be that the function returns physically contiguous memory.

Qwinci: It might not be necessary with the current way the API is used by uHDA where it only ever requests singular pages (0x1000 bytes in size) but its supposed to return contiguous memory so I think it makes sense to have the flag.

PR #1387, opened 2026-05-28
netserver: Support IPv4 fragmentation

Qwinci:
Add IPv4 fragmentation support to netserver and support setting route MTU using netlink RTA_METRICS.

### DIFF
--- a/servers/netserver/src/ip/ip4.cpp
+++ b/servers/netserver/src/ip/ip4.cpp
@@ -13,11 +13,19 @@
 #include <iostream>
 #include <iomanip>
 #include <protocols/fs/server.hpp>
+#include <helix/timer.hpp>
 #include <queue>
 
 using namespace protocols::fs;
 
 using Route = Ip4Router::Route;
+
+namespace {
+
+constexpr uint64_t fragmentExpiryTimeoutNs = uint64_t{30} * 1'000'000'000;
+constexpr uint64_t fragmentRouteExpiryTimeoutTicks = 4;
+
+} // anonymous namespace
 
 Ip4Router &ip4Router() {
 	static Ip4Router inst;
@@ -79,7 +87,11 @@ bool operator==(const Route &lhs, const Route &rhs) {
 	return operator<=>(lhs, rhs) == 0;
 }
 
-bool Ip4Packet::parse(arch::dma_buffer owner, arch::dma_buffer_view frame) {
+bool Ip4::FragmentRouteIdentification::operator<(const FragmentRouteIdentification &rhs) const {
+	return std::tie(sourceIp, destIp, protocol) < std::tie(rhs.sourceIp, rhs.destIp, rhs.protocol);
+}
+
+bool Ip4Packet::parse(arch::dma_buffer owner, arch::dma_buffer_view frame, bool resizeData) {
 	buffer_ = std::move(owner);
 	data = frame;
 	if (data.size() < sizeof(header)) {
@@ -94,8 +106,15 @@ bool Ip4Packet::parse(arch::dma_buffer owner, arch::dma_buffer_view frame) {
 	header.ensureEndian();
 
 	header.ihl = header.ihl & 0x0f;
-	// ensure we only access the correct parts of the buffer
-	data = data.subview(0, header.length);
+
+	// if this is a normal non-fragmented packet (fragmented packets may exceed header.length)
+	// ensure we only access the correct parts of the buffer.
+	if (resizeData)
+		data = data.subview(0, header.length);
+
+	if (data.size() < header.ihl * 4) {
+		return false;
+	}
 
 	Checksum csum;
 	csum.update(header_view());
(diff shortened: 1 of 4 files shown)


servers/netserver/src/netlink/queries.cpp
```diff
@@ -149,6 +149,39 @@ void NetlinkSocket::newRoute(struct nlmsghdr *hdr) {
 				route_changed = true;
 				break;
 			}
+			case RTA_METRICS: {
+				uint32_t locked = 0;
+				uint32_t rtax_mtu = 0;
```
no92: Might be nicer to use a `std::optional` instead.

servers/netserver/src/ip/ip4.cpp
```diff
@@ -370,6 +409,86 @@ void Ip4::feedPacket(nic::MacAddress, nic::MacAddress,
...
+			newHdr.length = std::min<size_t>(fragmentedPacket->data.size(), UINT16_MAX);
+			newHdr.ensureEndian();
+
+			Checksum chk;
+			chk.update(&newHdr, sizeof(newHdr));
+			newHdr.checksum = arch::convert_endian<arch::endian::big>(chk.finalize());
+
+			memcpy(buffer.data(), &newHdr, sizeof(newHdr));
+
+			arch::dma_buffer_view view = buffer;
```
no92, line 534: Instead of allocating a `dma_buffer` above, you can also just construct a `dma_buffer_view` from regular memory using the alternative constructor: `explicit dma_buffer_view(std::nullptr_t, void *data, size_t size)`

Qwinci: The buffer needs to be owned by the packet though, I guess it could be possible to add a vector inside `Ip4Packet` along with the dma buffer so then the vector from inside the `FragmentedPacket` could be moved there with a new constructor.

no92 (approved):

servers/netserver/src/ip/ip4.hpp
```diff
@@ -150,8 +156,24 @@ struct Ip4 {
...
+		bool lastReceived;
+	};
+
+	async::detached fragmentTimer_();
+
+	uint64_t fragmentTimerTick = 0;
+
 	std::multimap<int, smarter::shared_ptr<Ip4Socket>> sockets;
 	std::map<CidrAddress, std::weak_ptr<nic::Link>> ips;
+	std::unordered_map<uint16_t, FragmentedPacket> fragmentedPackets;
```
avdgrinten: This doesn't distinguish between source addresses. I'm not sure what the correct key is according to the RFCs / existing OSes but with the current key, different sources with the same ID can result in interleaved packets.

servers/netserver/src/ip/ip4.cpp
```diff
@@ -289,65 +304,89 @@ async::result<protocols::fs::Error> Ip4::sendFrame(Ip4TargetInfo ti,
...
+		// TODO(arsen): accomodate for options
+		chk.update(reinterpret_cast<void *>(&hdr), sizeof(hdr));
+		hdr.checksum = convert_endian<endian::big>(chk.finalize());
+
+		nic::Link::AllocatedBuffer fb;
+		if(mac) {
+			fb = target->allocateFrame(*mac, nic::ETHER_TYPE_IP4, packet_size);
+		} else {
+			fb = target->allocateFrame(packet_size);
+		}
```
avdgrinten, line 407: This allocates `packet_size`, not `fragmentLength` (+ header?).

servers/netserver/src/ip/ip4.cpp
```diff
@@ -370,6 +409,86 @@ void Ip4::feedPacket(nic::MacAddress, nic::MacAddress,
...
+			Checksum chk;
+			chk.update(&newHdr, sizeof(newHdr));
+			newHdr.checksum = arch::convert_endian<arch::endian::big>(chk.finalize());
+
+			memcpy(buffer.data(), &newHdr, sizeof(newHdr));
+
+			arch::dma_buffer_view view = buffer;
+
+			// The header is already parsed once so it has to be valid.
+			assert(hdr.parse(std::move(buffer), view, false));
```
avdgrinten: We should assume that asserts can be compiled out (even though we compile them into all builds).

servers/netserver/src/ip/ip4.cpp
```diff
@@ -289,65 +304,89 @@ async::result<protocols::fs::Error> Ip4::sendFrame(Ip4TargetInfo ti,
...
+	originalHdr.ident = globalPacketIdentification++;
+	originalHdr.flags_offset = 0;
+	originalHdr.ttl = 64;
+	originalHdr.protocol = proto;
+	// filled out later, 0 for purposes of computation
+	originalHdr.checksum = 0;
+	originalHdr.source = ti.source;
+	originalHdr.destination = ti.remote;
+
+	assert(fragmentMtu > header_size);
```
avdgrinten: We should probably just error out here (with a log?) instead of asserting.

servers/netserver/src/ip/ip4.cpp
```diff
@@ -289,65 +304,89 @@ async::result<protocols::fs::Error> Ip4::sendFrame(Ip4TargetInfo ti,
...
+	size_t fragment8Bytes = (fragmentMtu - header_size) / 8;
```
avdgrinten, line 362: What if the division clamps this to zero? Right now the assert can pass even though fragment8Bytes is zero.

Qwinci: That is a good point, I think it makes sense to add 1 to `fragment8Bytes` if `len <= fragmentMtu - header_size` so if the data fits to one packet (no fragmentation) then the below loop just uses len - progress as the packet size as it should (it's still possible for it to send a single no-data/zero-length packet but I don't think that is problematic and the old code also allowed that).

avdgrinten: The alternative would probably be to reject packets that result in zero fragment8Bytes (?).

Qwinci: That would make it impossible to send any packets less than 8 bytes or unnecessarily fragment all packets whose length isn't a multiple of 8, even if we want to reject zero-length packets I don't think this is the correct place to do it (it should be a check on `len`).

PR #1388, opened 2026-05-28
posix: Make findProcessWithCredentials() fallible

avdgrinten:
Fixes posix-subsystem crashes in case a process w/o credentials is able to do a POSIX request. This was reported by @Qwinci .


Qwinci (approved):
Looks good to me, there are a few places where `helix_ng::sendBuffer()` is used instead of `helix_ng::sendBragiHead{Only,Tail}` but it might make sense to do a separate pass to convert all of the existing uses of `helix_ng::sendBuffer()` to them.

no92 (approved):

Issue #1377, opened 2026-05-24
thor: AP boot must support VHE on ARM


avdgrinten:
This is needed to run on VHE systems.

PR #1389, opened 2026-05-28
eir, thor: Inital ACPI support for aarch64

marv7000:
Boots on QEMU with eir-limine and eir-uefi. Also disables logging (for eir-limine) when it's not possible to directly access device memory.

### DIFF
--- a/kernel/thor/arch/arm/gic.cpp
+++ b/kernel/thor/arch/arm/gic.cpp
@@ -1,3 +1,7 @@
+#include <frg/optional.hpp>
+#include <frg/scope_exit.hpp>
+#include <frg/vector.hpp>
+#include <thor-internal/acpi/acpi.hpp>
 #include <thor-internal/arch/gic.hpp>
 #include <thor-internal/arch/gic_v2.hpp>
 #include <thor-internal/arch/gic_v3.hpp>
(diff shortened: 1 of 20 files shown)


no92 (commented):
The eir + ACPI changes LGTM, I can't comment on the GIC code though.

kernel/thor/arch/arm/gic.cpp
```diff
@@ -8,18 +12,139 @@
...
+	auto *madt = reinterpret_cast<acpi_madt *>(madtTbl.ptr);
+	auto bspAffinity = getCpuData()->affinity;
+
+	frg::optional<acpi_madt_gicd> gicd;
+	uintptr_t cpuInterface = 0;
+	frg::vector<GicRedistributorRange, KernelAlloc> gicrRanges{*kernelAlloc};
+	frg::vector<GicRedistributorRange, KernelAlloc> giccRedistRanges{*kernelAlloc};
+
+	size_t offset = sizeof(acpi_madt);
+	while (offset < madt->hdr.length) {
```
d-tatianin: Consider using `uacpi_for_each_subtable`

kernel/thor/arch/arm/smp.cpp
```diff
@@ -271,20 +275,195 @@ bool bootSecondary(DeviceTreeNode *node, size_t cpuIndex) {
...
+		}
+		releaseAddress = ptr;
+	}
+
+	return bootSecondary(node->reg()[0].addr, cpuIndex, method, releaseAddress);
+}
+
+bool initPsciFromAcpi() {
+	uacpi_table fadtTbl;
+	if (uacpi_table_find_by_signature("FACP", &fadtTbl) != UACPI_STATUS_OK) {
```
d-tatianin: `uacpi_table_fadt()`

kernel/thor/arch/arm/smp.cpp
```diff
@@ -271,20 +275,195 @@ bool bootSecondary(DeviceTreeNode *node, size_t cpuIndex) {
...
+		panicLogger() << "thor: Unable to initialize APs, no MADT found" << frg::endlog;
+	frg::scope_exit finish{[&] { uacpi_table_unref(&madtTbl); }};
+
+	auto *madt = reinterpret_cast<acpi_madt *>(madtTbl.ptr);
+	auto bspAffinity = getCpuData()->affinity;
+
+	size_t apCpuIndex = 1;
+	size_t enabledCpuCount = 0;
+	size_t offset = sizeof(acpi_madt);
+	while (offset < madt->hdr.length) {
```
d-tatianin: again, maybe `uacpi_for_each_subtable` is better in this case

kernel/thor/arch/arm/trampoline.S
```diff
@@ -19,40 +19,53 @@ apEntryPoint:
...
-	b.eq .L_inEL1
+	b.eq .L_atEL1
 
 	// If we're in EL0 we're in trouble anyway
 .L_halt:
 	b .L_halt
 
 .L_inEL2:
-	// Drop to EL1
+	// Stay at EL2 and enable VHE. The kernel runs at EL2 on the BSP whenever VHE is supported.
```
avdgrinten: This would need to be gated by a check that VHE is supported. Maybe an easier option is to make a separate trampoline for VHE. We already have #1377 as a tracking issue so we could simply drop the VHE code here.

kernel/thor/arch/arm/smp.cpp
```diff
@@ -271,20 +275,182 @@ bool bootSecondary(DeviceTreeNode *node, size_t cpuIndex) {
...
+						        && hasAcpiField(generic.length,
+						            offsetof(acpi_madt_gicc, parked_address),
+						            sizeof(entry.parked_address))
+						        && entry.parking_protocol_version && entry.parked_address) {
+							bootSecondary(
+							    entry.mpidr,
+							    apCpuIndex,
+							    EnableMethod::spinTable,
+							    entry.parked_address
+							);
```
avdgrinten: It seems that the ACPI spin table layout differs from the DT one, so we need another `EnableMethod::spinTableOnAcpi` or similar.

kernel/thor/arch/arm/gic.cpp
```diff
@@ -8,18 +12,139 @@
...
     // Initialize the GIC.
     [] {
-	    if (initGicV2() || initGicV3()) {
+	    bool acpiGic = initGicFromAcpi();
+	    if (acpiGic || initGicV2() || initGicV3()) {
 		    initGicOnThisCpu();
+		    if (acpiGic)
+			    installAcpiGsiPins();
 	    }
     }
```
avdgrinten: This is broken. We should either take the ACPI code path or the DT code path but this first takes the ACPI code path and then (on failure) re-initializes the same global in `initGicV3()` on the DT code path.

kernel/thor/arch/arm/gic.cpp
```diff
@@ -8,18 +12,139 @@
...
+				        acpi::setGlobalSystemIrq(irq, pin);
+		        }
+	        },
+	        [](GicV3 *gic) {
+		        for (uint32_t irq = 16; irq < gic->irqCount(); ++irq) {
+			        auto pin = gic->getPin(irq);
+			        if (pin)
+				        acpi::setGlobalSystemIrq(irq, pin);
+		        }
+	        },
```
avdgrinten: GIC IRQ numbers 16 to 31 are processor local (PPIs) and not global (SGIs).

kernel/thor/arch/arm/gic_v2.cpp
```diff
@@ -154,8 +154,13 @@ auto GicDistributorV2::setupIrq(uint32_t irq, TriggerMode trigger) -> Pin * {
...
 }
 
+uint32_t GicDistributorV2::irqCount() const {
+	return irqPins_.size();
+}
+
 IrqStrategy GicDistributorV2::Pin::program(TriggerMode mode, Polarity polarity) {
-	bool success = setMode(mode, polarity);
+	(void)polarity;
+	bool success = setMode(mode, Polarity::high);
```
avdgrinten: If we ignore the polarity anyway (and it seems that the GIC does not have any polarity control), it doesn't make sense to pass it to `setMode()`.

kernel/thor/arch/arm/thor-internal/arch/gic.hpp
```diff
@@ -101,4 +102,6 @@ struct Gic : dt::IrqController {
 
 void initGicOnThisCpu();
 
+uint32_t affinityFromMpidr(uint64_t mpidr);
```
avdgrinten: If we're adding a helper function for this, we should also use in `cpu.cpp` where it's done in an ad-hoc way to set `affinity` for the BSP.

kernel/thor/arch/arm/smp.cpp
```diff
@@ -102,41 +134,22 @@ namespace {
...
-	while (i < methodProp->size()) {
-		frg::string_view sv{reinterpret_cast<const char *>(methodProp->data()) + i};
-		i += sv.size() + 1;
-
-		if (sv == "psci")
-			method = EnableMethod::psci;
-		else if (sv == "spin-table")
-			method = EnableMethod::spinTable;
-	}
+uint32_t affinityFromMpidr(uint64_t mpidr);
```
avdgrinten: This forward declaration is useless.

kernel/thor/arch/arm/thor-internal/arch/gic.hpp
avdgrinten: Actually, this function probably belongs to `cpu.hpp`, not `gic.hpp` as well.

kernel/thor/arch/arm/gic_v3.cpp
```diff
@@ -64,14 +71,18 @@ namespace dist_router {
 namespace redist_reg {
 	static constexpr arch::bit_register<uint64_t> type{0x8};
 	static constexpr arch::bit_register<uint32_t> waker{0x14};
+	static constexpr arch::bit_register<uint32_t> pidr2{dist_reg::pidr2};
 }
 
+#define redist_pidr2 dist_pidr2
```
avdgrinten: Why is this done?

avdgrinten: A namespace alias would be better.

kernel/thor/arch/arm/stubs.S
```diff
@@ -254,7 +254,10 @@ forkExecutorRegisters:
 	str x4, [x0, THOR_FRAME_SP]
 
 	// Set new PSTATE
-	mov x4, #5
+	mrs x4, currentel
+	lsr x4, x4, #2
+	lsl x4, x4, #2
+	orr x4, x4, #1
```
avdgrinten, line 261: Why is this done?

PR #1361, opened 2026-05-10
IOMMU work, part 3: kernel infrastructure setup

no92:

PR #1399, opened 2026-05-31
IOMMU, part 4: Queued Invalidation + more wiring

### LINKED ISSUE #1416: thor: VT-d code needs to handle non-coherent IOMMUs
We need to issue proper cache flushes for non-coherent IOMMUs.

no92:
Currently stacked on top of #1361.

### DIFF
--- a/core/arch/include/arch/os/dma_pool.hpp
+++ b/core/arch/include/arch/os/dma_pool.hpp
@@ -2,6 +2,9 @@
 
 #include <arch/dma_structs.hpp>
 #include <array>
+#include <async/result.hpp>
+#include <helix/ipc.hpp>
+#include <helix/memory.hpp>
 #include <mutex>
 #include <stddef.h>
 #include <stdint.h>
@@ -16,7 +19,15 @@ struct contiguous_pool_options {
 	// end up on different cache lines (such that writes to an allocation cannot cause
 	// adjacent cache lines to become dirty).
 	size_t minAllocationGap{64};
+	// TODO: default to non-continous, so that devices requiring contigous physical allocations
+	// have to opt-in to this behavior.
+	int allocFlags = kHelAllocContinuous;
+	// Additional flags passed when mapping the memory into DMA space.
+	uint32_t dmaMapFlags = 0;
 };
+
+struct dma_space;
+struct imported_dma_buffer;
 
 struct contiguous_pool : dma_pool {
 private:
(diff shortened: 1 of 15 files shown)


kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -572,6 +574,282 @@ struct IntelIommu final : Iommu, IrqSink {
...
+			while (length) {
+				// Determine the natural alignment of this address.
+				// (addr & -addr) isolates the lowest set bit, yielding the max power-of-2 alignment.
+				size_t alignment = (addr == 0) ? maxChunk : (addr & -addr);
+				size_t largestFittingChunk = std::bit_floor(length);
+				size_t chunk = std::min({alignment, largestFittingChunk, maxChunk});
+				assert(chunk);
+				auto am = std::countr_zero(chunk) - 12;
+
+				auto index = qiRingIdx_++;
```
avdgrinten: We have to check (also in the other `invalidate*()` functions) that enough space is available in the ring. We need to store a software copy of the ring head and tail for that purpose and advance it via a `recurring_event`.

kernel/thor/generic/thor-internal/util.hpp
```diff
@@ -43,4 +44,58 @@ inline FreqFraction computeFreqFraction(uint64_t num, uint64_t denom) {
...
+		auto temp = *this;
+		--(*this);
+		return temp;
+	}
+
+	constexpr bool operator==(const QueueIndex &other) const noexcept = default;
+
+private:
+	size_t index_;
+	size_t mod_;
```
avdgrinten: Using a modulo is a poor choice if we know that the size is a power of two. There are many cases in which we do not know the size at compile time. If we do know it at compile time, we could make it a template parameter.

core/arch/include/arch/os/dma_pool.hpp
```diff
@@ -32,34 +42,136 @@ struct contiguous_pool : dma_pool {
...
+
 	contiguous_pool(contiguous_pool_options options = {});
 
 	dma_ptr allocate(size_t size, size_t count, size_t align) override;
 	void deallocate(dma_ptr ptr, size_t size, size_t count, size_t align) override;
 
-private:
+	dma_space attachDmaSpace(helix::BorrowedDescriptor ioSpace, bool iommuActive);
+
+	dma_buffer_view importMemory(helix::BorrowedDescriptor memory, size_t offset, size_t size, uint32_t flags = 0);
```
avdgrinten: AFAICT the region that is constructed here is never destructed. This should probably return some `imported_contiguous_memory` object (or similar) that owns the imported region (with a `.view()` function that returns the `dma_buffer_view`).

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -572,6 +574,282 @@ struct IntelIommu final : Iommu, IrqSink {
...
+			iommu_->regs_.store(regs::invalidationEventControl, invalidationEventControl::interruptMask(false));
+		}
+
+		IrqStatus raise() override {
+			if (iommu_->regs_.load(regs::invalidationCompletionStatus) & invalidationCompletionStatus::complete) {
+				iommu_->regs_.store(regs::invalidationCompletionStatus, invalidationCompletionStatus::complete(true));
+
+				PageAccessor statusPage{qiStatusPhys_};
+
+				for (auto [i, a] : completions_ | std::views::enumerate) {
```
avdgrinten: This function accesses `completions_` without any locking.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -572,6 +574,282 @@ struct IntelIommu final : Iommu, IrqSink {
...
+					auto status = reinterpret_cast<volatile uint32_t *>(
+					    reinterpret_cast<uintptr_t>(statusPage.get()) + 0x100 + sizeof(uint32_t) * i
```
avdgrinten: AFIACT offset zero is used for the synchronous path but there is no reason not to always use `sizeof(uint32_t) * index`.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -572,6 +574,282 @@ struct IntelIommu final : Iommu, IrqSink {
...
+					);
+					if (*status) {
+						a->complete();
+						completions_[i] = nullptr;
+					}
```
avdgrinten: AFAICT the status page is never cleared.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -572,6 +574,282 @@ struct IntelIommu final : Iommu, IrqSink {
...
+		void invalidateRange(uint16_t domain, ShootNode *node) {
+			assert((node->address & 0xFFF) == 0);
+
+			auto length = node->size;
+			assert(length);
+			assert((length & 0xFFF) == 0);
+
+			auto addr = node->address;
+			uint8_t maxAddressMask = iommu_->cap_ & capability::maximumAddressMask;
+			size_t maxChunkPages = 1 << maxAddressMask;
```
avdgrinten: If `maxAddressMask` is >= 31, this becomes UB since `1` is treated as integer.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -572,6 +574,282 @@ struct IntelIommu final : Iommu, IrqSink {
...
+		using AsyncCompletionQueue = frg::intrusive_rcu_list<
+			AsyncCompletion,
+			frg::locate_member<
+				AsyncCompletion,
+				frg::intrusive_rcu_list_hook<AsyncCompletion>,
+				&AsyncCompletion::hook
+			>
+		>;
+
+		std::array<ShootNode *, 256> completions_;
```
avdgrinten: This is default initialization, not zero initialization.

core/arch/include/arch/os/dma_pool.hpp
```diff
@@ -32,34 +42,136 @@ struct contiguous_pool : dma_pool {
...
 
 	// Protected by mutex_.
 	std::array<bucket, num_size_classes> buckets_;
 };
 
+struct dma_space {
+	dma_space(size_t i, contiguous_pool *p, helix::BorrowedDescriptor space, bool iommuActive)
+	: index_{i}, pool_{p}, space_{space}, iommuActive_{iommuActive} {
+		pool_->spaces_.insert(pool_->spaces_.begin() + i, this);
+	}
```
avdgrinten, line 190: If we store `this` into `pool_->space_`, then the `dma_space` class needs to be non-movable. Also, `spaces_` needs to be protected by a mutex (like `buckets_`).

Alternatively, we could store all relevant data in `contiguous_pool` and let this class only remember the index (then it can become movable).

core/arch/include/arch/os/dma_pool.hpp
```diff
@@ -32,34 +42,136 @@ struct contiguous_pool : dma_pool {
...
 public:
+	friend dma_space;
+
 	contiguous_pool(contiguous_pool_options options = {});
 
 	dma_ptr allocate(size_t size, size_t count, size_t align) override;
 	void deallocate(dma_ptr ptr, size_t size, size_t count, size_t align) override;
 
-private:
+	dma_space attachDmaSpace(helix::BorrowedDescriptor ioSpace, bool iommuActive);
```
avdgrinten, line 53: AFAICT there is not way to detach a `dma_space`. Is the `dma_space` supposed to be an owning type or not?

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -572,6 +574,282 @@ struct IntelIommu final : Iommu, IrqSink {
...
+
+private:
+		// 128-bit QI descriptor
+		struct QiRingEntry128 {
+			arch::bit_value<uint64_t> low;
+			arch::bit_value<uint64_t> high;
+		};
+
+		// must be called with `qiRingLock_` held.
+		void submitSync() {
```
avdgrinten: Is synchronous submission needed? We could always submit async and use `KernelFiber::asyncBlockCurrent()` during initialization. That seems cleaner to me.

core/arch/src/dma_pool.cpp
```diff
@@ -50,13 +50,18 @@ dma_ptr contiguous_pool::allocate(size_t size, size_t count, size_t align) {
 }
 
 void contiguous_pool::deallocate(dma_ptr ptr, size_t size, size_t count, size_t align) {
+	auto rn = static_cast<region *>(ptr.region());
+	if (rn->imported()) {
+		delete rn;
+		return;
+	}
```
avdgrinten: Deallocate should not deal with imported regions, the purpose of deallocate is deallocating pointer that were obtained by allocate().

core/arch/src/dma_pool.cpp
```diff
@@ -69,29 +74,71 @@ void contiguous_pool::deallocate(dma_ptr ptr, size_t size, size_t count, size_t
...
+	helix::UniqueDescriptor handle{memory.dup()};
+	HEL_CHECK(helMapMemory(
+	    handle.getHandle(),
+	    kHelNullHandle,
+	    nullptr,
+	    offset,
+	    size,
+	    kHelMapProtRead | kHelMapProtWrite | flags,
+	    &p
+	));
```
avdgrinten: It's cleaner to let `region` own the mapping.

core/arch/src/dma_pool.cpp
```diff
@@ -69,29 +74,71 @@ void contiguous_pool::deallocate(dma_ptr ptr, size_t size, size_t count, size_t
...
+				auto dmaSpace = regionPool->spaces_[index];
+				HEL_CHECK(helUnmapMemory(dmaSpace->descriptor().getHandle(), reinterpret_cast<void *>(*ioVa), size));
+			}
+		}
+	}
+
+	// For imported regions, we are responsible for unmapping it from cpuVa.
+	if (base_va && imported_) {
+		HEL_CHECK(helUnmapMemory(kHelNullHandle, reinterpret_cast<void *>(*base_va), size));
+	}
```
avdgrinten: If `region` owns the mapping, this becomes unconditional (and it's dropped from deallocate).

core/arch/include/arch/os/dma_pool.hpp
```diff
@@ -32,34 +43,190 @@ struct contiguous_pool : dma_pool {
...
+		swap(a.ptr_, b.ptr_);
+		swap(a.size_, b.size_);
+	}
+
+	imported_dma_buffer(const imported_dma_buffer &) = delete;
+	imported_dma_buffer &operator=(const imported_dma_buffer &) = delete;
+
+	imported_dma_buffer(imported_dma_buffer &&other) noexcept {
+		swap(*this, other);
+	}
```
avdgrinten: This needs to delegate to the default ctor, otherwise `*this` contains garbage.

core/arch/include/arch/os/dma_pool.hpp
```diff
@@ -32,34 +43,190 @@ struct contiguous_pool : dma_pool {
...
 	int shift_of_(size_t size, size_t count, size_t align);
 
-	void *allocate_pages_(size_t region_size);
+	std::pair<void *, helix::UniqueDescriptor> allocate_pages_(size_t region_size);
 	void deallocate_pages_(void *p, size_t region_size);
 
 	contiguous_pool_options options_;
 
-	std::mutex mutex_;
+	size_t attachedDmaSpaces_ = 0;
```
avdgrinten, line 131: This needs to be protected by the mutex.

core/arch/include/arch/os/dma_pool.hpp
```diff
@@ -32,34 +43,190 @@ struct contiguous_pool : dma_pool {
...
+		}
+
+	private:
+		size_t size;
+
+		helix::UniqueDescriptor backingMemory_;
+		size_t backingMemoryOffset_;
+
+		// Holds the `ioVa` of the region in the DMA space, if already mapped there.
+		std::vector<std::optional<uintptr_t>> dmaSpaces_;
```
avdgrinten, line 115: This needs to be protected by the pool's `spacesMutex_`.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -582,8 +588,516 @@ struct IntelIommu final : Iommu, IrqSink {
...
+		      frg::string(*kernelAlloc, iommu->name())
+		      + frg::string<KernelAlloc>{*kernelAlloc, "-qi"}
+		  ),
+		  iommu_{iommu} {
+			qiRingPhys_ = physicalAllocator->allocate(0x1000);
+			PageAccessor qiRing{qiRingPhys_};
+			memset(qiRing.get(), 0, 0x1000);
+			qiStatusPhys_ = physicalAllocator->allocate(0x1000);
+			PageAccessor qiStatus{qiStatusPhys_};
+			memset(qiStatus.get(), 0, 0x1000);
```
avdgrinten, line 605: We should panic on allocation failure here instead of faulting.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -582,8 +588,516 @@ struct IntelIommu final : Iommu, IrqSink {
...
+		}
+
+		void init() {
+			auto name = frg::string<KernelAlloc>{*kernelAlloc, "iommu"}
+			            + frg::to_allocated_string(*kernelAlloc, iommu_->id())
+			            + frg::string<KernelAlloc>{*kernelAlloc, "-invalidation-msi"};
+
+			auto interrupt = allocateApicMsi(name);
+			assert(interrupt);
```
avdgrinten: This should be handled gracefully.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -582,8 +588,516 @@ struct IntelIommu final : Iommu, IrqSink {
...
+						auto a = completions_[index].exchange(nullptr, std::memory_order_acquire);
+						if (a)
+							a->complete();
+					}
+					qiHead_.store(completedSeq, std::memory_order_release);
+					qiProgressEvent_.raise();
+				}
+			}
+
+			return IrqStatus::acked;
```
avdgrinten, line 687: Nack in the else case?

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -582,8 +588,516 @@ struct IntelIommu final : Iommu, IrqSink {
...
+			spawnOnWorkQueue(*kernelAlloc, WorkQueue::generalQueue().lock(), runSubmissionLoop());
+		}
+
+	private:
+		IrqStatus raise() override {
+			if (iommu_->regs_.load(regs::invalidationCompletionStatus) & invalidationCompletionStatus::complete) {
+				iommu_->regs_.store(regs::invalidationCompletionStatus, invalidationCompletionStatus::complete(true));
+
+				PageAccessor statusPage{qiStatusPhys_};
+				volatile uint32_t *status = reinterpret_cast<volatile uint32_t *>(statusPage.get());
```
avdgrinten: The status page should be treated as an atomic, not as volatile (since we're actually relying on it being an atomic load here).

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -582,8 +588,516 @@ struct IntelIommu final : Iommu, IrqSink {
...
+				uint32_t completedSeq = *status;
+				uint32_t head = qiHead_.load(std::memory_order_relaxed);
+
+				if (static_cast<int32_t>(completedSeq - head) > 0) {
```
avdgrinten: This does not handle wrap around correctly.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -582,8 +588,516 @@ struct IntelIommu final : Iommu, IrqSink {
...
+				}
+			}
+
+			return IrqStatus::acked;
+		}
+
+		// must be called with `qiRingLock_` held.
+		bool isCompleted(uint32_t seq) {
+			uint32_t head = qiHead_.load(std::memory_order_acquire);
+			return static_cast<int32_t>(head - seq) >= 0;
```
avdgrinten, line 693: This also does not handle wrap around correctly.

kernel/thor/arch/x86/vt_d.cpp
```diff
@@ -582,8 +588,516 @@ struct IntelIommu final : Iommu, IrqSink {
...
+			qiTail_.store(tail + 1, std::memory_order_release);
+			return QueueIndex<invalidationQueueSize>{tail};
+		}
+
+		// must be called with `qiRingLock_` held.
+		std::pair<QueueIndex<invalidationQueueSize>, uint32_t> allocateSeq() {
+			uint32_t seq = qiTail_.load(std::memory_order_relaxed);
+			qiTail_.store(seq + 1, std::memory_order_release);
+			return {QueueIndex<invalidationQueueSize>{seq}, seq};
+		}
```
avdgrinten: What's the difference?

kernel/thor/arch/x86/vt_d.cpp
avdgrinten: Discussed offline: it does if we don't cast.

kernel/thor/arch/x86/vt_d.cpp
no92: `allocateSlot` does not return a seqnum, which is not needed for non-wait descriptors.

PR #1404, opened 2026-06-04
protocols/usb: Switch to range-based descriptor walking API

### LINKED ISSUE #1277: usb: Refactor configuration walking to use C++ iterators
For example, this could be done by using a `std::generator`.

avdgrinten:
This avoids multiple problems of the previous visitor-based API.

### DIFF
--- a/protocols/usb/include/protocols/usb/usb.hpp
+++ b/protocols/usb/include/protocols/usb/usb.hpp
@@ -1,11 +1,18 @@
 #pragma once
 
 #include <arch/bits.hpp>
+#include <array>
 #include <assert.h>
+#include <bit>
 #include <cstdint>
 #include <cstring>
+#include <iterator>
+#include <ranges>
+#include <span>
 #include <string>
+#include <string_view>
 #include <optional>
+#include <utility>
 #include <vector>
 
 namespace protocols::usb {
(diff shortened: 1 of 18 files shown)


protocols/usb/include/protocols/usb/usb.hpp
```diff
@@ -316,4 +323,204 @@ void walkConfiguration(std::string buffer, F functor) {
...
+// for a truncated descriptor instead of reading out of bounds. Returning a copy
+// also avoids reading through a potentially unaligned pointer into the buffer.
+template<typename T>
+std::optional<T> extractDescriptor(std::span<const std::byte> bytes) {
+	if(bytes.size() < sizeof(T))
+		return std::nullopt;
+	std::array<std::byte, sizeof(T)> storage;
+	std::memcpy(storage.data(), bytes.data(), sizeof(T));
+	return std::bit_cast<T>(storage);
+}
```
avdgrinten, line 390: This pattern is weird, I still have to fix this.

avdgrinten: Actually, never mind. This is needed because `CdcNcm` is not default constructible (and I don't want to refactor that right now).

no92 (approved):

PR #1406, opened 2026-06-04
drivers: Make addressBits mandatory for contiguous_pool

Qwinci:
Make providing address bits to `arch::dma_pool` mandatory and make all existing drivers use supported values for them.

### DIFF
--- a/drivers/nic/bcmgenet/src/main.hpp
+++ b/drivers/nic/bcmgenet/src/main.hpp
@@ -88,7 +88,7 @@ private:
 	std::vector<arch::dma_buffer> txBufs_;
 	std::vector<arch::dma_buffer> rxBufs_;
 	arch::mem_space space_;
-	arch::contiguous_pool dmaPool_;
+	arch::contiguous_pool dmaPool_{{.addressBits = 64}};
 	nic::PhyMode phyMode_;
 
 	async::recurring_event rxEvent_;
--- a/drivers/nic/usb_net/src/usb-net.hpp
+++ b/drivers/nic/usb_net/src/usb-net.hpp
@@ -14,7 +14,7 @@ struct UsbNic : nic::Link {
 protected:
 	virtual async::detached listenForNotifications() = 0;
 
-	arch::contiguous_pool dmaPool_;
+	arch::contiguous_pool dmaPool_{{.addressBits = 64}};
 	protocols::usb::Device device_;
 
 	/* the control interface */
--- a/drivers/usb/devices/serial/src/controller.hpp
+++ b/drivers/usb/devices/serial/src/controller.hpp
@@ -49,7 +49,7 @@ struct Controller {
 	struct termios activeSettings = {};
 	bool nonBlock_;
 	protocols::usb::Device hw_;
-	arch::contiguous_pool pool_;
+	arch::contiguous_pool pool_{{.addressBits = 64}};
 };
 
 struct WriteRequest {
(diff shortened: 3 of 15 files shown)


drivers/nic/bcmgenet/src/main.hpp
```diff
@@ -88,7 +88,7 @@ struct BcmGenetNic : nic::Link {
 	std::vector<arch::dma_buffer> txBufs_;
 	std::vector<arch::dma_buffer> rxBufs_;
 	arch::mem_space space_;
-	arch::contiguous_pool dmaPool_;
+	arch::contiguous_pool dmaPool_{{.addressBits = 64}};
```
avdgrinten, line 91: CC @qookei IIRC there were some constraints here?

drivers/nic/usb_net/src/usb-net.hpp
```diff
@@ -14,7 +14,8 @@ struct UsbNic : nic::Link {
 protected:
 	virtual async::detached listenForNotifications() = 0;
 
-	arch::contiguous_pool dmaPool_;
+	// TODO: This doesn't always need to be 32 bits, that depends on the HCD.
+	arch::contiguous_pool dmaPool_{{.addressBits = 32}};
```
avdgrinten: The HCD is in a different process and the packets are copied, not shared. So this can always be 64.

drivers/usb/devices/serial/src/controller.hpp
```diff
@@ -49,7 +49,8 @@ struct Controller {
 	struct termios activeSettings = {};
 	bool nonBlock_;
 	protocols::usb::Device hw_;
-	arch::contiguous_pool pool_;
+	// TODO: This doesn't always need to be 32 bits, that depends on the HCD.
+	arch::contiguous_pool pool_{{.addressBits = 32}};
```
avdgrinten: The HCD is in a different process and the packets are copied, not shared. So this can always be 64.

drivers/usb/devices/serial/src/main.cpp
```diff
@@ -45,7 +45,8 @@ using WriteQueue = frg::intrusive_list<
 
 std::vector<smarter::shared_ptr<Controller>> controllers;
 
-arch::contiguous_pool pool;
+// TODO: This doesn't always need to be 32 bits, that depends on the HCD.
+arch::contiguous_pool pool{{.addressBits = 32}};
```
avdgrinten: The HCD is in a different process and the packets are copied, not shared. So this can always be 64.

drivers/usb/hcds/xhci/src/main.cpp
```diff
@@ -39,7 +39,7 @@ Controller::Controller(protocols::hw::Device hw_device, mbus_ng::Entity entity,
 		helix::UniqueDescriptor mmio, helix::UniqueIrq irq, std::string name)
 : _hw_device{std::move(hw_device)}, _mapping{std::move(mapping)},
 		_mmio{std::move(mmio)}, _irq{std::move(irq)},
-		_space{_mapping.get()}, _name{name}, _memoryPool{},
+		_space{_mapping.get()}, _name{name}, _memoryPool{{.addressBits = 64}},
```
avdgrinten: Also CC @qookei IIRC there are constraints on Raspberry Pi.

qookei: Yes with our current configuration in Thor on the Pi 4 this should probably be limited to 31 bits. Also I don't think we have a check for it but technically the controller can indicate it can only deal with 32-bit pointers.

drivers/nic/bcmgenet/src/main.hpp
qookei: I don't immediately see any limitations here. AFAICT the DMA address limitation is specific to the PCIe RC in the Pi4

PR #1414, opened 2026-06-06
Set the process path to the right location for /proc/self/exe

### LINKED ISSUE #1405: posix: /proc/self/exe doesn't point to a fully resolved path
For example, on Linux:
```
$ /bin/file /proc/self/exe
/proc/self/exe: symbolic link to /usr/bin/file
```

But on managarm:
```
$ /bin/file /proc/self/exe
/proc/self/exe: symbolic link to /bin/file
```

marv7000:
Fixes #1405

### DIFF
--- a/posix/subsystem/src/exec.hpp
+++ b/posix/subsystem/src/exec.hpp
@@ -6,6 +6,7 @@ struct ExecuteResult {
 	helix::UniqueDescriptor thread;
 	void *auxBegin = nullptr;
 	void *auxEnd = nullptr;
+	std::string path;
 };
 
 async::result<frg::expected<Error, ExecuteResult>>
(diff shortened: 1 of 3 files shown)


posix/subsystem/src/exec.hpp
```diff
@@ -6,6 +6,7 @@ struct ExecuteResult {
 	helix::UniqueDescriptor thread;
 	void *auxBegin = nullptr;
 	void *auxEnd = nullptr;
+	std::string path;
```
avdgrinten, line 9: On Linux, this points to the new direction in case the executable is renamed. We need to store the pointer to the `FsLink` and construct the path at `/self/proc/exe` read time.

PR #1420, opened 2026-06-13
IOMMU part 5: XHCI + libblockfs

no92:


drivers/usb/hcds/xhci/src/main.cpp
```diff
@@ -35,11 +35,22 @@ namespace proto = protocols::usb;
...
-		_space{_mapping.get()}, _name{name},
+Controller::Controller(
+    protocols::hw::Device hw_device,
+    mbus_ng::Entity entity,
+    helix::Mapping mapping,
+    helix::UniqueDescriptor mmio,
+    helix::UniqueIrq irq,
+    std::string name,
+    bool iommuActive,
+    helix::UniqueDescriptor ioSpace
```
avdgrinten: I don't like the naming `ioSpace` if the proper name that we gave to the concept is `dmaSpace`. I get that it bites with the `dmaSpace` that is a `arch::dma_space` but if that's the case, the latter should maybe be called `attachedDmaSpace`.

no92: `dmaSpaceHandle` is what I used for other drivers.

core/virtio/src/core.cpp
```diff
@@ -690,16 +731,26 @@ discover(protocols::hw::Device hw_device, DiscoverMode mode) {
...
+			    std::move(queueMsi),
+			    std::move(dmaSpace),
+			    iommuActive
+			);
 		}
 	}
 
 	if(mode == DiscoverMode::legacyOnly || mode == DiscoverMode::transitional) {
 #ifdef __x86_64__
+		assert(!iommuActive);
```
avdgrinten: What guarantees this?

drivers/usb/hcds/xhci/src/ring.cpp
```diff
@@ -156,37 +156,40 @@ EventRing::EventRing(Controller *controller)
...
+async::result<uintptr_t> EventRing::getErstPtr() {
+	return _controller->dmaSpace().iova_of(_erst);
 }
 
-uintptr_t EventRing::getEventRingPtr() {
-	return helix::ptrToPhysical(_eventRing.data()) + _dequeue.index * sizeof(RawTrb);
+async::result<uintptr_t> EventRing::getEventRingPtr() {
+	auto view = _eventRing.view_buffer().subview(_dequeue.index * sizeof(RawTrb), sizeof(RawTrb));
+	co_return co_await _controller->dmaSpace().iova_of(view);
 }
```
avdgrinten: We should probably cache the event ring pointer such that this does not need to be asynchronous.

core/virtio/src/core.cpp
avdgrinten: AFAICT it's not guaranteed, so it's better to just log + discard the device instead of driving it.

drivers/block/nvme/src/queue.cpp
```diff
@@ -110,6 +108,10 @@ async::result<void> PciExpressQueue::submitCommandToDevice(std::unique_ptr<Comma
 	cmdBuf.common.commandId = (uint16_t)slot;
 
 	memcpy((uint8_t *)sqCmds_ + (sqTail_ << 6), &cmdBuf, sizeof(spec::Command));
+
+	arch::dma_barrier barrier{true};
+	barrier.writeback((uint8_t *) sqCmds_ + (sqTail_ << 6), sizeof(spec::Command));
```
avdgrinten, line 113: This seems unrelated?

drivers/block/virtio-blk/src/block.cpp
```diff
@@ -37,53 +40,52 @@ void Device::runDevice() {
 	_transport->runDevice();
 
 	// perform device specific setup
-	virtRequestBuffer = new VirtRequest[_requestQueue->numDescriptors()];
-	statusBuffer = new uint8_t[_requestQueue->numDescriptors()];
-
-	// natural alignment makes sure that request headers do not cross page boundaries
-	assert((uintptr_t)virtRequestBuffer % sizeof(VirtRequest) == 0);
```
avdgrinten, line 44: This is needed in the !IOMMU case, isn't it?

drivers/block/virtio-blk/src/block.cpp
```diff
@@ -119,37 +122,33 @@ async::detached Device::_processRequests() {
...
+		    statusBuffer.object_view(chain.front().tableIndex()).view_buffer()
+		);
 
 		// Submit the request to the device
 		_requestQueue->postDescriptor(chain.front(), request,
 				[] (virtio_core::Request *base_request) {
 			auto request = static_cast<UserRequest *>(base_request);
-			if(logInitiateRetire)
-				std::cout << "Retiring " << request->numSectors
-						<< " data descriptors" << std::endl;
```
avdgrinten, line 152: Why is this dropped?

drivers/gfx/virtio/src/commands.cpp
```diff
@@ -34,143 +34,142 @@ struct AwaitableRequest : virtio_core::Request {
...
-	memset(&xfer, 0, sizeof(spec::XferToHost2d));
-	xfer.header.type = spec::cmd::xferToHost2d;
-	xfer.rect.x = 0;
-	xfer.rect.y = 0;
-	xfer.rect.width = width;
-	xfer.rect.height = height;
-	xfer.resourceId = resourceId;
-
-	spec::Header xfer_result;
+	arch::dma_object<spec::XferToHost2d> xfer{&device->dmaPool()};
```
avdgrinten: This default constructs and thus doesn't zero AFAICT. We could use a designated initializer to initialize the dma_object through.

This also applies to the cases below.

drivers/block/virtio-blk/src/block.cpp
no92: we don't have a `numSectors` member any more, and capturing here is not really possible.

drivers/block/nvme/src/queue.cpp
no92: This broke boot on the real machine in the !IOMMU case.

PR #1421, opened 2026-06-14
Procfs fixes

Qwinci:
Fixes for procfs maps format/reading and ignoring dirfd in openat for absolute paths like it should.

### DIFF
--- a/posix/subsystem/src/process.cpp
+++ b/posix/subsystem/src/process.cpp
@@ -104,6 +104,7 @@ auto VmContext::splitAreaOn_(uintptr_t addr, size_t size) ->
 		if (base < addr && (base + area.areaSize) > addr) {
 			Area right;
 			right.copyOnWrite = area.copyOnWrite;
+			right.stack = area.stack;
 			right.areaSize = area.areaSize - (addr - base);
 			right.nativeFlags = area.nativeFlags;
 			right.fileView = area.fileView.dup();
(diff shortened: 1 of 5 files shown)


posix/subsystem/src/procfs.cpp
```diff
@@ -669,15 +669,18 @@ async::result<std::expected<std::string, Error>> MapNode::show(Process *) {
 
 	auto vmContext = p->vmContext();
 	std::stringstream stream;
+
 	for (auto area : *vmContext) {
-		stream << std::hex << area.baseAddress();
+		auto startSize = stream.tellp();
+
+		stream << std::hex << std::setfill('0') << std::setw(8) << area.baseAddress();
```
no92: It might be nicer to just use std::format here instead.

posix/subsystem/src/process.cpp
```diff
@@ -129,7 +130,7 @@ auto VmContext::splitAreaOn_(uintptr_t addr, size_t size) ->
 async::result<frg::expected<Error, void *>>
 VmContext::mapFile(uintptr_t hint, helix::UniqueDescriptor memory,
 		smarter::shared_ptr<File, FileHandle> file,
-		intptr_t offset, size_t size, bool copyOnWrite, uint32_t nativeFlags) {
+		intptr_t offset, size_t size, bool copyOnWrite, uint32_t nativeFlags, bool stack) {
```
no92, line 133: It might be nicer to follow Linux here, where information like `[stack]` is tracked per-process/per-thread and not for each area. AFAICT, there can only be a single stack per thread, and the annotating might not work for threads set up with clone/clone3, so possibly even pthreads.

Dennisbonke (approved):
LGTM

PR #1432, opened 2026-06-17
eir: Load GDT using inline assembly

viktorpopp:
Resolves the to-do in [kernel/eir/arch/x86/i386.cpp](https://github.com/managarm/managarm/blob/master/kernel/eir/arch/x86/i386.cpp#L7).


no92 (commented):
We can remove the `eirLoadGdt` function altogether and just make the inline assembly part of `initArchCpu()`. The inline assembly itself LGTM.

Also, the commits should be squashed into a single, atomic commit.

PR #1436, opened 2026-06-18
drivers/kbd: laptop fixes

no92:


drivers/kbd/src/main.cpp
```diff
@@ -288,6 +291,13 @@ void Controller::submitCommand(controller_cmd::SendBytePort2) {
 	sendCommandByte(0xD4); // TODO: define a constant?
 }
 
+void Controller::submitCommand(controller_cmd::SelfTest) {
+	sendCommandByte(0xAA); // TODO: define a constant?
+	auto result = recvResponseByte(default_timeout);
+	assert(result != std::nullopt && "timed out");
+	assert(result == 0x55);
+}
```
avdgrinten: We should handle the errors.

no92: How should we handle failures here? The rest of the code seems to just assert/panic in similar cases.

avdgrinten (approved):

PR #1441, opened 2026-06-19
drivers/hda: ignore controllers without chosen outputs instead of assert

no92:


Qwinci (approved):
Looks good as a quick fix, in the long term all of the outputs should be exposed as ALSA devices with input events for presence changes.

PR #1444, opened 2026-06-22
IOMMU, part 11: Realtek NICs

no92:


drivers/nic/rtl8168/src/rtl8168.cpp
```diff
@@ -460,9 +460,10 @@ async::result<void> RealtekNic::init() {
 		}
 	}
 
+	_rxQueue = co_await RxQueue::create(*this, NUM_RX_DESCRIPTORS);
+	_txQueue = co_await TxQueue::create(*this, NUM_RX_DESCRIPTORS);
```
avdgrinten: `NUM_TX_DESCRIPTORS`?

avdgrinten (approved):

PR #1445, opened 2026-06-22
drivers/nic: Add igc driver for Intel I225/226

marv7000:


drivers/nic/igc/src/main.cpp
```diff
@@ -0,0 +1,435 @@
...
+	    }))
+		std::println("igc: NVM auto-read did not complete after reset");
+
+	space_.store(reg::imc, ~uint32_t{0});
+	space_.load(reg::icr);
+}
+
+void IgcNic::readMac(std::array<uint8_t, 6> &out) {
+	uint32_t ral = space_.load(reg::ral0);
+	uint32_t rah = static_cast<uint32_t>(space_.load(reg::rah0));
```
no92: Why is the cast to uint32_t necessary? The logic suggests that this could be a uint16_t.

drivers/nic/igc/src/main.cpp
```diff
@@ -0,0 +1,435 @@
...
+	out[2] = ral >> 16;
+	out[3] = ral >> 24;
+	out[4] = rah;
+	out[5] = rah >> 8;
+}
+
+void IgcNic::initRxAddrs(const std::array<uint8_t, 6> &mac) {
+	uint32_t ralVal = uint32_t(mac[0]) | (uint32_t(mac[1]) << 8) | (uint32_t(mac[2]) << 16)
+	                  | (uint32_t(mac[3]) << 24);
+	auto rahVal =
```
no92: Same as above.

no92 (approved):

PR #1448, opened 2026-06-28
thor: Fix x86 AP startup hangs on some devices

marv7000:
Putting a fiber to sleep might affect the APIC state, which affects the INIT sequence.

### DIFF
--- a/kernel/thor/arch/x86/cpu.cpp
+++ b/kernel/thor/arch/x86/cpu.cpp
@@ -804,6 +804,10 @@ void bootSecondary(unsigned int apic_id, size_t cpuIndex) {
 	if(disableSmp)
 		return;
 
+	// Run this function with interrupts disabled to prevent other IPIs (e.g., broadcast IPIs for shootdown)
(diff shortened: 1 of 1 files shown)


Dennisbonke (approved):
LGTM

kernel/thor/arch/x86/cpu.cpp
```diff
@@ -842,14 +842,18 @@ void bootSecondary(unsigned int apic_id, size_t cpuIndex) {
 	// The BIOS is not involved in this process at all.
 	infoLogger() << "thor: Booting AP " << apic_id << "." << frg::endlog;
 	raiseInitAssertIpi(apic_id);
-	KernelFiber::asyncBlockCurrent(generalTimerEngine()->sleepFor(10'000'000)); // Wait for 10ms.
+	pollSleepNano(10'000'000); // Wait for 10ms.
```
avdgrinten, line 849: I think we should additionally disable IRQs in this functions entirely (e.g., by using irqMutex()).

PR #1449, opened 2026-06-29
libblockfs: Fix various rename() failure modes

### LINKED ISSUE #704: libblockfs: Rename request is missing error handling
When running configure for GNU sed, we trip the following assert.
`In function servePartition, file ../../../src/managarm/drivers/libblockfs/src/libblockfs.cpp:754: Assertion 'result.error() == protocols::fs::Error::fileNotFound' failed!`
This seems to happen in the `checking whetjer rename honors trailing slash on destination` check in configure.

avdgrinten:
This PR fixes various issues with rename() on ext2, mostly in failure cases.

Fix #704.

### DIFF
--- a/drivers/libblockfs/src/libblockfs.cpp
+++ b/drivers/libblockfs/src/libblockfs.cpp
@@ -232,31 +232,16 @@ struct HandlePartition {
 				}
 			}
 
+			// A missing destination is the common case and not an error.
 			auto result = co_await newInode->removeEntry(req.new_name());
-			if(!result) {
-				if(result.error() == protocols::fs::Error::fileNotFound) {
-					// Ignored
-				} else if(result.error() == protocols::fs::Error::directoryNotEmpty) {
-					resp.set_error(managarm::fs::Errors::DIRECTORY_NOT_EMPTY);
+			if(!result && result.error() != protocols::fs::Error::fileNotFound) {
+				resp.set_error(result.error() | protocols::fs::toFsError);
 
-					auto ser = resp.SerializeAsString();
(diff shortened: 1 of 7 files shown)


drivers/libblockfs/src/libblockfs.cpp
```diff
@@ -155,13 +155,27 @@ struct HandlePartition {
 		}
 
 		auto &fs = *fsPtr;
+
+		auto isInvalidName = [](const std::string &name) {
```
no92: Wouldn't a string view suffice?

drivers/libblockfs/src/libblockfs.cpp
```diff
@@ -218,7 +218,17 @@ struct HandlePartition {
...
 			}
-			co_await newInode->link(req.new_name(), old_file.value().inode, old_file.value().fileType);
+			auto link_result = co_await newInode->link(req.new_name(),
+					old_file.value().inode, old_file.value().fileType);
+			if(!link_result) {
+				resp.set_error(link_result.error() | protocols::fs::toFsError);
+
+				auto ser = resp.SerializeAsString();
+				auto [send_resp] = co_await helix_ng::exchangeMsgs(conversation,
+					helix_ng::sendBuffer(ser.data(), ser.size()));
```
no92, line 313: `sendBragiHead{Only,Tail}` avoids use of `SerializeAsString`. Same in other places.

avdgrinten: That's true but I think we should refactor the whole file in one go.

no92 (commented):
The logic looks sound, but I'm not sure if the `rename("file", "dir/");` case would be handled correctly.

PR #1456, opened 2026-06-29
posix: Correctly report UID/GID for SO_PASSCRED

Dennisbonke:


no92:
We should add a testcase for this, rest LGTM.

PR #1458, opened 2026-07-01
libblockfs: Remove cached uid/gid from BaseInode

### LINKED ISSUE #1451: chown does nothing
Running `chown` on a file reports success, but if one then runs `ls -lah` on the file, the owner is not updated. Observed as both root and managarm user, no logs of relevance printed.

avdgrinten:
Fix #1451 by not returning stale cached values from stat().


Dennisbonke (approved):
LGTM

PR #1462, opened 2026-07-01
libblockfs: Add ext2 inode/block allocation locks and inode cache locks

avdgrinten:
This is a first PR to add proper locking to libblockfs. We add locks for a few simple cases. Proper per-inode locks will come in a separate PR.


Dennisbonke (approved):
LGTM

Qwinci (approved):

PR #1464, opened 2026-07-01
libblockfs: Add inode metadata and rename mutexes

avdgrinten:
This ensures that all inode metadata and directory entry manipulation is safe regardless of concurrency and suspension point ordering.


no92 (approved):
LGTM

PR #1470, opened 2026-07-04
Add RISC-V hypervisor support

Qwinci:
Add RISC-V hypervisor support to thor along with an interrupt injection syscall (not yet implemented for x86_64). Also install all of the helix headers to the proper `helix` subdir inside the sysroot (previously they were put at `system-root/usr/include` and were broken due to that).


kernel/thor/arch/riscv/cpu.cpp
```diff
@@ -184,6 +185,13 @@ void initializeThisProcessor() {
 	if (((sstatus >> riscv::sstatus::uxlShift) & riscv::sstatus::uxlMask) != riscv::sstatus::uxl64)
 		panicLogger() << "thor: kernel only supports 64-bit userspace" << frg::endlog;
 
+	cpuData->haveVirtualization = riscvHartCapsNote->hasExtension(RiscvExtension::h);
+
+	if (cpuData->haveVirtualization) {
+		debugLogger() << "thor: Hypervisor extension is supported" << frg::endlog;
+		riscv_hypervisor::init();
+	}
```
avdgrinten, line 193: Instead of copying to `CpuData`, we could look at `riscvHartCapsNote` directly.

avdgrinten (commented):
I read through this once and most things make sense. I'm not 100% convinced about the exception handling though.

kernel/thor/arch/riscv/cpu.cpp
Qwinci: I though that it would be better to do like this to avoid having to have two `riscvHartCapsNote->hasExtension(RiscvExtension::h)` right after each other. The field needs to be filled for the generic code anyways.

kernel/thor/generic/thor-internal/load-balancing.hpp
```diff
@@ -135,4 +135,31 @@ struct LoadBalancer {
 	async::barrier barrier_;
 };
 
+struct MigrationGuard {
```
avdgrinten: `MigrationGuard` is now unused and hence we can drop it again.

hel/include/hel.h
```diff
@@ -801,17 +863,40 @@ struct HelThreadStats {
...
+  kHelVmexitHyperCall = 2,
+  kHelVmexitInstructionTrap = 3,
   kHelVmexitError = -1,
   kHelVmexitUnknownPlatformSpecificExitCode = -2,
 };
 
+enum {
+  kHelVmFaultRead = 0,
+  kHelVmFaultWrite = 1 << 1,
+  kHelVmFaultExecute = 1 << 2,
```
avdgrinten: Should read be `1 << 0` or `0`?

kernel/thor/arch/riscv/hypervisor.cpp
```diff
@@ -0,0 +1,440 @@
...
+	riscv::writeCsr<riscv::Csr::vsstatus>(state_.sstatus);
+	riscv::writeCsr<riscv::Csr::vsie>(state_.sie);
+	riscv::writeCsr<riscv::Csr::vstvec>(state_.stvec);
+	riscv::writeCsr<riscv::Csr::vsscratch>(state_.sscratch);
+	riscv::writeCsr<riscv::Csr::vsepc>(state_.sepc);
+	riscv::writeCsr<riscv::Csr::vscause>(state_.scause);
+	riscv::writeCsr<riscv::Csr::vstval>(state_.stval);
+	riscv::writeCsr<riscv::Csr::vsip>(state_.sip);
+	riscv::writeCsr<riscv::Csr::vsatp>(state_.satp);
+	riscv::writeCsr<riscv::Csr::vstimecmp>(state_.stimecmp);
```
avdgrinten: I think this needs to be gated behind the Sstc extension (we already check for Sstc for time keeping).

kernel/thor/generic/hel.cpp
```diff
@@ -2415,6 +2466,14 @@ HelError helStoreRegisters(HelHandle handle, int set, const void *image) {
 		if(!readUserObject(reinterpret_cast<const HelX86VirtualizationRegs *>(image), regs))
 			return kHelErrFault;
 		vcpu.vcpu->storeRegs(&regs);
+#elif defined(__riscv) && __riscv_xlen == 64
+		if(!vcpu.vcpu) {
+			return kHelErrIllegalArgs;
+		}
```
avdgrinten: We can move this out of the `#ifdef` (as in the load case).

avdgrinten (approved):

kernel/thor/generic/thor-internal/thread.hpp
```diff
@@ -220,6 +220,10 @@ struct Thread final : ScheduleEntity, Credentials {
 		return _pendingConditions.load(std::memory_order_relaxed);
 	}
 
+	bool checkCancelConditions() {
+		return _pendingConditions.load(std::memory_order_relaxed) & (condition::interrupt | condition::terminate);
```
avdgrinten: Nit: there is a `cancelConditions` constant defined at the top of the file.

avdgrinten (approved):

PR #1476, opened 2026-07-06
posix: Implement flock upgrades

no92:


posix/subsystem/src/fs.hpp
```diff
@@ -216,6 +216,8 @@ struct FsNode {
 
 	void notifyObservers(uint32_t inotifyEvents, const std::string &name, uint32_t cookie, bool isDir = false);
 
+	protocols::fs::FlockManager flockManager;
```
avdgrinten, line 219: Shouldn't this be per FS? Is it even allowed to lock sysfs and procfs files etc?

posix/subsystem/src/file.cpp
```diff
@@ -525,3 +530,10 @@ async::result<protocols::fs::Error> File::shutdown(int how) {
 
 	co_return protocols::fs::Error::notSocket;
 }
+
+async::result<protocols::fs::Error> File::flock(int flags) {
+	assert(_link);
+	auto node = _link->getTarget();
+	assert(node);
+	return node->flockManager.lock(&flock_, flags);
+}
```
avdgrinten: For `extern_fs` files, `File::flock()` will now disagree with what extern FS does. I think the fix is to scope this only to files internal to posix-subsystem (maybe by having a `struct PosixFile : File` or similar for internal files).

protocols/fs/src/file-locks.cpp
```diff
@@ -7,77 +7,85 @@ namespace protocols::fs {
...
+					co_return protocols::fs::Error::wouldBlock;
+
 				co_await flockNotify.async_wait();
+				continue;
+			}
 
-				flocks.push_back(newFlock);
-				newFlock->manager = this;
-				co_return protocols::fs::Error::none;
+			if (newFlock->hook_.in_list) {
```
avdgrinten: This is an internal debug-only field. We should store an explicit bool inside the Flock struct.

PR #1479, opened 2026-07-06
posix: setuid / setgid work

Dennisbonke:


posix/subsystem/src/requests/uid-gid.cpp
```diff
@@ -288,4 +308,48 @@ HandleRequest::operator()(managarm::posix::SetGroupsRequest &&req,
...
+async::result<std::expected<void, DispatchError>>
+HandleRequest::operator()(managarm::posix::GetSuidRequest &&req,
+		helix::BorrowedDescriptor conversation, bragi::preamble preamble,
+		std::shared_ptr<Process> self, std::shared_ptr<Generation>) {
+	id = preamble.id();
+	logBragiRequest(req);
+
+	logRequest(logRequests, self, "GET_SUID", "suid={}", self->threadGroup()->suid());
+
+	managarm::posix::SvrResponse resp;
```
no92: We should use dedicated messages instead of the overloaded SvrResponse.

posix/subsystem/src/requests/uid-gid.cpp
```diff
@@ -89,6 +89,26 @@ HandleRequest::operator()(managarm::posix::SetUidRequest &&req,
...
+		helix::BorrowedDescriptor conversation, bragi::preamble preamble,
+		std::shared_ptr<Process> self, std::shared_ptr<Generation>) {
+	id = preamble.id();
+	logBragiRequest(req);
+
+	logRequest(logRequests, self, "SET_SUID", "suid={}", req.suid());
+
+	Error err = self->threadGroup()->setSuid(req.suid());
+	if(err == Error::accessDenied) {
+		co_await sendErrorResponse(conversation, managarm::posix::Errors::ACCESS_DENIED);
```
no92: This should also use the dedicated messages as a template param to `sendErrorResponse`.

no92 (changes requested):

no92 (approved):

PR #1482, opened 2026-07-07
posix: return EOF when reading from disconnected SOCK_STREAM unix socket

no92:


avdgrinten:
This needs a test to validate vs. Linux (I think the PR is correct though).

posix/subsystem/src/un-socket.cpp
```diff
@@ -160,8 +160,8 @@ struct OpenFile : FileWithDefaults {
 public:
 	async::result<std::expected<size_t, Error>>
 	readSome(Process *, void *data, size_t max_length, async::cancellation_token ct) override {
-		if(socktype_ == SOCK_STREAM && _recvQueue.empty() && _currentState == State::remoteShutDown)
-			co_return std::unexpected{Error::brokenPipe};
+		if(socktype_ == SOCK_STREAM && _recvQueue.empty() && (_currentState == State::remoteShutDown || _remote->shutdownFlags_ & shutdownWrite))
```
avdgrinten, line 163: Why isn't `remoteShutDown` equivalent to `_remote->shutdownFlags & shutdownWrite`?

no92: The states do behave similarly, but not identically; for instance, the poll behavior is different.

avdgrinten: I guess `remoteShutDown` should then be renamed to `remoteClosed`. We don't need to do it in this PR though.

avdgrinten (approved):

PR #1486, opened 2026-07-09
drivers/libblockfs: Fix .. handling in traverseLinks()

### LINKED ISSUE #1472: ext2: Missing the `..` on directories created inside Managarm
If one creates a directory inside Managarm, it seems that the `..` is either missing or corrupted.

Image

avdgrinten:
Fix #1472


avdgrinten:
The bug here was the `nodes.pop_back()` in the `..` case: the code confused the `nodes` vector for a vector of entered directories but it is actually a vector that mirrors the `components` array (which is different when there are `.` or `..` components).

no92 (approved):

PR #1489, opened 2026-07-13
drivers/libevbackend: Advertise EV_SYN for every input device

Dennisbonke:


drivers/libevbackend/src/libevbackend.cpp
```diff
@@ -472,6 +472,9 @@ EventDevice::EventDevice(std::string name, uint16_t bustype, uint16_t vendor, ui
 	memset(_relBits.data(), 0, _relBits.size());
 	memset(_absBits.data(), 0, _absBits.size());
 
+	// Linux advertises EV_SYN for every input device.
+	_typeBits[EV_SYN / 8] |= (1 << (EV_SYN % 8));
```
no92: Wouldn't it be nicer to use the `setBit` helper from below?

no92 (approved):

PR #1492, opened 2026-07-16
posix: Various set/get id fixes for sudo

Dennisbonke:


no92 (approved):

protocols/posix/posix.bragi
```diff
@@ -1220,69 +1220,69 @@ head(128):
 	int32 result;
 }
 
-message SetResuidRequest 233 {
```
avdgrinten: I'd drop the compaction commit, the benefit is negligible and it has some downsides (Managarm vs. mlibc divergence).

avdgrinten (commented):
This PR needs tests, ideally ones that can be executed as non-root or that detect if they are executed as root.

PR #1499, opened 2026-07-22
thor: Rework Universe data structure to enable lock-free read side

avdgrinten:
Refactor `Universe` to use RCU. This makes the read size (= `getDescriptor()` etc.) lock-free.

Instead of using a locked hash map, the new `Universe` data structure is a vector of chunks where each chunk stores 64 descriptor slots. Slots are re-used and a generation counter is used to ensure that we never access stale slots.

### DIFF
--- a/kernel/thor/generic/thor-internal/rcu-base.hpp
+++ b/kernel/thor/generic/thor-internal/rcu-base.hpp
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <concepts>
+
 #include <frg/list.hpp>
 
 namespace thor {
@@ -11,5 +13,13 @@ private:
 	void (*call_)(RcuCallable *);
 	frg::default_list_hook<RcuCallable> hook_;
 };
+
+// Marker that promises that all instances of a type are RCU protected.
+// If shared_ptrs are used, this includes their refcoutns (which allocate_rcu_shared() guarantees).
+// Deriving classes can enforce this by providing appropriate create() factories instead of exposing public constructors.
+struct RcuProtected {};
+
+template<typename T>
+concept IsRcuProtected = std::derived_from<T, RcuProtected>;
 
 } // namespace thor
--- a/kernel/thor/generic/thor-internal/universe.hpp
+++ b/kernel/thor/generic/thor-internal/universe.hpp
@@ -244,6 +244,9 @@ struct AnyDescriptor {
 
 	AnyDescriptor() = default;
 
+	AnyDescriptor(smarter::adopt_rc_t, DescriptorType type, uint8_t extra, void *object, smarter::counter *ctr)
+	: type_{type}, extra_{extra}, object_{object}, ctr_{ctr} { }
+
 	AnyDescriptor(const AnyDescriptor &other)
 	: type_{other.type_}, extra_{other.extra_}, object_{other.object_}, ctr_{other.ctr_} {
 		if(ctr_)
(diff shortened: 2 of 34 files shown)


kernel/thor/generic/thor-internal/rcu-base.hpp
```diff
@@ -12,4 +14,12 @@ struct RcuCallable {
 	frg::default_list_hook<RcuCallable> hook_;
 };
 
+// Marker that promises that all instances of a type are RCU protected.
+// If shared_ptrs are used, this includes their refcoutns (which allocate_rcu_shared() guarantees).
```
marv7000, line 18:
```suggestion
// If shared_ptrs are used, this includes their refcounts (which allocate_rcu_shared() guarantees).
```

avdgrinten:
This gives a modest improvement for multi-threaded programs (~13% in the parallel IPC benchmark):
```
[vm-util] ci-boot stdout: ipc ops (parallel, 4 threads)
before:
[vm-util] ci-boot stdout:     avg: 13141040, std: 75126
after:
[vm-util] ci-boot stdout:     avg: 14822780, std: 50147
```

kernel/thor/generic/thor-internal/universe.hpp
```diff
@@ -364,17 +381,59 @@ AnyDescriptor::resolveObject<DescriptorType::lane>() const;
...
+		DescriptorType type = DescriptorType::none;
+		uint8_t extra = 0;
+	};
+	static_assert(sizeof(Slot) == 32);
+
+	struct Root : RcuCallable {
+		size_t numChunks;
 
+		std::atomic<Slot *> *chunks() {
+			return reinterpret_cast<std::atomic<Slot *> *>(this + 1);
```
qookei, line 424: This seems potentially fragile if we add new members to `Root` or `RcuCallable`

kernel/thor/generic/thor-internal/universe.hpp
```diff
@@ -364,17 +381,59 @@ AnyDescriptor::resolveObject<DescriptorType::lane>() const;
...
+		// Encodes the state of this slot.
+		// - Live slots: the invalidMarker bit is clear.
+		//   The value matches the handle of the slot.
+		// - Detached slots: the invalidMarker bit is set.
+		//   Generation bits store the next generation to use.
+		//   Index bits store the next slot in the free-list that the slot is part of.
+		std::atomic<uint64_t> state{invalidMarker | (uint64_t{1} << slotIndexBits)};
+		// The remainder of the fields are constant after attachDescriptor().
+		// They remain valid until reuse (i.e., until a RCU grace period has passed after detachDescriptor()).
+		void *object = nullptr;
```
qookei, line 413: If these are constants for the lifetime of the descriptor, maybe we could mark them `const`, and then when replacing slots use placement new instead of a regular assignment? (and `std::launder` when grabbing the slot). Not sure if it's worth the effort though.

kernel/thor/generic/thor-internal/universe.hpp
qookei: Maybe we just add `static_assert(sizeof(Root) == sizeof(std::atomic<Slot *>);` and a comment? (since the code allocating it also depends on that for alignment purposes etc).

qookei (commented):
In general LGTM. I'm wondering if there's a nicer way to write the `Root` class with the trailing pointers, but nothing comes to mind immediately.

kernel/thor/generic/thor-internal/universe.hpp
avdgrinten: The `+ 1` happens before the cast so this is fine even if the size of `Root` changes.

kernel/thor/generic/thor-internal/universe.hpp
avdgrinten: I'll keep this for now since it avoids messing with `std::launder`.

PR #1500, opened 2026-07-22
Minor epoll fixes

no92:
These came up during testing on a fresh sysroot, albeit with the mlibc ABI break applied locally.


Dennisbonke (approved):
LGTM

PR #1501, opened 2026-07-22
thor: Parallelize Thor load-balancer ownership transfers

Dennisbonke:
This parallelizes Thor’s load-balancer ownership transfers while keeping affinity, assigned CPU, ownership, and load accounting synchronized.

Previously, balancing serialized ownership changes and repeatedly scanned every CPU pair, increasing lock contention and balancing overhead. Accounting also held node locks while updating individual thread load state, and affinity tests left helper threads runnable after completing.

Ownership transfers now use ordered per-CPU node locks, bounded and rotating scans, and rotating CPU-pair scheduling. Load accounting avoids holding node locks while updating threads and aggregates system load once per round. The tests also cover deferred affinity migration, respect round-robin placement, skip unnecessary empty transfers, and park helper threads after validation.

Blocked on #1502

### DIFF
--- a/kernel/thor/generic/thor-internal/load-balancing.hpp
+++ b/kernel/thor/generic/thor-internal/load-balancing.hpp
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <async/barrier.hpp>
+#include <frg/array.hpp>
 #include <frg/span.hpp>
 #include <thor-internal/coroutine.hpp>
 #include <thor-internal/cpu-data.hpp>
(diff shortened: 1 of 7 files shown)


kernel/thor/generic/thor-internal/load-balancing.hpp
```diff
@@ -129,14 +147,23 @@ struct LoadBalancer {
...
+	void move_(LbControlBlock *cb, LbNode *newNode);
+
 	coroutine<void> run_(CpuData *cpu);
 
 	// Move tasks from srcNode to dstNode to balance load.
 	// newLoad: newLoad at dstNode after balancing.
 	void balanceBetween_(LbNode *srcNode, LbNode *dstNode, uint64_t &newLoad, uint64_t idealLoad);
 
 	async::barrier barrier_;
+	frg::ticket_spinlock mutex_;
```
avdgrinten: A global lock for all load balancing operations is too restrictive.

kernel/thor/generic/thread.cpp
```diff
@@ -654,6 +673,32 @@ void Thread::invoke() {
 				<< " is activated" << frg::endlog;
 
 	assert(_runState == kRunDeferred);
+	if(_pendingConditions.load(std::memory_order_relaxed) & condition::affinity) {
```
avdgrinten: Migrating threads before resumption breaks the assumption that migration only happens at specific points in the kernel (and not during general kernel code execution).

kernel/thor/generic/thor-internal/thread.hpp
```diff
@@ -27,6 +27,8 @@ inline constexpr Condition exceptionalWq = Condition{1} << 1;
 inline constexpr Condition interrupt = Condition{1} << 2;
 // Request transition to kRunTerminated.
 inline constexpr Condition terminate = Condition{1} << 3;
+// Request migration after an affinity-mask change.
+inline constexpr Condition affinity = Condition{1} << 4;
```
avdgrinten: Integrating migration with the condition logic makes sense but this refactoring should be done as a separate prerequisite step instead of mixing it with logic changes. Also, it should replace the existing `raiseSignals()` logic instead of adding a second mechanism to trigger migration.

avdgrinten:
Pushed a rework of this branch.

PR #1507, opened 2026-08-02
thor, kernletcc, drivers: Support kernlets on all architectures

avdgrinten:
This adds support to handle the relocation types needed for kernlets on all architectures. This makes the virtio and EHCI kernlets usable on all architectures (not UHCI though since that uses PIO).

### DIFF
--- a/core/virtio/src/core.cpp
+++ b/core/virtio/src/core.cpp
@@ -322,6 +322,7 @@ private:
 	arch::mem_space _deviceSpace() { return arch::mem_space{_deviceMapping.get()}; }
 
 	async::detached _processIrqs();
+	async::result<void> _processIrqsWithoutKernlet();
 	async::detached _processQueueMsi();
 
 	protocols::hw::Device _hwDevice;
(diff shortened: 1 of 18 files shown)


core/virtio/src/core.cpp
```diff
@@ -539,6 +539,12 @@ async::detached StandardPciTransport::_processIrqs() {
 	auto kernlet_object = co_await compile(kernlet_program.data(),
 			kernlet_program.size(), {BindType::memoryView, BindType::offset,
 			BindType::bitsetEvent});
+	if(!kernlet_object) {
+		std::cout << "core-virtio: Kernlets are unavailable,"
+				" acknowledging IRQs from user space" << std::endl;
+		co_await _processIrqsWithoutKernlet();
```
qookei, line 545: This seems like a weird way to do it, as opposed to making `_processIrqsWithoutKernlet` detached as well.

kernel/thor/arch/arm/thor-internal/arch/paging.hpp
```diff
@@ -59,6 +59,26 @@ inline size_t isICachePIPT() {
 	return ((ctr >> 14) & 0b11) == 0b11;
 }
 
+inline void syncInstructionCache(void *pointer, size_t size) {
```
qookei, line 62: Can you call this from `ptSyncICache` to avoid code duplication?

PR #1521, opened 2026-08-06
hel, thor: Remove helEnableFullIo()

avdgrinten:
Instead of relying on `helEnableFullIo()`, drivers should use `helAccessIo()` + `helEnableIo()`. In fact, `helEnableFullIo()` has no callers anymore.


Dennisbonke (approved):
LGTM

PR #1522, opened 2026-08-07
Swapping of anonymous memory

48cf:
This PR adds swapping of non-CoW anonymous memory.


avdgrinten:
As discussed offline, we need to fix some races w.r.t. `discardPage()` here before merging (`BackingMemory` can lock a page even during discarding).

avdgrinten (approved):

PR #1524, opened 2026-08-08
sif: Port thor's PciConfigIo trait over to sif

avdgrinten:
This PR ports thor's `PciConfigIo` abstract base class over by turning it into a trait. This is useful to support non-legacy non-ECAM PCIe controllers (e.g., on the Raspberry Pi 4). The port faithfully replicates thor's architecture and logic.


avdgrinten:
Pushed another change that allows `helAccessPhysical()` to specify the caching mode. This is needed to replicate thor's `mmio` caching mode instead of passing the default one (which !x86 translates to WB).

marv7000 (approved):

PR #1525, opened 2026-08-08
sif: Make sif work on Aarch64 and RISC-V

avdgrinten:
This is a short PR that enables sif on Aarch64 and RISC-V (ACPI only). The only thing that was missing is removing an `#ifdef` around `helAccessIrq()` and passing the right model to `_PIC`.

The `configureIsaIrqs()` change is a separate fix to make the UART IRQ work when running under sif on x86.


marv7000 (approved):

avdgrinten:
Making this a draft because I'm not sure if we shouldn't move the ISA IRQ configuration into userspace instead of also doing it for sif in the kernel.

avdgrinten:
Re-implemented the ISA override logic to happen in userspace instead of thor.

PR #1532, opened 2026-08-11
protocols/fs: Fix missing reset after recvInline() on client side

avdgrinten:
We fixed the server side by adding the request dispatcher; however, we do not have a request dispatcher on the client side. Fix the `.reset()` calls manually.


Dennisbonke (approved):
LGTM

PR #1549, opened 2026-08-18
core/virtio, drivers/virtio-blk: Make the block I/O thread safe

avdgrinten:
This makes block I/O thread safe for virtio-block.


Dennisbonke (approved):
LGTM

PR #1550, opened 2026-08-18
protocols/{fs,ostrace}: Make thread-safe

avdgrinten:
Fix thread-safety issues in the protocols used by the FS stack.


Dennisbonke (approved):
LGTM

PR #1553, opened 2026-08-19
thor: Let backing memory report a static 2^62 size

avdgrinten:
`resize()` now only applies to the size of `FrontalMemory`. This way, manage requests can still be completed via `BackingMemory` even after `resize()` shrinks the `ManagedMemory`.


Dennisbonke (approved):
LGTM

PR #1554, opened 2026-08-19
posix: Truncate oversized affinity masks to the kernel mask size

avdgrinten:
This is a simple fix: Thor rejects affinity masks that a larger than ceil(N/8) bytes. Let posix-subsystem clamp to that size in case a caller passes a larger mask.


Dennisbonke (approved):
LGTM

PR #1559, opened 2026-08-20
drivers/libblockfs: Use metadata cache for all ext2 meta data

avdgrinten:
Convert the inode and block bitmaps and the inode table to `MetadataCache`. Make `MetadataCache` I/O concurrent.


avdgrinten:
Merging this because its mostly code removal and trivial refactoring.

PR #1564, opened 2026-08-22
rust: Add rust port of libarch

avdgrinten:
This allows declaring a register like this:
```rs
bit_register! {
    BridgeCtl @ 0x04: u32 {
        RESET @ 0, 1: bool;
        SW_INIT @ 1, 1: bool;
        MAX_BURST_SIZE @ 20, 2: u8;
    }
}
```
while accessing it as:
```rs
let value = unsafe { space.load(BridgeCtl) }
    .with(BridgeCtl::MAX_BURST_SIZE, 1)
    .with(BridgeCtl::SW_INIT, false);
unsafe { space.store(BridgeCtl, value) };
```


rust/arch/src/bits.rs
```diff
@@ -0,0 +1,123 @@
...
+///
+/// Unlike a scalar register's value, a bit value is opaque: it is addressed through the fields of
+/// the register instead of through arithmetic.
+#[derive(Clone, Copy)]
+pub struct BitValue<R: Register> {
+    bits: R::Bits,
+}
+
+impl<R: Register> BitValue<R> {
+    pub fn new(bits: R::Bits) -> BitValue<R> {
```
marv7000, line 26: Many functions here could be made `const fn`

PR #1568, opened 2026-08-23
sif: Add support for the bcm711 PCIe controller and Raspberry Pi 4 XHCI firmware upload

avdgrinten:
Port the corresponding Thor code to `sif`.


avdgrinten:
Drafted because I did not test this on real hardware yet.

PR #1576, opened 2026-08-26
thor, libblockfs: Discard pages beyond end of file after truncation

### LINKED ISSUE #1122: libblockfs: HEL_CHECK failure when writeback triggers after truncation
In particular, `helUpdateMemory()` returns an error after being called on a memory range outside of the memory object. The fix is to simply ignore writebacks in the truncated area of a file.

avdgrinten:
This PR writes the `BackingMemory::invalidateRange()` function to re-use the discard code path that we introduced for anonymous swap. libblockfs uses this to discard the pages beyond the end of a file when the file is truncated.

Fix https://github.com/managarm/managarm/issues/1122


avdgrinten:
Tested by Dennis on Epiphany, so I'll merge this.

PR #1612, opened 2026-09-07
Various fixes to the ELF loader

Dennisbonke:
This PR aims to improve the robustness of the ELF loader against malformed ELFs and shebang shell scripts. It also contains bugfixes for suid/sgid handling and a fix for initrd files to have proper file metadata.


posix/subsystem/src/exec.cpp
```diff
@@ -294,24 +294,71 @@ execute(ViewPath root, ViewPath workdir,
...
+		co_return Error::argumentListTooLong;
+	wordCount += env.size();
+	if(wordCount & 1) {
+		if(wordCount == maxSize)
+			co_return Error::argumentListTooLong;
+		wordCount++;
+	}
+	if(wordCount > maxSize / sizeof(uintptr_t)
+			|| !addStackSize(wordCount * sizeof(uintptr_t))
+			|| !addStackSize(18 * sizeof(uintptr_t))
```
no92: Where does the magic 18 come from?

posix/subsystem/src/exec.cpp
```diff
@@ -294,24 +294,71 @@ execute(ViewPath root, ViewPath workdir,
...
+			co_return Error::argumentListTooLong;
+	}
+
+	// Account for string alignment, the fixed auxv and argv/envp words.
+	if(!addStackSize(alignof(uintptr_t)))
+		co_return Error::argumentListTooLong;
+	if(!addStackSize(15))
+		co_return Error::argumentListTooLong;
+
+	size_t wordCount = 3;
```
no92, line 538: What does wordCount do? This is non-obvious to me.

posix/subsystem/src/tmp_fs.cpp
```diff
@@ -422,6 +423,39 @@ struct InheritedNode final : Node {
...
+				case ENOTDIR:
+					co_return Error::notDirectory;
+				case ENAMETOOLONG:
+					co_return Error::nameTooLong;
+				default:
+					co_return Error::ioError;
+			}
+		}
+
+		FileStats stats{};
```
no92: It might be nice to have a `FileStats(struct stat &)` constructor instead.

posix/subsystem/src/exec.cpp
no92: Does readSome guarantee to read as many bytes as available? Or can it return partial reads, even if more bytes would be readable?

PR #1615, opened 2026-09-09
drivers/libsound: Fix some correctness issues

Qwinci:
- Populate driver capabitilies and msbits in hardware parameters.
- Populate timestamp fields.
- Add max available frame reporting.
- Make uHDA path setup to only run once.


Dennisbonke (approved):
LGTM

PR #1618, opened 2026-09-09
hel, thor: Add hierarchy capability to express resource management hierarchy

avdgrinten:
This PR adds a "hierarchy capability" that we will use for resource accounting. Hierarchy capabilities form a tree. The root node is created by the kernel and handed to the servers that we spawn initially. Userspace can add new children to existing nodes by using the new `helExtendHierarchy()` syscall.

In this PR, posix-subsystem creates a node for each process  and another node for shared memory. libblockfs creates a node for file data and a node for metadata.

Each memory view construction syscall takes a hierarchy capability handle. Whenever the memory view needs to allocate physical memory, it charges that memory to its hierarchy node (but it does not propagate this charge upwards). As a follow up, we will add a way to limit the memory that a given hierarchy node and all of its children can allocate.


avdgrinten:
One thing to note here is that after forking a `CopyOnWriteMemory` the shared pages are charged both to the parent hierarchy node and to the child hierarchy node. This overestimates the physical memory that we actually need but it avoids the need to remember which page came from the parent and which came from the child etc. and it also means that CoW cannot fail to charge resources when accessing already copied pages.

kernel/thor/generic/servers.cpp
```diff
@@ -292,7 +302,7 @@ coroutine<void> executeModule(managarm::svrctl::Description<KernelAlloc> &desc,
 
 	// allocate and map memory for the user mode stack
 	size_t stack_size = 0x10000;
-	auto stackMemoryOutcome = AllocatedMemory::create(stack_size);
+	auto stackMemoryOutcome = AllocatedMemory::create(rootHierarchy(), stack_size);
```
qookei, line 305: Shouldn't this be charged to the server's hierarchy?

posix/subsystem/src/process.cpp
```diff
@@ -1399,19 +1458,20 @@ async::result<std::shared_ptr<ThreadGroup>> Process::init(std::string path) {
...
+	process->_vmContext = co_await VmContext::clone(std::move(hierarchy), original->_vmContext);
 	process->_fsContext = FsContext::clone(original->_fsContext);
 	process->_fileContext = FileContext::clone(original->_fileContext);
 	process->threadGroup()->_signalContext = SignalContext::clone(original->threadGroup()->_signalContext);
 
 	original->pgPointer()->reassociateProcess(threadGroup);
 
 	HelHandle thread_memory;
-	HEL_CHECK(helAllocateMemory(0x1000, 0, nullptr, &thread_memory));
+	HEL_CHECK(helAllocateMemory(rootHierarchy(), 0x1000, 0, nullptr, &thread_memory));
```
qookei, line 1474: Shouldn't this be charged to the process' hierarchy?

posix/subsystem/src/process.cpp
```diff
@@ -397,7 +445,7 @@ std::shared_ptr<FileContext> FileContext::create() {
 	context->_universe = helix::UniqueDescriptor(universe);
 
 	HelHandle memory;
-	HEL_CHECK(helAllocateMemory(0x1000, 0, nullptr, &memory));
+	HEL_CHECK(helAllocateMemory(rootHierarchy(), 0x1000, 0, nullptr, &memory));
```
qookei, line 448: Shouldn't this be charged to the process' hierarchy?

posix/subsystem/src/process.cpp
```diff
@@ -421,7 +469,7 @@ std::shared_ptr<FileContext> FileContext::clone(std::shared_ptr<FileContext> ori
...
```
qookei, line 472: Shouldn't this be charged to the process' hierarchy?

core/lib/process-data.cpp
```diff
@@ -0,0 +1,22 @@
+#include <core/process-data.hpp>
+#include <hel.h>
+#include <hel-syscalls.h>
+#include <protocols/posix/data.hpp>
+#include <protocols/posix/supercalls.hpp>
+
+namespace core {
+
+HelHandle getProcessHierarchy() {
+	static HelHandle handle = [] {
```
qookei, line 10: This is a bit of an edge case, but is this handle valid after `fork()`?

hel/include/hel.h
qookei, line 236: Add a comment describing what derive means for hierarchy?

qookei, line 243: Add a comment describing what provision means for hierarchy?

avdgrinten:
Fixed everything except for the attribution of the file table and thread page. These are attributed to the generic posix node for now since `VmContext` and `FileContext` have different lifetimes and untangling this would be involved. It doesn't matter too much for now since these structures are very small compared to the overall process image.

PR #1623, opened 2026-09-10
drivers/xhci: Gate power cycling behind kernel command line

avdgrinten:
Power cycling is currently broken. Disable it and fix a few other USB related issues.


avdgrinten:
I guess the alternative to this is just removing the power cycling. We should eventually make power cycling available via sysfs but it should probably not stay in the initialization path.

qookei (approved):

PR #1629, opened 2026-09-13
implement fs syncing

ElectrodeYT:
Requires [mlibc#1896](https://github.com/managarm/mlibc/pull/1896).


kernel/thor/generic/memory-view.cpp
```diff
@@ -2089,6 +2089,14 @@ coroutine<frg::expected<Error>> BackingMemory::writebackFence(uintptr_t offset,
 	if (offset > backingMemoryLength || size > backingMemoryLength - offset)
 		co_return Error::bufferTooSmall;
 
+	if(!size)
+		co_return {};
+
+	// Mapped stores may only have dirtied PTEs, without marking the managed
+	// pages dirty yet. Collect them from every mapping (including other address
+	// spaces) before waiting for the resulting writeback transactions.
+	co_await _managed->_evictQueue.cleanRange(offset, size);
```
avdgrinten: This comment is a bit weird: "including other address space" -- this function is not really associated with any address space.

testsuites/kernel-tests/src/memory.cpp
```diff
@@ -33,25 +34,70 @@ async::result<void> handleManageRequests(helix::BorrowedDescriptor backingMemory
...
 	);
+	auto lock = lockMemory.descriptor();
+	helix::Mapping first{frontalMemory, pageOffset, 0x1000, kHelMapProtRead | kHelMapProtWrite};
+	helix::Mapping second{frontalMemory, pageOffset, 0x1000, kHelMapProtRead | kHelMapProtWrite};
+
+	// No writeMemory() or synchronizeSpace(): only the mapping's PTE is dirty.
+	// Repeat through another alias to check that the fence collects all mappings
+	// and that stores after the first shootdown are picked up again.
+	for(auto mapping : {first.get(), second.get()}) {
+		auto expected = mapping == first.get() ? std::byte{42} : std::byte{43};
```
avdgrinten, line 69: This relies on the dirty bit not getting cleared randomly (e.g., by page aging), so it could be flaky.

drivers/libblockfs/src/metadata-cache.cpp
```diff
@@ -305,45 +305,79 @@ void MetadataCache::markDirty_(uint64_t block) {
...
+			if(dirtyBlocks_.empty() && !flushingDirty_)
+				break;
+		}
+		co_await dirtyDrainedEvent_.async_wait(seenSeq);
 	}
+
+	auto writeback = co_await helix_ng::writebackFence(
+			backing_, 0, numBlocks_ << blockPagesShift_);
+	HEL_CHECK(writeback.error());
+}
```
avdgrinten, line 354: Since the previous commit make `writebackFence()` also clean all mappings, we do not actually need to wait for `flushDirty_()` here. This means that we can revert the `flushDirty_()` changes and the patch becomes overall simpler.

drivers/libblockfs/src/ext2/ext2fs.cpp
```diff
@@ -1042,25 +1042,19 @@ async::result<void> FileSystem::init() {
...
-
-	auto bgdt_offset = (2048 + blockSize - 1) & ~size_t(blockSize - 1);
+	auto bgdtSize = size_t{numBlockGroups} * blockGroupDescriptorSize;
+	auto bgdtBlock = (2048 + blockSize - 1) >> blockShift;
 	protocols::ostrace::Timer bgdtTimer;
-	co_await device->readSectors((bgdt_offset >> blockShift) * sectorsPerBlock,
-			blockGroupDescriptorBuffer);
+	std::vector<MetadataCache::BlockWindow> bgdtBlocks;
+	for(size_t offset = 0; offset < bgdtSize; offset += blockSize)
+		bgdtBlocks.push_back(co_await accessMetadata(bgdtBlock + (offset >> blockShift), false));
```
avdgrinten: This passes `false` for the `writable` argument but we do in fact write through this window.

drivers/libblockfs/src/ext2/ext2fs.hpp
```diff
@@ -478,21 +478,32 @@ struct Inode final : BaseInode, std::enable_shared_from_this<Inode> {
...
 struct BlockGroupDescriptorTable {
-	inline void init(std::byte *ptr, uint16_t descriptorSize) {
-		ptr_ = ptr;
+	inline void init(std::vector<MetadataCache::BlockWindow> blocks,
+			size_t blockSize, uint16_t descriptorSize) {
+		blocks_ = std::move(blocks);
+		blockSize_ = blockSize;
+		assert(descriptorSize && !(blockSize % descriptorSize));
 		descriptorSize_ = descriptorSize;
 	}
```
avdgrinten, line 489: We can take `blockShift` instead of `blockSize`, that gets rid of the divisions by a runtime value below.

drivers/libblockfs/src/ext2/ext2fs.hpp
```diff
@@ -658,6 +664,9 @@ struct FileSystem final : BaseFileSystem {
 	// Serializes block/inode allocation and BGDT modifications.
 	async::mutex allocationMutex;
 
+	// Serializes filesystem-wide and per-inode synchronization sequences.
+	async::mutex synchronizeMutex;
```
avdgrinten: This looks unnecessary to me since all the primitives (`writebackFence()`, `device->flush()` etc) are already thread-safe.

avdgrinten (commented):
Looks good to me regarding correctness. Some minor points below.

testsuites/kernel-tests/src/memory.cpp
ElectrodeYT: Not really sure how to best fix this tbh, what would you suggest? If I'm looking at this right, there isn't much we can do to make this less flaky

avdgrinten (approved):
LGTM.

PR #1630, opened 2026-09-13
Swapping of CoW memory

48cf:
Patches cleaned with the help of @avdgrinten, this allows us to swap virtually all of the memory allocated by posix-subsystem for processes and threads, like stacks and private anonymous memory.


avdgrinten:
LGTM. We should add tests in a separate PR but it is nice to get this merged since hierarchy caps + OOM handling change that I want to do touch the same code paths.

PR #1639, opened 2026-09-17
sif: Add IOMMU support

avdgrinten:
This lets sif attach PCI devices to IOMMUs. The IOMMU driver stays in the kernel but we add new APIs to interact with it:
- `helAccessIommu()` is a new syscall to get an IOMMU capability. IOMMUs are resolved based on their MMIO base.
- `helCreateDmaSpace()` now takes the IOMMU and set the of RMRRs that the new DMA space has to respect.
- There are new SQ operations to bind/unbind PCI devices from IOMMUs and to enable IOMMU translation globally. These operations have to be async because they need to invalidate IOMMU translation caches.

In sif itself, the only code that was ported over is the PCI device -> IOMMU space association which works by walking the PCIe hierarchy.


avdgrinten:
Merging because without it, sif dies on Intel hardware and that makes it harder for Marvin to test the battery driver.

PR #1640, opened 2026-09-17
drivers/gfx: Fix various bugs in the initial DRM state

avdgrinten:
Fixes a few simple bugs:
- Potential nullptr dereference in `DRM_IOCTL_MODE_GETCRTC`
- The initial DRM state should have no plane or framebuffer attached to the CRTC (as we're not in fact rendering any framebuffer before the initial modeset). This matches Linux.
- All drivers need to call `claimDevice()` _before_ reprogramming the device.


avdgrinten:
Since plainfb no longer attaches a framebuffer during boot, this also means that thor renders up until the point where Weston launches. But this only works for plainfb, other drivers need to `claimDevice()` earlier and they'll probably need a userspace DRM client to draw the kernel log.

no92 (approved):

PR #1641, opened 2026-09-18
thor: Fix spawn functions to run start() on their target WQs

avdgrinten:
`spawnOnWorkQueue()` and `onExceptionalWq()` did set the receiver's associated work queue to the target WQ but they still ran `start()` inline. Fix them to also run `start()` on the target WQ.


avdgrinten:
The CI failure is triggered by a separate bug that's caused by scheduling RCU barriers on offline CPUs. I'll post a fix to this PR also for that bug.

PR #1651, opened 2026-09-22
thor: Allow userspace to manually track ManagedSpace dirty states

avdgrinten:
This adds a `kHelMapNoDirtyTracking` bit for `helMapMemory()` to suppress dirty tracking of the mapping and a `kHelUpdateMarkDirty` mode for `helUpdateMemory()` to manually set the dirty bit of pages.


kernel/thor/generic/memory-view.cpp
```diff
@@ -2548,6 +2560,47 @@ Error BackingMemory::updateRange(ManageRequest type, size_t offset, size_t lengt
...
+			++it;
+			cursor = pit->cachePage.identity + 1;
+			_managed->markDirtyPage(pit, needsEvent, needsExpedite);
+		}
+	}
+
+	if(needsEvent)
+		_managed->_dirtyEvent.raise();
+	if(needsExpedite)
+		_managed->_expediteEvent.raise();
```
Qwinci, line 2599: ~~This seems to be dead code?~~ Nvm, see above.

Qwinci (approved):
~~Looks mostly good to me, though there seems to be some dead code in BackingMemory::markDirtyRange? Not sure whether that's intentional or not so I though Ill bring it up just in case.~~ Ignore that, I missed the fact that it takes in references.

